PR #6, opened 2018-03-30
CustomCraft currentIndex Fix

yenzgamingyt:
Fixing the currentIndex to allow changing it individually per item.
This is used to create a custom entry in the customcraft files.
I am just starting to get into coding so if something is wrong, i will try to fix it.

Usage would be like so:
TechTypePatcher.currentIndex = 10001;
(Your added items via the modding helper)

TechTypePatcher.currentIndex = 10002;
(Your added items via the modding helper)


ahk1221:
Doesn't make sense, goes against the very idea of adding items be modular. Don't worry, someones working on a Custom Craft remake that is compatible with this.

PR #9, opened 2018-05-22
CraftTree.Type patcher + Custom CraftTree addition

PrimeSonic:
New features
- Allows modders to extend the CraftTree.Type enum (in much the same way the current TechType patching is done)
- With a new CraftTree.Type, a completely new independent craft tree can be added and associated to it.
  - This new tree is not part of any of the existing fabricators in the base game. As such, a new type of fabricator should be created to make use of this new crafting tree.
  - A new set of classes are available to assist in creating the custom crafting tree.


PrimeSonic:
Rolled back to last known good commit

PrimeSonic:
Branch is in a working state and (new) code is clean enough for my tastes

PR #10, opened 2018-05-26
Added Precompilation Directives

K07H:
Added DEBUG directives to remove logging from Release-builds during compilation process.
DuckFat


AlexejheroYTB:
What?

AlexejheroYTB:
The logging is needed

PrimeSonic:
Like I said over Discord.
Most of this logging is done during the initial save loading, and it uses Subnautica's own log writer, so it wouldn't impact game performance.

PR #11, opened 2018-06-08
CrossMod CustomCraftNode

PrimeSonic:
Update to allow adding TechTypes created by other SMLHelper mods into CustomCraftNode


ahk1221:
Looks pretty much fine.

PR #12, opened 2018-06-13
Better exceptions message

K07H:
Most of the time you don't get any useful information from the Exception but instead you get some from the InnerException. I use this all the time and I was wondering why it's not there already as it really helps to track down issues.

### DIFF
--- a/SMLHelper/Initializer.cs
+++ b/SMLHelper/Initializer.cs
@@ -29,7 +29,13 @@ namespace SMLHelper
             }
             catch (Exception e)
             {
-                Console.WriteLine(e.StackTrace);
+                Console.WriteLine("Exception caught" + ((e != null && !String.IsNullOrEmpty(e.Message)) ? " Message=[" + e.Message + "]" : ""));
+                Console.WriteLine("StackTrace=[" + e.StackTrace + "]");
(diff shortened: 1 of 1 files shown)


PrimeSonic:
Better option is to simply use the exception's `ToString` method.
All the Exception is written and labeled  with ToString().

K07H:
You don't have inner exception with ToString.

K07H:
When I have a bug at SMLHelper level I don't have any useful information from the log.
This modification allows me to know why SMLHelper failed and to fix my code accordingly

PrimeSonic:
The default `ToString` also calls ToString on the InnerException if one exits.
https://msdn.microsoft.com/en-us/library/system.exception.tostring%28v=vs.90%29.aspx
```The default implementation of ToString obtains the name of the class that threw the current exception, the message, the result of calling ToString on the inner exception, and the result of calling Environment.StackTrace. If any of these members is null, its value is not included in the returned string.```

ahk1221:
Have to change the brackets, I don't like them imo, but accepting since I have to go

K07H:
You don't get InnerException details with ToString.
That's the only way to get full information.
@PrimeSonic

PR #14, opened 2018-06-14
- Updated project setting to build XML Documentation file.

PrimeSonic:
- This file allows the XML documentation we write to be visible to other projects that reference SMLHelper by DLL.
- For external projects using SMLHelper, the file "SMLHelper.xml" only needs to be in the same directory where SMLHelper.dll is being referenced from.

References:
https://stackoverflow.com/questions/1632942/how-do-you-get-xml-comments-to-appear-in-a-different-project-dll
https://docs.microsoft.com/en-us/dotnet/csharp/codedoc

### DIFF
--- a/SMLHelper/SMLHelper.csproj
+++ b/SMLHelper/SMLHelper.csproj
@@ -21,6 +21,7 @@
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <ErrorReport>prompt</ErrorReport>
     <WarningLevel>4</WarningLevel>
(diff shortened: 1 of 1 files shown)


PrimeSonic:
I would recommend that the SMLHelper.xml file be a separate download on the Nexus page.
Have it read something like "For developers only".

PrimeSonic:
Before this is merged in, I would recommend we add XML Documentation on all public facing classes, methods, properties, and fields.

PrimeSonic:
Changes applied, closing pull request.

PR #22, opened 2018-06-27
Minor modifications and small fixes

AlexejheroYTB:

- [ ] ~~Changed ReflectionHelper class~~

~~New usage:~~

- ~~For instance fields:~~
~~`__instance.GetInstanceField("fieldName"); // Returns an object`~~
~~`__instance.SetInstanceField("fieldName", object);`~~
- ~~For static fields:~~
~~`__instance.GetStaticField("fieldName"); // Returns an object`~~
~~`__instance.SetStaticField("fieldName", object);`~~
~~or, if not used in a harmony patch:~~
~~`ReflectionHelper.GetStaticField<class>("fieldName"); // Returns an object`~~
~~`ReflectionHelper.SetStaticField<class>("fieldName", object");`~~

~~Comments: The old reflection helper had a lot of unnecessary methods that were just calling other ones, so I thought it would be better to simplify things.~~

- [x] Changed [Obsolete] attributes

Example 1:

- Old: 
`[Obsolete("CraftSchemes are oboslete. Use CraftTree.Types instead", false)]`
would output in the editor:
`'CraftScheme' is obsolete: 'CraftSchemes are oboslete. Use CraftTree.Types instead'`

- New:
`[Obsolete("Use CraftTree.Type instead")]`
outputs in the editor:
`'CraftScheme' is obsolete: 'Use CraftTree.Type instead'`

Example 2:

- Old:
`[Obsolete("CraftDataPatcher.customTechData is obsolete. Please use SMLHelper.V2 instead.")]`
would output in the editor:
`'CraftDataPatcher.customTechData' is obsolete: 'CraftDataPatcher.customTechData is obsolete. Please use SMLHelper.V2 instead.'`

- New:
`[Obsolete("Use SMLHelper.V2 instead.")]`
outputs in the editor:
`CraftDataPatcher.customTechData` is obsolete: `Use SMLHelper.V2 instead.`

Comments: In the editor, the user would see the same message twice (examples above), and that didn't look very nice, so I changed it. Plus, I removed all "Please"s to make it more consistent and to better impose to the user that he should use `V2` instead.

- [x] Added `#pragma warning disable CS0618` to the beginning of all files that were causing a warning

Comments: no more warnings!

- [x] Renamed `BaseBioReactorHandler` to `BioReactorHandler`

Comments: More user-friendly

**Note: this pull request might cause some edit conflicts which will need to be resolved**

### DIFF
--- a/SMLHelper/CraftNodeToScrub.cs
+++ b/SMLHelper/CraftNodeToScrub.cs
@@ -1,6 +1,6 @@
 ﻿namespace SMLHelper
 {    
-    [System.Obsolete("SMLHelper.CraftNodeToScrub is obsolete. Please use SMLHelper.V2 instead.")]
+    [System.Obsolete("Use SMLHelper.V2 instead.")]
     public class CraftNodeToScrub
     {
         public CraftTree.Type Scheme;
@@ -12,7 +12,7 @@
             Path = path;
         }
 
-        [System.Obsolete("CraftSchemes are obsolete. Use CraftTree.Types instead.")]
+        [System.Obsolete("Use CraftTree.Type instead.")]
         public CraftNodeToScrub(CraftScheme scheme, string path)
             : this(Utility.CraftSchemeMap[scheme], path)
         {
--- a/SMLHelper/CraftScheme.cs
+++ b/SMLHelper/CraftScheme.cs
@@ -1,6 +1,6 @@
 ﻿namespace SMLHelper
 {
-    [System.Obsolete("CraftScheme is obsolete. Use CraftTree.Type instead.")]
+    [System.Obsolete("Use CraftTree.Type instead.")]
     public enum CraftScheme
     {
         Fabricator,
--- a/SMLHelper/CustomCraftNode.cs
+++ b/SMLHelper/CustomCraftNode.cs
@@ -3,7 +3,7 @@ using SMLHelper.V2.Util;
 
 namespace SMLHelper
 {
-    [System.Obsolete("SMLHelper.CustomCraftNode is obsolete. Please use SMLHelper.V2 instead.")]
+    [System.Obsolete("Use SMLHelper.V2 instead.")]
     public class CustomCraftNode
     {
         public TechType TechType;
@@ -34,7 +34,7 @@ namespace SMLHelper
             }
         }
 
-        [System.Obsolete("CraftSchemes are obsolete. Use CraftTree.Types instead.")]
+        [System.Obsolete("Use CraftTree.Type instead.")]
         public CustomCraftNode(TechType techType, CraftScheme scheme, string path)
             : this(techType, Utility.CraftSchemeMap[scheme], path)
         {
--- a/SMLHelper/CustomCraftTab.cs
+++ b/SMLHelper/CustomCraftTab.cs
@@ -2,7 +2,7 @@
 
 namespace SMLHelper
 {
-    [System.Obsolete("SMLHelper.CustomCraftTab is obsolete. Please use SMLHelper.V2 instead.")]
+    [System.Obsolete("Use SMLHelper.V2 instead.")]
     public class CustomCraftTab
     {
         public string SpriteId
@@ -41,13 +41,13 @@ namespace SMLHelper
         {
         }
 
-        [System.Obsolete("CraftSchemes are obsolete. Use CraftTree.Types instead.")]
+        [System.Obsolete("Use CraftTree.Type instead.")]
         public CustomCraftTab(string path, string name, CraftScheme scheme, Atlas.Sprite sprite) 
             : this(path, name, Utility.CraftSchemeMap[scheme], sprite)
         {
         }
 
-        [System.Obsolete("CraftSchemes are obsolete. Use CraftTree.Types instead.")]
+        [System.Obsolete("Use CraftTree.Type instead.")]
         public CustomCraftTab(string path, string name, CraftScheme scheme, UnityEngine.Sprite sprite)
             : this(path, name, Utility.CraftSchemeMap[scheme], sprite)
         {
--- a/SMLHelper/CustomCraftTreeFamily.cs
+++ b/SMLHelper/CustomCraftTreeFamily.cs
@@ -9,7 +9,7 @@ namespace SMLHelper
     /// <summary>
     /// Basic data structure of a custom crafting tree node.
     /// </summary>
-    [System.Obsolete("SMLHelper.CustomCraftTreeNode is obsolete. Please use SMLHelper.V2 instead.")]
+    [System.Obsolete("Use SMLHelper.V2 instead.")]
     public abstract class CustomCraftTreeNode
     {
         internal static bool Initialized = false;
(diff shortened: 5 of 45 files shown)


PrimeSonic:
`Comments: The old reflection helper had a lot of unnecessary methods that were just calling other ones, so I thought it would be better to simplify things.`
Those methods, while redundant, offer more ways to easily use the provided extension methods.
There were options to call static fields and methods both with an without an instance reference.

Also, without the `this` notation, these methods can't be invoked as extension methods anymore.
At least I'm not able to invoke them without also using the `ReflectionHelper` name.

This utility class is publicly accessible and could be used externally for ease of use.
More methods can make for feature that are easier to discover by outside consumers of the API.
See `/2.0/SMLHelper/V2/Handlers/CraftDataHandler.cs` for an example of just that.

There's also methods here that were removed that have no suitable replacement.
`CopyFields` is just gone. Even if it's not in use right now, it's still utility that can be used later.

Could you rollback the changes Reflection Helper for now?
Everything else looks good but I think this part needs needs a little more discussion before we make such a radical change to it.

AlexejheroYTB:
I did not realize I also removed the method calls and the CopyFields.
In regard to extension methods, I've kept the `this` keyword both for the instance and for the static methods

PrimeSonic:
I see that was added back in on a later commit.
The rest still stands.

Let's go over Reflection Helper as a separate thing for now and get the rest merged in.

AlexejheroYTB:
Reverted ReflectionHelper changes and updated post

AlexejheroYTB:
Question: Can a method info be used for something else other than invoking? (Something else that is actually useful and could be needed in SMLHelper)
Proposal: If not, directly invoke the method in the reflection helper
Benefits: this way we wouldn't need to write the instance twice
(Example: `__instance.GetPrivateMethod("method").Invoke(__instance, techType, 1, "example");` would become `__instance.InvokeMethod("method", techType, 1, "example");`)

PrimeSonic:
> `Can a method info be used for something else other than invoking?`
> `Proposal: If not, directly invoke the method in the reflection helper`

That sounds reasonable.
Check for any references/usages of GetPrivateMethod/GetPrivateStaticMethod and make sure that's what they're being used for.

Beyond that, as a public API, here's a few things I would personally want to ensure from this utility class

1. All methods should be capable of be called using either the extension method syntax or also by invoking the `ReflectionHelper` static class name. This can be easily accomplished with the use of method overloads.
2. Methods dealing with static fields or methods can be invoked by an instance of the class or using the class itself directly. The second being a more clear usage because the static item does not belong to the instance, even though the instance can call it directly.
3. Maintain flexibility. The `BindingFlags` parameter that defaults to `BindingFlags.Default` was added because there may be times when it is necessary to pass an additional flag, for example `BindingFlags.FlattenHierarchy`. Using default parameters and overloaded methods will keep the API usage flexible and robust while keeping the code simple for the 99 of 100 use cases.
4. Clear semantics and intention: The method name and corresponding XML documentation should make it quick and clear what the method is intended to be used for. This class shouldn't be a catch all way of accessing any field or method on a class. Ideally, it should only be used when it's absolutely necessary to use Reflection to access something non-public. That means using a longer name like `GetPrivateField` instead of `GetField` is preferable because it is clear that you won't need to use this all the time.

In short, this class doesn't need to be the one that's compact and minimalist.
It is here to allow code elsewhere to be more compact by avoiding the usual boilerplate Reflection code.
Take a look at the above items, feel free to comment on them, and we'll keep an open conversation on the subject.

AlexejheroYTB:
> Maintain flexibility. The `BindingFlags` parameter that defaults to `BindingFlags.Default` was added because there may be times when it is necessary to pass an additional flag, for example `BindingFlags.FlattenHierarchy`. Using default parameters and overloaded methods will keep the API usage flexible and robust while keeping the code simple for the 99 of 100 use cases.

I find that having all of the binding flags as a parameter, and not just an additional would be more clear to the user.

    public static object GetPrivateField<T>(this T instance, string fieldName, BindingFlags bindingFlags = BindingFlags.Default)
    {
        FieldInfo fieldInfo = typeof(T).GetField(fieldName, BindingFlags.NonPublic | BindingFlags.Instance | bindingFlags);
        return fieldInfo.GetValue(instance);
    }

Could be turned into

    public static object GetField<T>(this T instance, string fieldName, BindingFlags bindingFlags = BindingFlags.NonPublic | BindingFlags.Instance)
    {
        FieldInfo fieldInfo = typeof(T).GetField(bindingFlags);
        return fieldInfo.GetValue(instance);
    }

Or better...

    public enum RHType
    {
        PublicStatic = BindingFlags.Public | BindingFlags.Static,
        PrivateStatic = BindingFlags.NonPublic | BindingFlags.Static,
        PublicInstance = BindingFlags.Public | BindingFlags.Instance,
        PrivateInstance = BindingFlags.NonPublic | BindingFlags.Instance,
    }

And

    public static object GetField<T>(this T instance, string fieldName, HRType type)
    {
        FieldInfo fieldInfo = typeof(T).GetField((BindingFlags)type);
        return fieldInfo.GetValue(instance);
    }

What do you think?

PrimeSonic:
I've looked these over.

On the first proposal,  I feel like this change might actually make the code less readable, even if it's made more compact.
When calling something using the default parameters like `obj.GetField("name")` there's no longer any visible semantics that indicate what type of field is being requested. Without the context of what the default parameters are, you wouldn't be able to read/review the code outside of the IDE.

While this is just as readable as having a method for each basic case, it comes at the cost of having to add a second parameter every time.

There is a possible middle ground here:
This being a utility class, it's perfectly acceptable to offer "easy mode" and "expert mode" in the same place.

We can add a method like this as something for experts to use.
```CSharp
public static object GetField<T>(this T instance, string fieldName, BindingFlags bindingFlags)
{
    FieldInfo fieldInfo = typeof(T).GetField(bindingFlags);
    return fieldInfo.GetValue(instance);
}
```
But then when can then include methods like these
```CSharp
public static object GetPrivateField<T>(this T instance, string fieldName, BindingFlags bindingFlags = BindingFlags.Default) 
    => GetField(instance, fieldName, BindingFlags.NonPublic | BindingFlags.Instance | bindingFlags)
```
which can be simplified by looping back around onto the "expert mode" methods like in this example.

Remember, these methods are all effectively shortcuts to avoid having to write lots of reflection code.
Personally, I think it makes more sense to have more shortcuts than fewer.
Shortcuts that are semantically verbose can be quicker and cleaner to write than more compact shortcuts that require almost as much boilerplate code to use than the code they're supposed to be simplifying.

Ask yourself: 
If a single method with only a single parameter is enough to get the field you need, 
why would want to call a method that requires you to send in two or three parameters while also referencing the `System.Reflection` namespace to get the same thing?

AlexejheroYTB:
Well... the `System.Reflection` namespace wouldn't need to be referenced if we added our own enum.

AlexejheroYTB:
But I like your middle ground idea

PrimeSonic:
I find the custom enum to be a bit too half-way-there to really want to commit to it.
If someone wants to quickly use "expert mode", then they'll have no issues passing in the one or two `BindingFlags` needed rather than use some go-between enum.
If someone would rather use "easy mode", they'll gravitate to the simplest option.

That said, if you think it makes sense to have the custom enum along side the standard `BindingFlags`, then there's no reason we can't have both options by overloading the appropriate method calls.

Using something like that, I would suggest a more descriptive name for the enum. Example:
```CSharp
public enum BindingFlagShortcuts
{
    PrivateStatic = BindingFlags.NonPublic | BindingFlags.Static,
    PrivateInstance = BindingFlags.NonPublic | BindingFlags.Instance,
    PrivateInstanceFlat = BindingFlags.NonPublic | BindingFlags.Instance | BindingFlags.FlattenHierarchy,
}
```
Now, personally I wouldn't bother adding entries dealing with `BindingFlags.Public` because if the field or method is already public, then there's no need to use Reflection.

PR #23, opened 2018-06-28
*SMALL* changes to ReflectionHelper

AlexejheroYTB:

1. Renamed the `Util` namespace to `Utilities`

Comments: In my opinion, this would be better because it would tell the user that the Utilities are for him, and that they are not just for internal use

2. Made small tweaks to ReflectionHelper

- Removed all occurrences of `Private` in the method names

- Removed block bodies for methods that didn't need one

Example:

    public static object GetPrivateField<T>(this T instance, string fieldName, BindingFlags bindingFlags = BindingFlags.Default)
    {
        FieldInfo fieldInfo = typeof(T).GetField(fieldName, BindingFlags.NonPublic | BindingFlags.Instance | bindingFlags);
        return fieldInfo.GetValue(instance);
    }

Turned into:

    public static void SetInstanceField<T>(this T instance, string fieldName, object value, BindingFlags bindingFlags = BindingFlags.Default) where T : class
        => typeof(T).GetField(fieldName, BindingFlags.NonPublic | BindingFlags.Instance | bindingFlags).SetValue(instance, value);

- Added `where T : class` statements to all methods. 

Comments: This should prevent the user from using these methods wrongfully

### DIFF
--- a/SMLHelper/CustomCraftNode.cs
+++ b/SMLHelper/CustomCraftNode.cs
@@ -1,5 +1,5 @@
 ﻿using SMLHelper.V2.Patchers;
-using SMLHelper.V2.Util;
+using SMLHelper.V2.Utilities;
 
 namespace SMLHelper
 {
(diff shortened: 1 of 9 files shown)


AlexejheroYTB:
Disregard the last commit

PR #25, opened 2018-06-28
File renames

AlexejheroYTB:

- Renamed `Utilities` namespace to `Utility`
- Renamed `Utility` file to `MiscUtils` (+ changed class name)
- Moved `MiscUtils` file into `Utility` folder
- Renamed `BaseBioReactorPatcher` to `BioReactorPatcher` (handler already renamed in other pr)
- In `BioReactorPatcher`, moved `using` statements after the namespace declaration

Feel free to request changes if needed


SMLHelper/Utility/MiscUtils.cs
```diff
@@ -6,7 +6,7 @@
     using System.IO;
     using UnityEngine;
 
-    public class Utility
+    public class MiscUtils
```
PrimeSonic: @AlexejheroYTB  The usage of this class seems to indicate "PatchUtilities" or "PatchingUtilities" rather than anything miscellaneous.

@ahk1221 Question: Is everything being public here intentional?
If so, then it should probably be moved into the same `SMLHelper.V2.Utility` namespace.

PrimeSonic:
The question is still open to @ahk1221 
Is the utility class that exposes various patching methods intended to be public?

Disregard my earlier recommendation. Having this class in the `Patching` namespace does sense.
But if it is intended for external use, then it should be expanded on with at least the addition of XML Documentation.

If this class wasn't intended to be used externally, well, then it shouldn't be public.

PR #27, opened 2018-06-29
Added required DLLs in a "Dependencies" folder

AlexejheroYTB:
This should fix the .csproj reference issues.
Next time making a pull request, you won't have to change the .csproj reference paths
Hope that makes sense.


AlexejheroYTB:
:cosmicbrain:

PrimeSonic:
Looks like a cool idea given that collaborators won't always have Subnautica installed on the same path.

AlexejheroYTB:
For e.g., I have Subnautica installed in my external G:\ drive, but to reference files I use a junction link to my D:\ drive

ahk1221:
I had a similar idea but the problem with this is that we're distributing the Assembly-CSharp files. We could do something similar to what Nitrox does, it requires a path to your Subnautica directory and then the reference paths are extracted from that.

AlexejheroYTB:
> I had a similar idea but the problem with this is that we're distributing the Assembly-CSharp files

Could you please clarify what you mean by this?

ahk1221:
What I meant was that we're not supposed to put those files on the internet because anyone can download them (it violates the ToS) but what we could do is put the dlls into the folder ourselves from our own copy of the game, without them being on the repo.

AlexejheroYTB:
Oh I know what you mean

PrimeSonic:
To properly do this, we can't have any commits in which the assembly-csharp DLLs existed.
If they exist in the git-history, they can be retrieved.

AlexejheroYTB:
I will remove them from any commit (i've done this before)

PrimeSonic:
I can handle the commits in the SMLHelper branch, to remove those DLLs from the history, but that will have to wait until tonight.

PR #30, opened 2018-06-30
2.0 Old CraftTreePatcher Fix 

PrimeSonic:
Reworked the Old CraftTreePatcher logic to work with the new form of handling CraftTree overrides that combine fully custom trees and modded trees in the same framework.


PrimeSonic:
Craft trees may be working, but this doesn't resolve the issue that's causing the builder tool to fail to build anything.

PR #34, opened 2018-07-01
[2.0] CraftDataHandler method renames

AlexejheroYTB:
***
* Renamed EditHarvestOutputList to EditHarvestable
* Renamed EditHarvestTypeList to EditHarvestType
* Renamed EditCookedCreatureList to EditCookable
* Added an EditItemSize overload (TechType, int, int)

--- Changes also done to the "redundant but user-friendly" methods

***

### DIFF
--- a/SMLHelper/Handlers/CraftDataHandler.cs
+++ b/SMLHelper/Handlers/CraftDataHandler.cs
@@ -50,7 +50,7 @@
         /// </summary>
         /// <param name="techType">The TechType whose harvest output you want to edit.</param>
         /// <param name="harvestOutput">The harvest output for that TechType.</param>
-        public static void EditHarvestOutputList(TechType techType, TechType harvestOutput)
+        public static void EditHarvestable(TechType techType, TechType harvestOutput)
         {
             CraftDataPatcher.CustomHarvestOutputList[techType] = harvestOutput;
         }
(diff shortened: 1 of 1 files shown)


SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -88,6 +88,18 @@ public static void EditItemSize(TechType techType, Vector2int size)
...
         }
 
+        /// <summary>
+        /// <para>Allows you to edit item sizes for TechTypes.</para>
+        /// <para>Can be used for existing TechTypes too.</para>
+        /// </summary>
+        /// <param name="techType">The TechType whose item size you want to edit.</param>
+        /// <param name="x">The width of the item</param>
+        /// <param name="y">The hight of the item</param>
+        public static void EditItemSize(TechType techType, int x, int y)
```
PrimeSonic: Very nice. I like this. Very appropriate for a public API.

AlexejheroYTB: Just to be clear, X is the width and Y is the height, right?

SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -61,7 +61,7 @@ public static void EditHarvestOutputList(TechType techType, TechType harvestOutp
         /// </summary>
         /// <param name="techType">The TechType whose HarvestType you want to edit.</param>
         /// <param name="harvestType">The HarvestType for that TechType.</param>
-        public static void EditHarvestTypeList(TechType techType, HarvestType harvestType)
+        public static void EditHarvestType(TechType techType, HarvestType harvestType)
```
PrimeSonic: This is a good change. The original method name is informed too much by the underlying implementation.
The new method name is more straight to the point about what it intends to accomplish.

AlexejheroYTB:
Fixed typos, ready to merge

SMLHelper/Handlers/CraftDataHandler.cs
PrimeSonic: Wouldn't make sense otherwise

PrimeSonic:
I'll go ahead and call this good here.

PR #41, opened 2018-07-07
External ban list for enum registering

### LINKED ISSUE #39: Not compatible with other mods that adds new items
With the current state of development, this mod can conflict with any other mods that adds new items into the game.
This happens because the cache system of SMLHelper can associate already used IDs (from other mods) to new items (from SMLHelper).
This incompatibility problem will not occur for RandyKnapp's mods tho (since his mods items ID have been hardcoded into SMLHelper).

ahk1221:
Fixes #39 
Closes #42

### DIFF
--- a/SMLHelper/Patchers/CraftTreeTypePatcher.cs
+++ b/SMLHelper/Patchers/CraftTreeTypePatcher.cs
@@ -10,12 +10,17 @@
 
     internal class CraftTreeTypePatcher
     {
-        #region Internal Fields
+        private const string CraftTreeTypeEnumName = "CraftTreeType";
 
-        internal const int startingIndex = 11; // The default CraftTree.Type contains indexes 0 through 10
-        internal static readonly EnumCacheManager<CraftTree.Type> cacheManager = new EnumCacheManager<CraftTree.Type>("CraftTreeType", startingIndex);
+        internal const int startingIndex = 11; // The default CraftTree.Type contains indexes 0 through 10        
 
-        #endregion
+        internal static readonly EnumCacheManager<CraftTree.Type> cacheManager =
+            new EnumCacheManager<CraftTree.Type>(
+                enumTypeName: CraftTreeTypeEnumName,
+                startingIndex: startingIndex,
+                bannedIndices: ExtBannedIdManager.GetBannedIdsFor(CraftTreeTypeEnumName));
+
+
 
         #region Adding CraftTreeTypes and TreeRoots
 
--- a/SMLHelper/Patchers/CraftTreeTypePatcher.cs
+++ b/SMLHelper/Patchers/CraftTreeTypePatcher.cs
@@ -32,7 +32,7 @@
                 };
             }
 
-            if (cacheManager.IsIndexConflicting(cache.Index))
+            if (cacheManager.IsIndexConflicting(cache.Index) || cacheManager.IsIndexBanned(cache.Index))
                 cache.Index = cacheManager.GetNextFreeIndex();
 
             craftTreeType = (CraftTree.Type)cache.Index;
--- a/SMLHelper/Patchers/CraftTreeTypePatcher.cs
+++ b/SMLHelper/Patchers/CraftTreeTypePatcher.cs
@@ -19,9 +19,7 @@
                 enumTypeName: CraftTreeTypeEnumName,
                 startingIndex: startingIndex,
                 bannedIndices: ExtBannedIdManager.GetBannedIdsFor(CraftTreeTypeEnumName));
-
-
-
+        
         #region Adding CraftTreeTypes and TreeRoots
 
         internal static ModCraftTreeRoot CreateCustomCraftTreeAndType(string name, out CraftTree.Type craftTreeType)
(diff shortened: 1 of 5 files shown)


SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -36,14 +38,55 @@ internal EnumCacheManager(string enumTypeName, int startingIndex, IEnumerable<in
...
+
+        private string GetCachePath()
+        {
+            return Path.Combine(GetCacheDirectoryPath(), $"{enumTypeName}Cache.txt");
+        }
+
+        private string GetBanListPath()
+        {
+            return Path.Combine(GetCacheDirectoryPath(), "Banlist.txt");
+        }
```
AlexejheroYTB: This means that users installing need to manually edit the ban list, because it can't be overridden, or else other banned indices might be lost

AlexejheroYTB: I think it would be better if you could have a banned indices folder where people could add their own text files, instead of editing the same one

PrimeSonic: From what I'm seeing, this external file loads indexes into the `bannedIndices` collection.
This collection can also be instantiated through the constructor `EnumCacheManager(string enumTypeName, int startingIndex, IEnumerable<int> bannedIndices)` to provide an initial hard coded list.

So the external indexes get combined with the internal ones.

PrimeSonic: Yes, this would be more flexible since having a folder of multiple files means a simple "drag and drop" action for the end user rather than opening a file, copy/pasting text, and saving it.

SMLHelper/Patchers/CraftTreeTypePatcher.cs
```diff
@@ -32,7 +32,7 @@ internal static ModCraftTreeRoot CreateCustomCraftTreeAndType(string name, out C
                 };
             }
 
-            if (cacheManager.IsIndexConflicting(cache.Index))
+            if (cacheManager.IsIndexConflicting(cache.Index) || cacheManager.IsIndexBanned(cache.Index))
```
PrimeSonic: Consider combining `IsIndexConflicting` and `IsIndexBanned` into a single method since the patcher classes don't really need to worry about why the index can't be used. They just need to know it can't be so they can request another one.

SMLHelper/Utility/EnumCacheManager.cs
ahk1221: So, basically something like this?
Modding Helper/Banned Ids/TestMod.txt
Modding Helper/Banned Ids/TestMod2.txt

And TestMod.txt can be something like this:
```
1111:TechType
10:CraftTreeType
```

SMLHelper/Patchers/CraftTreeTypePatcher.cs
ahk1221: Yeah, true.

PrimeSonic: On second thought, having to call both of this will only add to the code bloat if another enum patcher is ever added.
Best to combine these into `IsValidIndex` or `IsInvalidIndex` instead of having to remember to call both methods each time.

SMLHelper/Utility/EnumCacheManager.cs
AlexejheroYTB: yes

SMLHelper/Patchers/CraftTreeTypePatcher.cs
PrimeSonic: To elaborate,  `IsIndexConflicting` and `IsIndexBanned` can be kept as private methods while the new, combined method calls into both of them.
Whatever the combined method is called, it should be easy to read that it is informing the caller that the index is or isn't available to use.

SMLHelper/Utility/EnumCacheManager.cs
AlexejheroYTB: But keep the folder name without spaces (like the other ones)

SMLHelper/Utility/ExtBannedIdManager.cs
```diff
@@ -0,0 +1,116 @@
...
+    using System.Collections.Generic;
+    using System;
+
+    internal static class ExtBannedIdManager
+    {
+        private static bool IsInitialized = false;
+
+        private static readonly Dictionary<string, List<int>> BannedIdDictionary = new Dictionary<string, List<int>>();
+
+        private const string BannedIdDirectory = @"./QMods/Modding Helper/RestrictedIDs";
```
AlexejheroYTB: Maybe we could come up with a better name

PrimeSonic: I'm always open to suggestions for better names.
Since this is a folder the end-user may be expected to interact with, it should be easily understood what it's purpose is.

ahk1221:
Looks good to me

AlexejheroYTB:
Don't merge yet. There are still changes to be done #42

ahk1221:
Yeah, that's why I haven't merged yet. Was looking at the other PR.

SMLHelper/Utility/ExtBannedIdManager.cs
PrimeSonic: Any final word on this? I can go either way with this.

PR #42, opened 2018-07-07
2.0.OSubMarin - Additional checked for PreRegistered TechTypes and CraftTreeTypes

PrimeSonic:
`CraftTreeTypePatcher` and `TechTypePatcher` now have an additional layer of collision protection against pre-registered IDs from other mods.

While not 100% effective given the possible race conditions, this does still offer one more safety net that can be used along side the hard-coded ban list and the external ban list.


ahk1221:
Are SortedLists really a good idea? I don't really know much about them, but a quick Google search tells me that its some kind of dictionary. Max() should return the largest value.

PrimeSonic:
Right, I should be calling Last instead of Max if I want the benefit I was going for.
That being, getting the largest index in a single operation instead of having to crawl the whole list each time.

ahk1221:
Nice!

ahk1221:
(Sorry, misclick)

PR #45, opened 2018-07-08
2.0NodeHandling - Reworked node handling, now nodes are changed individually instead of…

### LINKED ISSUE #44: Not compatible with mods that adds new entries to the Mobile Vehicle Bay
With actual state of development, SML Helper is overwriting entries of the Mobile Vehicle Bay added by other mods.
This happens because SML Helper is taking ownership of game's CraftTree (thus CraftNodes) by overriding the CraftTree method "Get()" instead of manipulating the CraftTree directly.

ahk1221:
… changing entire tree

Fixes #44 (I hope) @K07H you should test it out.

### DIFF
--- a/SMLHelper/Handlers/CraftTreeHandler.cs
+++ b/SMLHelper/Handlers/CraftTreeHandler.cs
@@ -1,9 +1,8 @@
 ﻿namespace SMLHelper.V2.Handlers
 {
-    using Patchers;
     using Crafting;
-    using System.Collections.Generic;
-    using Assets;
+    using Patchers;
+    using UnityEngine.Assertions;
 
     /// <summary>
     /// A handler class for creating and editing of crafting trees.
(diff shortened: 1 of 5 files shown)


SMLHelper/Handlers/CraftTreeHandler.cs
```diff
@@ -86,45 +45,8 @@ public static ModCraftTreeRoot GetExistingTree(CraftTree.Type Scheme)
...
-            if (root == null)
-                return;
-
-            ModCraftTreeTab tab = root.GetTabNode(stepsToTab);
-
-            if (tab == null)
-                return;
-
-            tab.AddCraftingNode(craftingItems);
+            // Add to game
```
PrimeSonic: Just about all the methods that used to use the `GetExistingTree` method still have this line in their XML Documentation
` /// <seealso cref="GetExistingTree"/>`
Remove these entries to avoid any confusion in the documentation, if this hasn't already caused compiler warnings.

ahk1221: Removed :)

PrimeSonic: Verified

PrimeSonic:
Since this is mainly a reimplementation of the original craft tree editing code, and testing shows it to work with currently published mods that add craft nodes, add craft tabs, and remove nodes, it's sufficient for me.

SMLHelper/Handlers/CraftTreeHandler.cs
```diff
@@ -81,86 +40,24 @@ public static ModCraftTreeRoot GetExistingTree(CraftTree.Type Scheme)
...
-        /// <seealso cref="ModCraftTreeLinkingNode.AddCraftingNode(IEnumerable{TechType})"/>
-        public static void AddCraftingNode(CraftTree.Type craftTree, IEnumerable<TechType> craftingItems)
-        {
-            ModCraftTreeRoot root = GetExistingTree(craftTree);
-
-            if (root == null)
-                return;
-
-            root.AddCraftingNode(craftingItems);
+            CraftTreePatcher.CraftingNodes.Add(new CraftingNode(new string[0], craftTree, craftingItem));
```
PrimeSonic: Just noticed: These methods are now only for use with standard CraftTree.Types.
There's a bit more work to do here.
1. Remove the remaining XML Documentation referencing methods no longer used.
2. Add an Assert line to check that the CraftTree.Type is one of the standard trees.
3. Update the XML Documentation to mark that restriction.

PrimeSonic: Updated in fc5f0f8

PrimeSonic:
While the code changes are sound and verified to work with currently published mods, this may not be enough to solve Issue https://github.com/ahk1221/SMLHelper/issues/44

I still recommend me move forward with these changes and handle Issue 44 with another round of updates.

PR #46, opened 2018-07-18
Experimental change in attempts to halt Shared Access Violation errors

PrimeSonic:
THIS IS EXPERIMENTAL : **DO NOT MERGE**

While these changes were given a minor bit of testing to make sure it wouldn't break anything, there's no proof that it resolves the Shared Access Violation as seen in this log: https://hastebin.com/oqetivufoc

### DIFF
--- a/SMLHelper/Utility/EnumCacheManager.cs
+++ b/SMLHelper/Utility/EnumCacheManager.cs
@@ -4,6 +4,7 @@
     using System.Collections.Generic;
     using System.IO;
     using System.Text;
+    using System.Threading;
 
     internal class EnumTypeCache
     {
@@ -20,6 +21,8 @@
         private List<EnumTypeCache> cacheList = new List<EnumTypeCache>();
         private readonly HashSet<int> BannedIDs;
         private readonly int LargestBannedID;
+
+        private static bool fileInUse = false;
 
         internal Dictionary<T, EnumTypeCache> customEnumTypes = new Dictionary<T, EnumTypeCache>();
 
@@ -68,7 +71,11 @@
                 return;
             }
 
+            WaitForFile();
+
+            fileInUse = true;
             var allText = File.ReadAllLines(savePathDir);
+            fileInUse = false;
 
             foreach (var line in allText)
             {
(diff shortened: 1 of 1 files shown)


AlexejheroYTB:
I think it would be better if you did this with a while loop

AlexejheroYTB:
```cs
int passes = 0;
while (fileInUse)
{
    if (passes >= 50)
        break;
    passes++
}```

**Comment by K07H** [comment]
Your 50 passes will go really fast if you don't add a sleep().

**Comment by K07H** [comment]
Oh my bad, I didn't see the sleep() at first.

**Comment by AlexejheroYTB** [comment]
This will never work because the boolean will always be false when the WaitForFolder method is called.

**Comment by AlexejheroYTB** [comment]
Check https://stackoverflow.com/questions/876473/is-there-a-way-to-check-if-a-file-is-in-use for info on how this could could be done more easily

**Comment by AlexejheroYTB** [comment]
```cs
private static bool IsFileLocked(FileInfo file)
{
    FileStream stream = null;
    try
    {
        stream = file.Open(FileMode.Open, FileAccess.Read, FileShare.None);
    }
    catch (IOException)
    {
        //the file is unavailable because it is:
        //still being written to
        //or being processed by another thread
        //or does not exist (has already been processed)
        return true;
    }
    finally
    {
        if (stream != null)
            stream.Close();
    }
    //file is not locked
    return false;
}
```

PrimeSonic:
The bool I added was for the assumption that multiple threads might be trying to hit this file at once.

A different, likely better solution is being proposed which would render this change unneeded.

SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -102,7 +109,11 @@ internal void SaveCache()
                 stringBuilder.AppendLine(string.Format("{0}:{1}", entry.Value.Name, entry.Value.Index));
             }
 
+            WaitForFile();
```
AlexejheroYTB: By the time this is called, `fileInUse` would already be `false`, regardless of whether it's being used or not

AlexejheroYTB:
Sorry, I started the review before you closed it

PR #47, opened 2018-07-19
Cache Saving Optimization

PrimeSonic:
# Cache is saved when Patcher is being patched
Minimizing the number of writes being done to file to prevent Shared Access Violation exceptions as seen in this log: https://hastebin.com/oqetivufoc

### DIFF
--- a/SMLHelper/Patchers/CraftTreeTypePatcher.cs
+++ b/SMLHelper/Patchers/CraftTreeTypePatcher.cs
@@ -42,8 +42,6 @@
 
             Logger.Log("Successfully added CraftTree Type: \"{0}\" to Index: \"{1}\"", name, cache.Index);
 
-            cacheManager.SaveCache();
-
             var customTreeRoot = new ModCraftTreeRoot(craftTreeType, name);
 
             CraftTreePatcher.CustomTrees[craftTreeType] = customTreeRoot;
@@ -88,6 +86,8 @@
 
         internal static void Patch(HarmonyInstance harmony)
         {
+            cacheManager.SaveCache();
+
             var enumType = typeof(Enum);
             var thisType = typeof(CraftTreeTypePatcher);
             var techTypeType = typeof(CraftTree.Type);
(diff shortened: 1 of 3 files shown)


SMLHelper/Patchers/CraftTreeTypePatcher.cs
```diff
@@ -88,6 +86,8 @@ private static List<int> PreRegisteredCraftTreeTypes()
 
         internal static void Patch(HarmonyInstance harmony)
         {
+            cacheManager.SaveCache();
+
```
AlexejheroYTB: This reduces the chances of having errors, but they might still happen.

AlexejheroYTB:
Just put everything in a try catch block for IOException

SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -92,6 +92,15 @@ internal void LoadCache()
         internal void SaveCache()
         {
             var savePathDir = GetCachePath();
+
+            if(File.Exists(savePathDir)) // File exists
+            {
+                if(FileUtils.IsFileLocked(savePathDir))
+                {
+                    return; // File is locked
```
PrimeSonic: Assuming this condition happens, what part of the code is responsible for reattempting to write the file after this call into `SaveCache` was cut early?

ahk1221: TechTypePatcher.Patch? Don't get what you mean

PrimeSonic: `TechTypePatcher.Patch` is only calling `cacheManager.SaveCache();` and doesn't appear to be checking for any success/fail response.

Perhaps the `SaveCache` method should return a bool informing if the file was successfully saved.  
So in the event that it isn't, the caller can reattempt it after a short wait.

SMLHelper/Patchers/TechTypePatcher.cs
```diff
@@ -107,6 +105,8 @@ private static List<int> PreRegisteredTechTypes()
 
         internal static void Patch(HarmonyInstance harmony)
         {
+            cacheManager.SaveCache();
```
PrimeSonic: What happens if saving fails here?

PrimeSonic: And also the same question on the CraftTreeTypePatcher

ahk1221: Well, nothing.

PR #48, opened 2018-07-23
2.0 add group add after

ahk1221:

  - CraftDataPatcher.AddToCustomGroup now takes in a fourth parameter called `after` of type `TechType`
  - Instead of calling `techCategory.Add(techType)`, it calls `techCategory.Insert`, based on the index of the `after` parameter. 
  - The `after` parameter is optional, and the API has 2 overloads for AddToGroup: the first overload takes in nothing (so it will work as it did before) and the second overload takes in the, you guessed it, `after` parameter.


AlexejheroYTB:
:+1:

PR #49, opened 2018-07-29
2.0 modded craft node reverts

PrimeSonic:

  - Reverts `AddModdedCraftingNode` back to its original form to deal with errors
  - Adds two new methods `TryAddModdedCraftingNode` to replace it


PrimeSonic:
Since there changes were already reverted, the only change left in this branch is the addition of two methods to replace the functionality of what was lost.

I can go either way regarding including these or not.

PR #53, opened 2018-08-19
CraftData_NoGet

PrimeSonic:
- We no longer need to Prefix the CraftData Get method.
- The techData Dictionary can be accessed and updated with all modded recipes on the game's first load.


PrimeSonic:
I'm working now to apply the same sort of work to the SpriteManager.  
This could be what we need to improve performance on the Inventory screen.

ahk1221:
Looks very nice!

PR #54, opened 2018-08-19
 SpriteManager_NoGet

PrimeSonic:
- SpritePatcher no longer needs to patch the Get methods.
- The modded sprites are loaded directly into the dictionary where they will be loaded from.
- This should improve performance on screens with lots of sprite request calls

### DIFF
--- a/SMLHelper/Assets/ModSprite.cs
+++ b/SMLHelper/Assets/ModSprite.cs
@@ -1,5 +1,6 @@
 ﻿namespace SMLHelper.V2.Assets
 {
+    using System;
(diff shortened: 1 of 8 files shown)


PrimeSonic:
In conjunction with the update to [CraftData](https://github.com/ahk1221/SMLHelper/pull/53), opening the inventory screen on my game actually saw FPS numbers go UP by 10-15 FPS.

On the original DLL, opening the inventory screen would make FPS go DOWN by about 20.

I'm highly recommending that we see these through and release a minor revision update release.

PR #61, opened 2018-09-17
Fixes and error handling

PrimeSonic:
  - Correction to AssetsFolder property in Spawnable 
  - Checks and error logs for potential duplicate TechData entries

### DIFF
--- a/SMLHelper/Patchers/CraftDataPatcher.cs
+++ b/SMLHelper/Patchers/CraftDataPatcher.cs
@@ -10,6 +10,7 @@
     {
         #region Internal Fields
 
+        internal static List<TechType> DuplicateTechDataAttempts = new List<TechType>();
         internal static Dictionary<TechType, ITechData> CustomTechData = new Dictionary<TechType, ITechData>();
         internal static Dictionary<TechType, TechType> CustomHarvestOutputList = new Dictionary<TechType, TechType>();
         internal static Dictionary<TechType, HarvestType> CustomHarvestTypeList = new Dictionary<TechType, HarvestType>();
@@ -131,6 +132,18 @@
 
         private static void AddCustomTechDataToOriginalDictionary()
         {
+            if (DuplicateTechDataAttempts.Count > 0)
+            {
+                Logger.Log($"Removing conflicting TechData entries from patching.{Environment.NewLine}" +
+                           $"This is so the user will take notice and know to resolve the conflict.{Environment.NewLine}" +
+                           "Only one custom TechData may be added per TechType.");
+
+                foreach (TechType dup in DuplicateTechDataAttempts)
+                {
+                    CustomTechData.Remove(dup);
+                }
+            }
+
             Type CraftDataType = typeof(CraftData);
             Type TechDataType = CraftDataType.GetNestedType("TechData", BindingFlags.NonPublic);
             Type IngredientType = CraftDataType.GetNestedType("Ingredient", BindingFlags.NonPublic);
--- a/SMLHelper/Assets/Buildable.cs
+++ b/SMLHelper/Assets/Buildable.cs
@@ -12,7 +12,7 @@
         /// <summary>
         /// Override with the text to be displayed when your hover the cursor over the built item.
         /// </summary>
-        public abstract string HandOverText { get; }
+        public virtual string HandOverText => null;
 
         /// <summary>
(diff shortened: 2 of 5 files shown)


SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -131,6 +132,18 @@ private static void PreparePrefabIDCachePostfix()
 
         private static void AddCustomTechDataToOriginalDictionary()
         {
+            if (DuplicateTechDataAttempts.Count > 0)
+            {
+                Logger.Log($"Removing conflicting TechData entries from patching.{Environment.NewLine}" +
```
AlexejheroYTB: Couldn't you use "\n"?

PrimeSonic: It's generally better to always use `Environment.NewLine` for best compatibility.
This will use the correct line-break character for the file appropriate the OS and encoding,  
whether that be `\n` (Unix style) or `\r\n` (Windows Style).

The `output_log` file uses CRLF (`\r\n`) on Windows but it's possible that it could be using LF (`\n`) on the Mac version (or potentially someone running this on Linux using the new Proton API).  
And as a general rule, you don't want to have mixed line-ending styles in a single file.

So that's why it's generally a good habit to let .NET's `Environment.NewLine` deal with the line-endings for you.

Measurity:
Not really an issue for this PR but just to play it safe:

If `AssetsFolder` is required to not be null, which seems to be the case based on this: https://github.com/SMLHelper/SMLHelper/blob/1c78f22b731f172fad41b2ee34aba9af9fe0b541/SMLHelper/Assets/Spawnable.cs#L107 then it's more appropriate to set it through the constructor and check that it is not null. Or put a null check around that line.

If the code is fine as is then I recommend still adding a null check to that line and give a more descriptive exception+message (instead of NullRefException).

PrimeSonic:
> @Measurity  
> give a more descriptive exception+message (instead of standard NullRefException).

It is true that most of these properties would be better suited to being set by a constructor.  
The issue is that the final constructor would end up having so many parameters that it would become unreadable; Worse still when you consider that some of the properties are okay to default.

From a pure, software architecture perspective, yes, many of these should be implemented as `readonly` fields set through the abstract class's constructor.  
But, as mentioned, this results in a class that is no longer easy to use as it becomes harder to see what the values the user passes to the constructor are getting mapped to.

With that in mind, I don't think it's unreasonable to add special checks around these properties in the code to give the first-time-modder some extra guidance.

I won't be able to get to that myself right now, but I can certainly add that in tonight if no one else does.

AlexejheroYTB:
(Accidentally clicked resolve instead of comment)
Another way would be to have an optional parameter in the constructor that users can override if they want (make the default null, and do a nullcheck to set it to Assembly.GetCallingAseembly().etc....)

PrimeSonic:
> @AlexejheroYTB 
> > Another way would be to have an optional parameter in the constructor that users can override if they want (make the default null, and do a nullcheck to set it to Assembly.GetCallingAseembly().etc....)

`Assembly.GetCallingAseembly()` was the source of the problem.  
The calling assembly for this class is actually SMLHelper, and not the modder's assembly, meaning that this default value was being set to `SMLHelper/Assets`, and thus was not finding the Assets folder in my UpgradedVehicles mod.  
Because of this, we can't actually offer any defaults for this property.

Again on the subject of handling these through the constructor:  
between all these quick-start classes, there are 5 abstract and 3 virtual value typed properties.  
Adding an extra 8 parameters to the constructor, with all the possible values of handling the 3 that are optional, just seems like way more of a mess than it needs to be when this is intended to be making it easier for first-time users of SMlHelper.

As much as it would be more "correct" to add these values through the constructor, having them set through abstract and virtual get methods is a compromise I'm willing to champion for the sake of making things easier for the user. 

For the sake of better visualizing this in action, I already started using one of these base classes in an upcoming update to one of my mods.   
See link: [SpeedBooster with SMLHelper Craftable class](https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/blob/UpgradedVehicles_DownScale/UpgradedVehicles/Craftables/SpeedBooster.cs)

From a readability standpoint, it's actually not that bad if you compare to what this looks like using my old `Craftable` base class that handles everything through the constructors.   
See link: [SpeedBooster with protype Craftable class](https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/blob/master/UpgradedVehicles/Craftables/SpeedBooster.cs)

PR #65, opened 2018-12-22
Use dynamic starting index

killzoms:
Based on last value in TechType


killzoms:
ah didnt realize Unable to assign static readonly outside of initializers

killzoms:
Note i do have a small complaint, Usage of var keyword instead of type definition for variables

ahk1221:
I fear that this would break some mods or cause conflicts with other mods. It works fine as it is currently, and I don't want to risk conflicts.

PR #67, opened 2018-12-28
Use implicitly-typed variables instead of var

AlexejheroYTB:

  - Changed all `var` variables to be explicitly-typed
  - Simplified initialization of some objects (recommended by vs, this might need to be done in other files too)
  - Changed some `default(...);` to `default;` (recommended by vs, this might need to be done in other files too)
  - ~~Fixed Mods page not displaying anything if no mods are installed~~

### DIFF
--- a/SMLHelper/Legacy/Utility.cs
+++ b/SMLHelper/Legacy/Utility.cs
@@ -112,8 +112,8 @@ namespace SMLHelper
 
         public static void PatchList(Type type, string name, IList list, BindingFlags flags)
         {
-            var craftDataList = listField.GetValue(null) as IList;
             FieldInfo listField = type.GetField(name, flags);
+            var craftDataList = listField.GetValue(null) as IList;
 
             foreach (object obj in list)
             {
--- a/SMLHelper/Legacy/Utility.cs
+++ b/SMLHelper/Legacy/Utility.cs
@@ -22,15 +22,15 @@ namespace SMLHelper
 
         public static void AddBasicComponents(ref GameObject _object, string classId)
         {
-            var rb = _object.AddComponent<Rigidbody>();
+            Rigidbody rb = _object.AddComponent<Rigidbody>();
             _object.AddComponent<PrefabIdentifier>().ClassId = classId;
             _object.AddComponent<LargeWorldEntity>().cellLevel = LargeWorldEntity.CellLevel.Near;
-            var rend = _object.GetComponentInChildren<Renderer>();
+            Renderer rend = _object.GetComponentInChildren<Renderer>();
             rend.material.shader = Shader.Find("MarmosetUBER");
-            var applier = _object.AddComponent<SkyApplier>();
+            SkyApplier applier = _object.AddComponent<SkyApplier>();
             applier.renderers = new Renderer[] { rend };
             applier.anchorSky = Skies.Auto;
-            var forces = _object.AddComponent<WorldForces>();
+            WorldForces forces = _object.AddComponent<WorldForces>();
             forces.useRigidbody = rb;
         }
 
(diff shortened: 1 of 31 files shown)


SMLHelper/Legacy/Utility.cs
```diff
@@ -22,15 +22,15 @@ public class Utility
 
         public static void AddBasicComponents(ref GameObject _object, string classId)
         {
-            var rb = _object.AddComponent<Rigidbody>();
+            Rigidbody rb = _object.AddComponent<Rigidbody>();
```
PrimeSonic: This is another example of the type being explicit on the right-hand side.  
The type in the constructor as part of the generic definition.

This can be `var`.

PrimeSonic: I left another comment with the general `var` guidelines.  
The idea is to minimize how verbose the code gets without sacrificing readability.

PrimeSonic:
Looks good now.

SMLHelper/Patchers/OptionsPanelPatcher.cs
```diff
@@ -11,17 +12,19 @@ internal class OptionsPanelPatcher
...
         }
 
         internal static void AddTabs_Postfix(uGUI_OptionsPanel __instance)
         {
-            var optionsPanel = __instance;
-            var modsTab = optionsPanel.AddTab("Mods");
+            uGUI_OptionsPanel optionsPanel = __instance;
+            int modsTab = optionsPanel.AddTab("Mods");
+
+            if (modOptions.Count <= 0) optionsPanel.AddHeading(modsTab, "No options here...");
```
ahk1221: This won't work if mods outside of SMLHelper add mod options in the mod options tab. Better solution is to use reflection or something else, to see if anything exists in the mods tab.

PrimeSonic:
Looks neat. Will have to test before merging.

AlexejheroYTB:
Closed in favor of some upcoming changes to QModManager

AlexejheroYTB:
Actually, this PR is mostly about avoiding to use var.

SMLHelper/Utility/PatchUtils.cs
```diff
@@ -13,8 +13,8 @@ public static void PatchDictionary(Type type, string name, IDictionary dictionar
 
         public static void PatchDictionary(Type type, string name, IDictionary dictionary, BindingFlags flags)
         {
-            var dictionaryField = type.GetField(name, flags);
             var dict = dictionaryField.GetValue(null) as IDictionary;
+            FieldInfo dictionaryField = type.GetField(name, flags);
```
AlexejheroYTB: Some line orders have changed. This needs to be fixed here, for example

SMLHelper/Utility/PatchUtils.cs
```diff
@@ -29,10 +29,10 @@ public static void PatchList(Type type, string name, IList list)
 
         public static void PatchList(Type type, string name, IList list, BindingFlags flags)
         {
-            var listField = type.GetField(name, flags);
             var craftDataList = listField.GetValue(null) as IList;
+            FieldInfo listField = type.GetField(name, flags);
```
AlexejheroYTB: and here

SMLHelper/Legacy/Utility.cs
```diff
@@ -112,10 +112,10 @@ public static void PatchList(Type type, string name, IList list)
 
         public static void PatchList(Type type, string name, IList list, BindingFlags flags)
         {
-            var listField = type.GetField(name, flags);
             var craftDataList = listField.GetValue(null) as IList;
+            FieldInfo listField = type.GetField(name, flags);
```
AlexejheroYTB: and here

AlexejheroYTB:
This pull request literally doesn't do anything new, so I have taken the liberty to merge it myself.

PR #68, opened 2019-01-11
ItemsContainer Patcher

PrimeSonic:
This replaces the original `InventoryPatcher`, first created to improve performance when the player inventory started getting too full.

Now the same patches apply to _all_  `ItemContainer` instances, lending the same caching feature to every item container in the game.


SMLHelper/Patchers/ItemsContainerPatcher.cs
```diff
@@ -0,0 +1,111 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.Reflection;
+    using Harmony;
+
+    // This entire patcher is only here for performance reasons.
+    // This is not intended to be exposed by the public API.
+    internal static class ItemsContainerPatcher
+    {
+        private class StorageCache : Dictionary<Vector2int, bool>
```
ahk1221: Why make an empty class just to represent this? Wouldn't it be better to just have that stated explicitly in the main declaration of the cache?

PrimeSonic: Mostly for readability.  
But it is entirely optional and likely doesn't have any real hit to performance.

SMLHelper/Patchers/ItemsContainerPatcher.cs
```diff
@@ -0,0 +1,111 @@
...
+        private static void HasRoomFor_Postfix(ItemsContainer __instance, bool __result, ref Vector2int __state)
+        {
+            // We should only enter this method if the Prefix didn't have a cached value to use.
+            // Catch the result and map it to the size provided by the Prefix.
+            if (__state.x == 0 || __state.y == 0)
+                return; // Somehow, this can still happen. Don't store 0,0 in the cache. It breaks the game.
+
+            // The large calculation won't have to be done again until something is added or removed from the inventory.
+            // How does the base game get away with calculating this mess on every frame???
+            HasRoomCacheCollection[__instance][__state] = __result;
```
PrimeSonic: If the cache receives a `true` for anything bigger than a 1x1 slot,  
then we can also add `true` for all sizes smaller than what it just received.

For example: No need to perform the heavy calculation for a 1x1 if the cache already knows there's room for a 3x3.

SMLHelper/Patchers/ItemsContainerPatcher.cs
PrimeSonic: There. No more empty class.

SMLHelper/Patchers/ItemsContainerPatcher.cs
PrimeSonic: Added additional preemptive cache checks if we get something bigger than a 1x1.

ahk1221:
Looks good

Issue #69, opened 2019-01-11
Language Overrides for User provided translations


PrimeSonic:
Multi-language support is very limited among Subnautica mods today.  
Typically, only if the mod author directly adds multi-language support themselves will their mod have it.

This proposed feature would allow any mod built with SMLHelper to receive user provided translations and enable multi-language support with no additional work for the mod author.  

I have a proposed, multi-step process to achieving this.

> **Step 1**
> Capture all calls into the `LanguageHandler`'s `SetLanguageLine` method and store the requests into a separate data structure.
> For ease of use, separate all requests by calling assembly, so each mod's language lines are put together.
> 
> **Step 2**
> In a new folder "/Modding Helper/LanguageOriginal", write a file for each mod that supplied language entries.
> Use a format that is easy to read and edit, along with being easy serialize, and deserialize to and from a Dictionary structure.
> This will be the template or starting file that a user will copy and edit to provide transactions.
> The file should be named after the calling assembly to make it easier to match files with mods.
> 
> **Step 3**
> Inside a different folder "/Modding Helper/LanguageOverrides", look for files that match the same names as the mod assemblies.
> This file will provide a the overrides to the language entries that the mod originally provide.
> Use these entries to override everything in the `LanguagePatcher` `customLines` dictionary with the user provided lines.

When all is said and done, the end result will be that the user provided text will be used instead of the text originally provided by the mod assembly.  
This will allow users to create translation files and incorporate them into their mods without any intervention on behalf of the mod author.

PrimeSonic:
Initial feature ready for use.

While it can still be further refined into something more resembling a multi-language support option, 
that can be tackled separately.

Closing this issue.

PR #70, opened 2019-01-16
Language overrides

PrimeSonic:
From Feature Request #69 

Functions as so:
1. The `LanguagePatcher` collects all custom language strings as normal but now also organized them by mod.
2. When the `Patch` method if finally called, this data is saved to files for users to examine.
  a. A txt file for each mod's language entries will be saved to `/Modding Helper/Language/Originals/`
  b. If the corresponding txt file already exists and matches the collected data, it will not be re-written.
  c. Each file is named after the mod assembly, making them easy to identify.
3. Users can copy these original language files into `/Modding Helper/Language/Overrides/` and replace the original text provided by the mod for their own text.
4. `LanguagePatcher` will look for these files during its `Patch` method call.
  a. The file is parsed and each line verified to make sure it is correctly formatted and actually belongs to the mod in the file name.
  b. When a match is found, the user provided text will be used in place of the text provided by the mod.


ahk1221:
Looks good, nice to have this available to the users.

PR #71, opened 2019-01-19
Change mod options and add example mod

AlexejheroYTB:


PrimeSonic:
Looked good so far. I'll try this out when I get the chance.  
If you can write some sample test code for all these like we had before, it would be helpful, as well as serving as useful documentation.

SMLHelper/Options/ModOptions.cs
```diff
@@ -312,4 +427,48 @@ internal ModChoiceOption(string id, string label, string[] options, int index) :
...
+        /// <param name="id">The internal ID of this option.</param>
+        /// <param name="label">The display text to show on the in-game menus.</param>
+        /// <param name="options">The collection of available values.</param>
+        /// <param name="index">The starting value.</param>
+        internal ModDropdownOption(string id, string label, string[] options, int index) : base(ModOptionType.Dropdown, label, id)
+        {
+            Options = options;
+            Index = index;
+        }
+    }
```
PrimeSonic: This file is starting to get big, with too many barely related things contained.  
Let's split this up into distinct files for each class.

ahk1221:
Looks good. As for what's going to happen with the choices/dropdown's options, I think maybe storing objects instead of strings, and then calling ToString on said object when building the options menu would be the best idea. Maybe could work with generics as well. And if you can get that custom wrapper to work, that would be awesome, so that we can get the actual object changed/selected instead of having to manually do that.

AlexejheroYTB:
I am currently thinking about adding the following overloads to the choice and dropdown options:

    public static void AddChoiceOption(string id, string name, object[] options, int choice)
    public static void AddChoiceOption(string id, string name, object[] options, object value)
    public static void AddChoiceOption(string id, string name, string[] options, string value)
    public static void AddChoiceOption<T>(string id, string name, T value) where T : Enum
    public static void AddChoiceOption<T>(string id, string name, int choice) where T : Enum

SMLHelper/Options/ModOptions.cs
AlexejheroYTB: Addressed in commit 20c314b

PrimeSonic:
If no one else does, I can add documentation on the remaining public classes and methods that don't have any.

AlexejheroYTB:
> If no one else does, I can add documentation on the remaining public classes and methods that don't have any.

I added docs for all of the classes that I created. I didn't know there were others. I will try to find and document them myself

AlexejheroYTB:
I think that documenting public members should be done in a different pr.

AlexejheroYTB:
@ahk1221 This PR is ready to be merged IMO

PR #74, opened 2019-02-04
Sort mods options list

AlexejheroYTB:


SMLHelper/Handlers/OptionsPanelHandler.cs
```diff
@@ -16,6 +16,7 @@ public class OptionsPanelHandler
         public static void RegisterModOptions(ModOptions options)
         {
             OptionsPanelPatcher.modOptions.Add(options);
+            OptionsPanelPatcher.modOptions.Sort();
```
PrimeSonic: This is fine for now, but if the idea is to sort on insert, we might as well be using change the `modOptions`  List into a [SortedList](https://docs.microsoft.com/en-us/dotnet/api/system.collections.sortedlist?view=netframework-4.7.2) type.

And then this would look more like this
```CSharp
OptionsPanelPatcher.modOptions.Add(options.Name, options);
```
With no actual need to add the `IComparable` interface to the `ModOptions` class.

SMLHelper/Options/ModOptions.cs
```diff
@@ -72,9 +73,19 @@ public ModOptions(string name)
...
         public abstract void BuildModOptions();
+
+        /// <summary>
+        /// Compares one <see cref="ModOptions"/> object to another, based on their <see cref="Name"/>
+        /// </summary>
+        /// <param name="obj"></param>
+        /// <returns></returns>
+        public int CompareTo(object obj)
+        {
+            return Name.CompareTo(obj);
```
PrimeSonic: A `CompareTo` method that takes an `object` should do a few basic checks first, 

```CSharp
if (ReferenceEquals(this, obj) // Is the same object
    return 0;

if (obj is ModOptions other) // type safe check
    return Name.CompareTo(other.Name);
```
And then this because the catch-all case at the end
```CSharp
return Name.CompareTo(obj);

SMLHelper/Handlers/OptionsPanelHandler.cs
AlexejheroYTB: Addressed in commit 0be7232

SMLHelper/Options/ModOptions.cs
AlexejheroYTB: Addressed in commit 0be7232

PrimeSonic:
Tested and ready to merge

PR #75, opened 2019-02-06
PDAScanPatching - New Handler and Patcher

PrimeSonic:
- Can now edit number of required fragments to be scanned for a blueprint unlock
- Can also edit the time taken to scan a fragment
- Reverse lookups handled so the modder can use either the fragment techtype or the blueprint techtype

### DIFF
--- a/SMLHelper/Patchers/PdaPatcher.cs
+++ b/SMLHelper/Patchers/PdaPatcher.cs
@@ -1,0 +1,88 @@
+﻿namespace SMLHelper.V2.Patchers
+{
+    using System.Collections.Generic;
+    using System.Reflection;
+    using Harmony;
+
+    // Special thanks to Gorillazilla9 for sharing this method of fragment count patching
+    // https://github.com/Gorillazilla9/SubnauticaFragReqBoost/blob/master/PDAScannerPatcher.cs
+    internal class PdaPatcher
+    {
+        internal static Dictionary<TechType, int> FragmentCount = new Dictionary<TechType, int>();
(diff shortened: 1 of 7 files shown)


ahk1221:
Looks good to me in functionality; would like it if you would rename `Pda` to `PDA`, to better match the game's code style, and it also (in my opinion) looks better.

SMLHelper/Patchers/PdaPatcher.cs
```diff
@@ -0,0 +1,88 @@
...
+    using Harmony;
+
+    // Special thanks to Gorillazilla9 for sharing this method of fragment count patching
+    // https://github.com/Gorillazilla9/SubnauticaFragReqBoost/blob/master/PDAScannerPatcher.cs
+    internal class PDAPatcher
+    {
+        internal static Dictionary<TechType, int> FragmentCount = new Dictionary<TechType, int>();
+        internal static Dictionary<TechType, float> FragmentScanTime = new Dictionary<TechType, float>();
+
+        private static Dictionary<TechType, PDAScanner.EntryData> BlueprintToFragment = new Dictionary<TechType, PDAScanner.EntryData>();
```
PrimeSonic: All three of these `internal` collections should be `readonly`

PrimeSonic:
Branch in this current start causes this error in many locations.
 ```
NullReferenceException: Object reference not set to an instance of an object
  at KnownTech.Contains (TechType techType) [0x00000] in <filename unknown>:0
```

PR #77, opened 2019-02-19
CraftDataDupChecks - Check for duplicate keys

PrimeSonic:
- New SelfCheckingDictionary automatically checks for and excludes duplicate keys
- Created unit test project with unit tests for the SelfCheckingDictionary
- Removed AddToCustomTechData method from CraftDataHandler and updated usages. No longer needed.

### DIFF
--- a/SMLHelper.Tests/Properties/AssemblyInfo.cs
+++ b/SMLHelper.Tests/Properties/AssemblyInfo.cs
@@ -1,0 +1,20 @@
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+[assembly: AssemblyTitle("SMLHelper.Tests")]
+[assembly: AssemblyDescription("")]
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("SMLHelper.Tests")]
+[assembly: AssemblyCopyright("Copyright ©  2019")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+[assembly: ComVisible(false)]
+
+[assembly: Guid("ebc4aaf3-a13b-427e-bf99-3ae23edeb4b8")]
+
+// [assembly: AssemblyVersion("1.0.*")]
+[assembly: AssemblyVersion("1.0.0.0")]
+[assembly: AssemblyFileVersion("1.0.0.0")]
--- a/SMLHelper/Patchers/BioReactorPatcher.cs
+++ b/SMLHelper/Patchers/BioReactorPatcher.cs
@@ -2,7 +2,6 @@
 {
     using Harmony;
     using System.Collections.Generic;
-    using Utility;
 
     internal class BioReactorPatcher
     {
--- a/SMLHelper/Patchers/BioReactorPatcher.cs
+++ b/SMLHelper/Patchers/BioReactorPatcher.cs
@@ -8,7 +8,7 @@
     {
         #region Internal Fields
 
-        internal static Dictionary<TechType, float> CustomBioreactorCharges = new Dictionary<TechType, float>(TechTypeExtensions.sTechTypeComparer);
+        internal static IDictionary<TechType, float> CustomBioreactorCharges = new SelfCheckingDictionary<TechType, float>("CustomBioreactorCharges", TechTypeExtensions.sTechTypeComparer);
 
         #endregion
 
--- a/SMLHelper/Patchers/CraftDataPatcher.cs
+++ b/SMLHelper/Patchers/CraftDataPatcher.cs
@@ -9,18 +9,17 @@
     internal class CraftDataPatcher
     {
         #region Internal Fields
-
-        internal static List<TechType> DuplicateTechDataAttempts = new List<TechType>();
-        internal static Dictionary<TechType, ITechData> CustomTechData = new Dictionary<TechType, ITechData>();
-        internal static Dictionary<TechType, TechType> CustomHarvestOutputList = new Dictionary<TechType, TechType>();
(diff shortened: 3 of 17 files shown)


SMLHelper.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,20 @@
...
+[assembly: AssemblyConfiguration("")]
+[assembly: AssemblyCompany("")]
+[assembly: AssemblyProduct("SMLHelper.Tests")]
+[assembly: AssemblyCopyright("Copyright ©  2019")]
+[assembly: AssemblyTrademark("")]
+[assembly: AssemblyCulture("")]
+
+[assembly: ComVisible(false)]
+
+[assembly: Guid("ebc4aaf3-a13b-427e-bf99-3ae23edeb4b8")]
```
AlexejheroYTB: Guid is not required if ComVisible is false

SMLHelper/Patchers/BioReactorPatcher.cs
```diff
@@ -2,13 +2,12 @@
...
     using Harmony;
     using System.Collections.Generic;
-    using Utility;
 
     internal class BioReactorPatcher
     {
         #region Internal Fields
 
-        internal static Dictionary<TechType, float> CustomBioreactorCharges = new Dictionary<TechType, float>(TechTypeExtensions.sTechTypeComparer);
+        internal static IDictionary<TechType, float> CustomBioreactorCharges = new SelfCheckingDictionary<TechType, float>("CustomBioreactorCharges", TechTypeExtensions.sTechTypeComparer);
```
AlexejheroYTB: `internal static SelfCheckingDictionary<TechType, float>`

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -9,18 +9,17 @@
...
+        internal static IDictionary<TechType, int> CustomFinalCutBonusList = new SelfCheckingDictionary<TechType, int>("CustomFinalCutBonusList", TechTypeExtensions.sTechTypeComparer);
+        internal static IDictionary<TechType, Vector2int> CustomItemSizes = new SelfCheckingDictionary<TechType, Vector2int>("CustomItemSizes");
+        internal static IDictionary<TechType, EquipmentType> CustomEquipmentTypes = new SelfCheckingDictionary<TechType, EquipmentType>("CustomEquipmentTypes");
+        internal static IDictionary<TechType, QuickSlotType> CustomSlotTypes = new SelfCheckingDictionary<TechType, QuickSlotType>("CustomSlotTypes");
+        internal static IDictionary<TechType, float> CustomCraftingTimes = new SelfCheckingDictionary<TechType, float>("CustomCraftingTimes");
+        internal static IDictionary<TechType, TechType> CustomCookedCreatureList = new SelfCheckingDictionary<TechType, TechType>("CustomCookedCreatureList");
+        internal static IDictionary<TechType, CraftData.BackgroundType> CustomBackgroundTypes = new SelfCheckingDictionary<TechType, CraftData.BackgroundType>("CustomBackgroundTypes", TechTypeExtensions.sTechTypeComparer);
```
AlexejheroYTB: When defining class members, don't downgrade the variable type to an interface. Keep the class type

SMLHelper/Patchers/SelfCheckingDictionary.cs
```diff
@@ -0,0 +1,89 @@
...
+
+        public IEnumerator<KeyValuePair<K, V>> GetEnumerator() => UniqueEntries.GetEnumerator();
+
+        public bool Remove(K key) => UniqueEntries.Remove(key) | DuplicatesFound.Remove(key);
+
+        public bool Remove(KeyValuePair<K, V> item) => UniqueEntries.Remove(item.Key) | DuplicatesFound.Remove(item.Key);
+
+        public bool TryGetValue(K key, out V value) => UniqueEntries.TryGetValue(key, out value);
+
+        IEnumerator IEnumerable.GetEnumerator() => UniqueEntries.GetEnumerator();
```
AlexejheroYTB: Import the interface normally, not explicitly

SMLHelper.Tests/Properties/AssemblyInfo.cs
PrimeSonic: This is auto-generated by Visual Studio and is ultimately harmless.

SMLHelper/Patchers/BioReactorPatcher.cs
PrimeSonic: Nothing outside of the `SelfCheckingDictionary` itself needs to be concerned with what type it is.  
Given its usages, these could likely all be exposed as simple `ICollection`s instead of even the full range of options given to `IDictionary`.

All this is in keeping with **Principle of least privilege**.  
https://en.wikipedia.org/wiki/Principle_of_least_privilege

SMLHelper/Patchers/CraftDataPatcher.cs
PrimeSonic: Same as above.  


> Nothing outside of the SelfCheckingDictionary itself needs to be concerned with what type it is.
Given its usages, these could likely all be exposed as simple ICollections instead of even the full range of options given to IDictionary.

> All this is in keeping with Principle of least privilege.
> https://en.wikipedia.org/wiki/Principle_of_least_privilege

SMLHelper/Patchers/SelfCheckingDictionary.cs
PrimeSonic: The `IDictionary` interface actually demands that this is how the `GetEnumerator` method be specified.  
Anything else will result in the compiler error of not fully implementing the interface.

SMLHelper/Patchers/BioReactorPatcher.cs
PrimeSonic: Marking this as resolved. Further cleanup of this kind can and will be done over time.

PR #78, opened 2019-02-19
Add different logging levels

ahk1221:

  - Added log levels to the logging system.
  - Different log levels can be set from the configuration file.

Closes #79

### DIFF
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -5,10 +5,10 @@
 
     internal enum LogLevel
     {
-        Debug,
-        Warn,
-        Error,
-        Info
+        Debug = 0,
+        Warn = 1,
+        Error = 2,
+        Info = 3
     }
 
     internal static class Logger
@@ -21,7 +21,7 @@
 
             if(!File.Exists(configPath))
             {
-                File.AppendAllText(configPath, "Warn");
+                File.WriteAllText(configPath, "Warn");
                 LoggingLevel = LogLevel.Warn;
 
                 return;
@@ -39,8 +39,7 @@
             {
                 LoggingLevel = LogLevel.Warn;
 
-                File.Delete(configPath);
-                File.AppendAllText(configPath, "Warn");
+                File.WriteAllText(configPath, "Warn");
 
                 Log("Error reading log level configuration file. Defaulted to LogLevel.Warn.", LogLevel.Warn);
             }
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -52,13 +52,13 @@
             }
         }
 
-        internal static void Log(string text, LogLevel level = LogLevel.Error)
+        internal static void Log(string text, LogLevel level = LogLevel.Info)
         {
             if(level >= LoggingLevel)
                 Console.WriteLine($"[SMLHelper/{level.ToString()}] {text}");
         }
 
-        internal static void Log(string text, LogLevel level = LogLevel.Error, params object[] args)
+        internal static void Log(string text, LogLevel level = LogLevel.Info, params object[] args)
         {
             if (args != null && args.Length > 0)
                 text = string.Format(text, args);
@@ -67,7 +67,7 @@
                 Console.WriteLine($"[SMLHelper/{level.ToString()}] {text}");
         }
 
-        internal static void Announce(string text, LogLevel level, bool logToFile = false)
+        internal static void Announce(string text, LogLevel level = LogLevel.Info, bool logToFile = false)
         {
             ErrorMessage.AddMessage(text);
 
@@ -75,7 +75,7 @@
                 Log(text, level);
         }
 
-        internal static void Announce(string text, LogLevel level, bool logToFile = false, params object[] args)
+        internal static void Announce(string text, LogLevel level = LogLevel.Info, bool logToFile = false, params object[] args)
         {
             ErrorMessage.AddMessage(string.Format(text, args));
 
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -33,6 +33,13 @@
             {
                 LoggingLevel = (LogLevel)Enum.Parse(typeof(LogLevel), level);
 
+                if (LoggingLevel > LogLevel.Error)
+                {
+                    LoggingLevel = LogLevel.Error;
+
+                    File.WriteAllText(configPath, "Error");
+                }
+
                 Log($"Log level set to: LogLevel.{LoggingLevel.ToString()}.", LogLevel.Info);
             }
             catch (Exception)
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -1,6 +1,7 @@
 ﻿namespace SMLHelper.V2
 {
     using System;
+    using System.IO;
 
     internal enum LogLevel
     {
@@ -12,11 +13,37 @@
 
     internal static class Logger
     {
-        internal static LogLevel LoggingLevel { get; private set; } = LogLevel.Warn;
+        internal static LogLevel LoggingLevel { get; private set; }
 
-        internal static void Initialize(LogLevel level)
+        internal static void Initialize()
         {
-            LoggingLevel = level;
+            string configPath = "./QMods/Modding Helper/loglevel.txt";
+
+            if(!File.Exists(configPath))
+            {
+                File.AppendAllText(configPath, "Warn");
+                LoggingLevel = LogLevel.Warn;
+
+                return;
+            }
+
+            string level = File.ReadAllText(configPath);
+
+            try
+            {
+                LoggingLevel = (LogLevel)Enum.Parse(typeof(LogLevel), level);
+
+                Log($"Log level set to: LogLevel.{LoggingLevel.ToString()}.", LogLevel.Info);
+            }
+            catch (Exception)
+            {
+                LoggingLevel = LogLevel.Warn;
+
+                File.Delete(configPath);
+                File.AppendAllText(configPath, "Warn");
+
+                Log("Error reading log level configuration file. Defaulted to LogLevel.Warn.", LogLevel.Warn);
+            }
         }
 
         internal static void Log(string text, LogLevel level = LogLevel.Error)
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -2,35 +2,52 @@
 {
     using System;
 
+    internal enum LogLevel
+    {
+        Debug,
+        Warn,
+        Error,
+        Info
+    }
+
     internal static class Logger
     {
-        internal static void Log(string text)
+        internal static LogLevel LoggingLevel { get; private set; } = LogLevel.Warn;
+
+        internal static void Initialize(LogLevel level)
         {
-            Console.WriteLine($"[SMLHelper] {text}");
+            LoggingLevel = level;
         }
 
-        internal static void Log(string text, params object[] args)
+        internal static void Log(string text, LogLevel level = LogLevel.Error)
+        {
+            if(level >= LoggingLevel)
+                Console.WriteLine($"[SMLHelper/{level.ToString()}] {text}");
+        }
+
+        internal static void Log(string text, LogLevel level = LogLevel.Error, params object[] args)
         {
             if (args != null && args.Length > 0)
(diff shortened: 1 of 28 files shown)


SMLHelper/Logger.cs
```diff
@@ -1,36 +1,80 @@
...
-        internal static void Log(string text)
+        internal static LogLevel LoggingLevel { get; private set; }
+
+        internal static void Initialize()
+        {
+            string configPath = "./QMods/Modding Helper/loglevel.txt";
+
+            if(!File.Exists(configPath))
+            {
+                File.AppendAllText(configPath, "Warn");
```
AlexejheroYTB: `File.WriteAllText` when creating new files

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,80 @@
...
 {
     using System;
+    using System.IO;
+
+    internal enum LogLevel
+    {
+        Debug,
+        Warn,
+        Error,
+        Info
```
AlexejheroYTB: The order should be Debug, Info, Warn, Error

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,80 @@
...
+
+                Log("Error reading log level configuration file. Defaulted to LogLevel.Warn.", LogLevel.Warn);
+            }
+        }
+
+        internal static void Log(string text, LogLevel level = LogLevel.Error)
         {
-            Console.WriteLine($"[SMLHelper] {text}");
+            if(level >= LoggingLevel)
+                Console.WriteLine($"[SMLHelper/{level.ToString()}] {text}");
```
AlexejheroYTB: Maybe also check for invalid log ids before calling To String

SMLHelper/Logger.cs
AlexejheroYTB: If the order of the elements in the enum matters, usually the values should be expressed explicitly
```cs
Debug = 0,
Info = 1,
Warn = 2,
Error = 3,
```

ahk1221: Info should always be displayed. In order to do that, it should be last.

SMLHelper/Logger.cs
ahk1221: Will update.

SMLHelper/Logger.cs
ahk1221: You really can't have invalid log ids.

SMLHelper/Logger.cs
ahk1221: They can be, but its not required.

AlexejheroYTB: On that same note, Errors should always appear too. I am resolving this conversation for now,  I will add another review regarding this.

ahk1221: Hmm, that is very true, perhaps if anything below Error is set, it can be set back to Error?

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,86 @@
...
```
PrimeSonic: It would make things easier to read if we also add a new set of methods to the `Logger` class.
```CSharp
internal static void Debug(string text) => Log(text, LogLevel.Debug);  
internal static void Warn(string text) => Log(text, LogLevel.Warn);  
internal static void Error(string text) => Log(text, LogLevel.Error);  
internal static void Info(string text) => Log(text, LogLevel.Info);
```
If you also want to update the current method calls you've updated to use these methods, you're welcome to.  
For now, just having the option is enough.

SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -29,7 +29,6 @@ private void Update()
 
             if (transform.position == new Vector3(-5000, -5000, -5000) && gameObject != prefab && Time.time > time)
             {
-                Logger.Log("Destroying object: " + gameObject);
```
PrimeSonic: Keep this as a `Debug` log.

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -75,28 +75,28 @@ internal static void RemoveFromCustomGroup(TechGroup group, TechCategory categor
...
         }
 
         internal static void AddToCustomTechData(TechType techType, ITechData techData)
         {
             if (CustomTechData.ContainsKey(techType))
             {
-                Logger.Log($"[ERROR] Custom TechData already exists for '{techType}'. {Environment.NewLine}" +
-                            "All entries will be removed so conflict can be noted and resolved.");
+                Logger.Log($"Custom TechData already exists for '{techType}'. {Environment.NewLine}" +
+                            "All entries will be removed so conflict can be noted and resolved.", LogLevel.Error);
```
PrimeSonic: Let's demote this to be a `Warn` since we should be able to recover from this error.  
Also, this will have to later be reconciled with the changes in #77

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -149,7 +149,7 @@ private static void AddCustomTechDataToOriginalDictionary()
             {
                 Logger.Log($"Removing conflicting TechData entries from patching.{Environment.NewLine}" +
                            $"This is so the user will take notice and know to resolve the conflict.{Environment.NewLine}" +
-                           "Only one custom TechData may be added per TechType.");
+                           "Only one custom TechData may be added per TechType.", LogLevel.Debug);
```
PrimeSonic: Let's promote this to be a `Warn` since we should be able to recover from this error.
Also, this will have to later be reconciled with the changes in #77

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -221,7 +221,7 @@ private static void AddCustomTechDataToOriginalDictionary()
                 if (techDataExists)
                 {
                     techData_Remove.Invoke(techData, new object[] { techType });
-                    Logger.Log($"{techType} TechType already existed in the CraftData.techData dictionary. Original value was replaced.");
+                    Logger.Log($"{techType} TechType already existed in the CraftData.techData dictionary. Original value was replaced.", LogLevel.Debug);
```
PrimeSonic: Let's promote this to a `Warn`.  
Also, this will have to later be reconciled with the changes in #77

SMLHelper/Patchers/CraftTreePatcher.cs
```diff
@@ -153,7 +153,6 @@ private static void AddCustomTabs(ref CraftNode nodes, List<TabNode> customTabs,
                 for (int i = 0; i < tab.Path.Length; i++)
                 {
                     string currentPath = tab.Path[i];
-                    Logger.Log("Tab Current Path: " + currentPath + " Tab: " + tab.Name + " Crafter: " + tab.Scheme.ToString());
```
PrimeSonic: Let's keep this as a `Debug` log.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -68,7 +68,7 @@ private static void WriteOriginalCustomLines()
                 filesWritten++;
             }
 
-            Logger.Log($"Updated {filesWritten} of {originalCustomLines.Count} original language files.");
+            Logger.Log($"Updated {filesWritten} of {originalCustomLines.Count} original language files.", LogLevel.Info);
```
PrimeSonic: Let's limit this log entry to only when
```CSharp
If (filesWritten > 0)

File: SMLHelper/Patchers/LanguagePatcher.cs [hunk 11]
```diff
@@ -118,23 +118,23 @@ private static void ReadOverrideCustomLines()
 
                     if (split.Length != 2)
                     {
-                        Logger.Log($"Line '{lineIndex}' in language override file for '{modName}' was not correctly formatted.");
+                        Logger.Log($"Line '{lineIndex}' in language override file for '{modName}' was not correctly formatted.", LogLevel.Error);
```
PrimeSonic: We can demote this to a `Warn`

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -118,23 +118,23 @@ private static void ReadOverrideCustomLines()
...
                         continue; // Not correctly formatter
                     }
 
                     string key = split[0];
 
                     if (!originalLines.ContainsKey(key))
                     {
-                        Logger.Log($"Key '{key}' on line '{lineIndex}' in language override file for '{modName}' did not match an original key.");
+                        Logger.Log($"Key '{key}' on line '{lineIndex}' in language override file for '{modName}' did not match an original key.", LogLevel.Error);
```
PrimeSonic: We can also demote this to a `Warn`

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,86 @@
...
+
+            try
+            {
+                LoggingLevel = (LogLevel)Enum.Parse(typeof(LogLevel), level);
+
+                if (LoggingLevel > LogLevel.Error)
+                {
+                    LoggingLevel = LogLevel.Error;
+
+                    File.WriteAllText(configPath, "Error");
```
AlexejheroYTB: Maybe also log this action?

SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -131,9 +131,9 @@ internal void SaveCache()
             }
             catch (Exception exception)
             {
-                Logger.Log("Caught exception when saving cache!");
-                Logger.Log("Exception message: " + exception.Message);
-                Logger.Log("StackTrace: " + Environment.NewLine + exception.StackTrace);
+                Logger.Log("Caught exception when saving cache!", LogLevel.Error);
```
PrimeSonic: Can be replaced with a single log entry.
```CSharp
Logger.Log($"Caught exception when saving cache!{Environment.NewLine}{exception.}", LogLevel.Error);
// ToString() is called implicitly when in an interpolated string
```
https://docs.microsoft.com/en-us/dotnet/api/system.exception.tostring?view=netframework-4.7.2
> ToString returns a representation of the current exception that is intended to be understood by humans. Where the exception contains culture-sensitive data, the string representation returned by ToString is required to take into account the current system culture. Although there are no exact requirements for the format of the returned string, it should attempt to reflect the value of the object as perceived by the user.

SMLHelper/Utility/ExtBannedIdManager.cs
```diff
@@ -135,7 +135,7 @@ private static void LogBadEntry(string filePath, string line)
             Logger.Log($"Badly formatted entry for Retricted IDs{Environment.NewLine}" +
                        $"        File: '{filePath}{Environment.NewLine}'" +
                        $"        Line: '{line}'{Environment.NewLine}" +
-                       $"        This entry has been skipped.");
+                       $"        This entry has been skipped.", LogLevel.Error);
```
PrimeSonic: Demote to `Warn`

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,86 @@
...
             if (args != null && args.Length > 0)
                 text = string.Format(text, args);
 
-            Console.WriteLine($"[SMLHelper] {text}");
+            if(level >= LoggingLevel)
+                Console.WriteLine($"[SMLHelper/{level.ToString()}] {text}");
         }
 
-        internal static void Announce(string text, bool logToFile = false)
+        internal static void Announce(string text, LogLevel level, bool logToFile = false)
```
PrimeSonic: Would `LogLevel.Info` be a good default log level for this method?

SMLHelper/Logger.cs
```diff
@@ -1,36 +1,86 @@
...
         {
             ErrorMessage.AddMessage(text);
 
             if (logToFile)
-                Log(text);
+                Log(text, level);
         }
 
-        internal static void Announce(string text, bool logToFile = false, params object[] args)
+        internal static void Announce(string text, LogLevel level, bool logToFile = false, params object[] args)
```
PrimeSonic: Would `LogLevel.Info` be a good default log level for this method?

SMLHelper/Logger.cs
PrimeSonic: This if-block looks to be disallowing users from setting their log level to `Info`, and forcing `Error` as the minimum log level.  
If that was intentional, leave a comment (or log) to inform what happened and why.

PR #79, opened 2019-02-19
My take on the log levels

AlexejheroYTB:
This is my take on the log levels pr @ahk1221 made (#78)

- Infos, Warns and Errors are always shown, only Debugs can be disabled
- Changed Debug/Info/Warn/Error in the LogLevel.txt file to Important/All

Merging #78 will automatically close this PR if it's not accepted.

- There are improvements to be made

### DIFF
--- a/SMLHelper/Logger.cs
+++ b/SMLHelper/Logger.cs
@@ -16,20 +16,19 @@
         internal static bool EnableDebugging { get; private set; }
         internal static void SetDebugging(bool value)
         {
-            string configPath = "./QMods/Modding Helper/LogLevel.txt";
+            string configPath = "./QMods/Modding Helper/EnableDebugLogs.txt";
 
-            File.WriteAllText(configPath, value ? "All" : "Important");
-
+            File.WriteAllText(configPath, value.ToString());
             EnableDebugging = value;
         }
 
         internal static void Initialize()
         {
-            string configPath = "./QMods/Modding Helper/LogLevel.txt";
+            string configPath = "./QMods/Modding Helper/EnableDebugLogs.txt";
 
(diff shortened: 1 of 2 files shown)


PrimeSonic:
If the choice that the user has is whether or not to enable debug logging or not, then rather than use arbitrary words like "All" and "Important", call it by what it is.  
"Important" to whom exactly? What am I missing if I don't have "All" logs enabled?

This language can be simplified into something like a simple boolean config choice, without additional qualifiers that might confuse some users.

Some options
`EnableDebugLogs: TRUE|FALSE`  `EnableDebugLogs: YES|NO`  
`DebugLogs:ON|OFF` `DebugLogs:ENABLED|DISABLED`  
`LogsForDevelopers:ENABLE|DISABLE` `DeveloperLogs:ENABLED|DISABLED` `TurnOnDeveloperLogs:YES|NO`

AlexejheroYTB:
> If the choice that the user has is whether or not to enable debug logging or not, then rather than use arbitrary words like "All" and "Important", call it by what it is.
> "Important" to whom exactly? What am I missing if I don't have "All" logs enabled?
> 
> This language can be simplified into something like a simple boolean config choice, without additional qualifiers that might confuse some users.
> 
> Some options
> `EnableDebugLogs: TRUE|FALSE` `EnableDebugLogs: YES|NO`
> `DebugLogs:ON|OFF` `DebugLogs:ENABLED|DISABLED`
> `LogsForDevelopers:ENABLE|DISABLE` `DeveloperLogs:ENABLED|DISABLED` `TurnOnDeveloperLogs:YES|NO`

Addressed in commits 8be198f and e3f67ce

PR #80, opened 2019-02-25
SMLHelper 2.2 Update

PrimeSonic:
# Changes made in SMLHelper 2.2

## New Logging
  - #78|#79 ~ New logging levels will greatly minimize logs when not needed
    - Debug logging (for development and troubleshooting) can be enabled by changing the value in the new  file `EnableDebugLogs.txt`.
    - Debug logging can also be toggled from the in-game Mods menu, however, since most of the SMLHelper actions (and thus logs) take place before the main menu loads, this will mostly only be useful for subsequent loads of the game.
  - Be default, debug logging will be turned off. Only logs that require some attention from the user will be displayed
    - WARN: There was an issue that we were able to catch and compensate for.  
The game may continue to run normally but the issue should be looked at.
    - ERROR: There was an issue that we just barely managed to catch but likely we cannot compensate for it.  
The game will likely not run normally and the issue should be looked at.

## Enhanced duplicate error checks
- #77 ~ Greatly minimizes game breaking errors when different mods both request a change on the same item.
- These conflicts between mods must ultimately be resolved by the user and mod author;  
Either the user removes one of the mods that conflicts with another or, if the mod author has made the mod configurable, disabling the feature that causes the conflict.

## Updated to Harmony 1.2.0.1
- 9af087f7a9c8911bac4d47e014309802fd15954f
- Newer version of Harmony with updated internal handling and new features


SMLHelper/Properties/AssemblyInfo.cs
```diff
@@ -32,5 +32,5 @@
 // You can specify all the values or you can default the Build and Revision Numbers
 // by using the '*' as shown below:
 // [assembly: AssemblyVersion("1.0.*")]
-[assembly: AssemblyVersion("2.1.0.0")]
-[assembly: AssemblyFileVersion("2.1.0.0")]
+[assembly: AssemblyVersion("2.1.1.0")]
```
PrimeSonic: Pending: Update version on **both**  `AssemblyInfo` and `mod.json`.
Must be done before release.

PrimeSonic:
[SMLHelper_v2-2.zip](https://github.com/SMLHelper/SMLHelper/files/2907322/SMLHelper_v2-2.zip)
Release package for version 2.2

PrimeSonic:
Testing has been successful. The new log levels come in handy.

PR #81, opened 2019-03-02
Address reviews from #78

AlexejheroYTB:


PrimeSonic:
Tested the fix for the logging levels. They finally work correctly!

PR #82, opened 2019-03-03
Use publicized assemblies as dependencies to avoid most reflection calls

PrimeSonic:
- Now using the publicized versions of Assembly-CSharp and firstpass
- Greatly reduced the amount of necessary Reflection as many private fields and types can now be accessed directly

Made possible thanks to https://github.com/CabbageCrow/AssemblyPublicizer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -261,3 +261,5 @@ __pycache__/
 *.pyc
 Dependencies/Assembly-CSharp-firstpass.dll
 Dependencies/Assembly-CSharp.dll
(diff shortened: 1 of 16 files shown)


ahk1221:
Looking good, however I would like it if there was some sort of readme to tell future contributors what they need to setup their own development environment.

PrimeSonic:
@ahk1221 the `README`.md` file, as well as all source code files likely to show errors during a build (when the publicized assemblies are not present), have been updated to direct new contributors going forward.

PR #83, opened 2019-03-04
2.2.1 Update

PrimeSonic:
  - #82 Using publicized assemblies to minimize Reflection calls during patching
  - #81 Fixing bugs in debug logging levels


PrimeSonic:
I'll release this later tonight if no one else does sooner.

PR #85, opened 2019-03-07
Split item storage processing

DingoDjango:
Split ItemsContainerPatcher.cs into three parts:
  - Harmony patches
  - Helper class
  - Interface

### DIFF
--- a/SMLHelper/Patchers/ItemsContainerPatcher.cs
+++ b/SMLHelper/Patchers/ItemsContainerPatcher.cs
@@ -57,7 +57,9 @@
 
 			__state = new Vector2int(width, height);
 
-			return ItemStorageHelper.Helper.TryGetCachedHasRoom(__instance, __state, ref __result);
+			// If there is no cached result (false), return true and continue with code execution
+			// If a result exists (true), detour the rest of the method and return the cached result
+			return !ItemStorageHelper.Helper.TryGetCachedHasRoom(__instance, __state, ref __result);
 		}
 
 		private static void HasRoomFor_Postfix(ItemsContainer __instance, bool __result, Vector2int __state)
--- a/SMLHelper/Patchers/ItemsContainerPatcher.cs
+++ b/SMLHelper/Patchers/ItemsContainerPatcher.cs
@@ -1,174 +1,76 @@
 ﻿namespace SMLHelper.V2.Patchers
 {
-    using System;
-    using System.Collections.Generic;
-    using System.Reflection;
-    using Harmony;
+	using System;
+	using System.Reflection;
+	using Harmony;
+	using SMLHelper.V2.Utility;
 
-    // This entire patcher is only here for performance reasons.
-    // This is not intended to be exposed by the public API.
-    internal static class ItemsContainerPatcher
-    {
-        private static Dictionary<IItemsContainer, Dictionary<Vector2int, bool>> HasRoomCacheCollection = new Dictionary<IItemsContainer, Dictionary<Vector2int, bool>>();
+	internal static class ItemsContainerPatcher
+	{
+		internal static void Patch(HarmonyInstance harmony)
+		{
+			// Original methods
+			Type itemsContainerType = typeof(ItemsContainer); // Vanilla Subnautica class for all item storage containers
+			MethodInfo HasRoomFor_XY_Method = itemsContainerType.GetMethod(nameof(ItemsContainer.HasRoomFor), new Type[] { typeof(int), typeof(int) });
+			MethodInfo NotifyAddItem_Method = itemsContainerType.GetMethod("NotifyAddItem", BindingFlags.NonPublic | BindingFlags.Instance); // Private void
+			MethodInfo NotifyRemoveItem_Method = itemsContainerType.GetMethod("NotifyRemoveItem", BindingFlags.NonPublic | BindingFlags.Instance); // Private void
 
-        internal static void Patch(HarmonyInstance harmony)
-        {
-            // Original methods
-            Type itemsContainerType = typeof(ItemsContainer); // The class for all item storages
-            MethodInfo HasRoomFor_XY_Method = itemsContainerType.GetMethod("HasRoomFor", new Type[] { typeof(int), typeof(int) });
-            MethodInfo NotifyAddItem_Method = itemsContainerType.GetMethod("NotifyAddItem", BindingFlags.NonPublic | BindingFlags.Instance);
-            MethodInfo NotifyRemoveItem_Method = itemsContainerType.GetMethod("NotifyRemoveItem", BindingFlags.NonPublic | BindingFlags.Instance);
+			// Harmony methods
+			Type patcherType = typeof(ItemsContainerPatcher);
+			MethodInfo HasRoomFor_XY_Prefix_Method = patcherType.GetMethod(nameof(ItemsContainerPatcher.HasRoomFor_XY_Prefix), BindingFlags.NonPublic | BindingFlags.Static);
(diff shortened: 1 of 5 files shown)


SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+
+			return this.Just1x1;
+		}
+
+		public void CacheNewContainer(IItemsContainer container)
+		{
+			this.HasRoomCacheCollection.Add(container, new Dictionary<Vector2int, bool>());
+		}
+
+		public void RecacheContainer(IItemsContainer container)
```
PrimeSonic: This method also doesn't seem to need to be public.

SMLHelper/Utility/IStorageHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace SMLHelper.V2.Utility
+{
+	interface IStorageHelper
+	{
+		void CacheNewContainer(IItemsContainer container);
+
+		void RecacheContainer(IItemsContainer container);
+
+		void CacheNewHasRoomData(IItemsContainer container, Vector2int itemSize, bool hasRoom);
```
PrimeSonic:
```CSharp
void CacheNewContainer(IItemsContainer container);
void RecacheContainer(IItemsContainer container);
void CacheNewHasRoomData(IItemsContainer container, Vector2int itemSize, bool hasRoom);
```
These three methods don't appear to serve a purpose externally.  
If other mods were to call these, we could end up in a very strange state.

SMLHelper/SMLHelper.csproj
```diff
@@ -42,10 +42,10 @@
       <HintPath>..\Dependencies\0Harmony-1.2.0.1.dll</HintPath>
     </Reference>
     <Reference Include="Assembly-CSharp">
-      <HintPath>..\Dependencies\Assembly-CSharp.dll</HintPath>
+      <HintPath>C:\Program Files (x86)\Steam\steamapps\common\Subnautica\Subnautica_Data\Managed\Assembly-CSharp.dll</HintPath>
     </Reference>
     <Reference Include="Assembly-CSharp-firstpass">
-      <HintPath>..\Dependencies\Assembly-CSharp-firstpass.dll</HintPath>
+      <HintPath>C:\Program Files (x86)\Steam\steamapps\common\Subnautica\Subnautica_Data\Managed\Assembly-CSharp-firstpass.dll</HintPath>
```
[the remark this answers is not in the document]
DingoDjango: Fixed in the next commit.

SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,124 @@
+﻿namespace SMLHelper.V2.Utility
+{
+	using System.Collections.Generic;
+
+	public class ItemStorageHelper : IStorageHelper
+	{
+		private readonly Dictionary<IItemsContainer, Dictionary<Vector2int, bool>> HasRoomCacheCollection = new Dictionary<IItemsContainer, Dictionary<Vector2int, bool>>();
+
+		public static ItemStorageHelper helper;
```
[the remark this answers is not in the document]
DingoDjango: 1. Can be achieved with `public static readonly ItemStorageHelper Helper = new ItemStorageHelper();`
2. What is the preferred method for doing this?
3. Expand on that please. If the methods are internal then this is literally true, isn't it?

SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+
+			if (original.x == 2 && original.y == 2) // Size of common game items
+			{
+				return this.SmallerThan2x2;
+			}
+
+			return this.Just1x1;
+		}
+
+		public void CacheNewContainer(IItemsContainer container)
```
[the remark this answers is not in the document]
DingoDjango: I think it's best we do an extension method class for this. Even more ease of use.

SMLHelper/Patchers/ItemsContainerPatcher.cs
```diff
@@ -1,174 +1,76 @@
...
-            {
-                return SmallerThan2x2;
-            }
-
-            // 1x2 is the size of a Creature Decoy
-
-            return Just1x1;
-        }
-    }
+	using System;
```
AlexejheroYTB: The indentation triggers my OCD. Please use spaces instead of tabs

SMLHelper/Utility/IStorageHelper.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace SMLHelper.V2.Utility
+{
+	interface IStorageHelper
```
AlexejheroYTB: Should this be public?

DingoDjango:
If a mod calls `container.TryGetCachedHasRoom` and there is no cache, should we run our own HasRoom calculation?
Right now we are only caching when called from Harmony (because the game runs the calculation). If a mod calls `TryGet` it will return false without caching anything.

SMLHelper/Patchers/ItemsContainerPatcher.cs
PrimeSonic: I'll clean this up later on another branch.

SMLHelper/Utility/ItemStorageHelper.cs
```diff
@@ -0,0 +1,124 @@
...
+			if (hasRoom && (itemSize.x > 1 || itemSize.y > 1))
+			{
+				foreach (Vector2int size in this.CommonSmallerSizes(itemSize))
+				{
+					this.HasRoomCacheCollection[container][size] = true;
+				}
+			}
+		}
+
+		public bool TryGetCachedHasRoom(IItemsContainer container, int width, int height, ref bool cachedResult)
```
PrimeSonic: @DingoDjango said
> If a mod calls `container.TryGetCachedHasRoom` and there is no cache, should we run our own HasRoom calculation?
> Right now we are only caching when called from Harmony (because the game runs the calculation). If a mod calls `TryGet` it will return false without caching anything.
That is indeed the original behavior.  
This whole patch was created under the expectation that the only time these methods are being called is because there is a game event requesting it.  
Now we open the possibility of mods to start checking these methods outside of the normal game loops.

At this point, a decision should be made.  
In my opinion, because the mod already has access to the `IItemsContainer` it wants to deal with,  
when it gets a `false` return value from `TryGetCachedHasRoom` it can simply go straight into the original `HasRoomFor`, with the benefit of the patches here caching the value.

It seems bit redundant, but when the data just isn't available, there isn't much else to do.

DingoDjango: So just to be clear, are we going to fall back to calling the original `HasRoomFor` or do we leave that up to others mods?
My thinking was make a `TryGetCachedHasRoomInternal` for the Harmony patch, and other `TryGet` methods will be simplified and default to calling `ItemsContainer.HasRoom` if they don't get a result right away.

DingoDjango: If the `IItemsContainer` in question is not `ItemsContainer` then we still return false, of course.

PrimeSonic: It shouldn't matter what actually call is behind the `IItemsContainer` interface.  
This interface has all the methods we ever need to interact with.

As for what a mod will do when it gets a `false` out of the `TryGetCachedHasRoom` that seems best left to the external caller to decide.

SMLHelper/Utility/ItemStorageHelper.cs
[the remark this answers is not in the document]
DingoDjango: Addressed in latest updates as extension methods

SMLHelper/Utility/ItemStorageHelper.cs
DingoDjango: Addressed as we have shifted to only addressing `ItemsContainer`

PR #86, opened 2019-03-10
ItemAction - Patcher and Handler for Middle Click Actions

AlexejheroYTB:


SMLHelper/Patchers/ItemActionPatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+using System.Reflection;
+using System.Text;
+
+namespace SMLHelper.V2.Patchers
+{
+    public class ItemActionPatcher
+    {
+        #region Internal Fields
+
+        internal const ItemAction MiddleClickItemAction = (ItemAction)1337;
```
PrimeSonic: Can we leave a comment regarding with 1337 is?

SMLHelper/Handlers/ItemActionHandler.cs
```diff
@@ -0,0 +1,23 @@
...
+        /// <summary>
+        /// Registers a middle click action for an <see cref="InventoryItem"/> based off of its <see cref="TechType"/>
+        /// </summary>
+        /// <param name="targetTechType">The <see cref="TechType"/> which the middle click action will be assigned to</param>
+        /// <param name="callback">The method which is called when the <see cref="InventoryItem"/> was middle-clicked</param>
+        /// <param name="tooltip">The secondary tooltip which appears in the description of the item</param>
+        public static void RegisterMiddleClickAction(TechType targetTechType, Action<InventoryItem> callback, string tooltip = null)
+        {
+            ItemActionPatcher.CustomItemActions.Add(targetTechType, callback);
+            if (!string.IsNullOrEmpty(tooltip)) ItemActionPatcher.CustomItemActionTooltips.Add(targetTechType, tooltip);
```
PrimeSonic: The tooltip value should be routed through the LanguagePatcher, if possible, so it can over handled with the language overrides.

SMLHelper/Patchers/ItemActionPatcher.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            Type InventoryType = typeof(Inventory);
+            MethodInfo ExecuteItemActionMethod = InventoryType.GetMethod("ExecuteItemAction", BindingFlags.NonPublic | BindingFlags.Instance);
+            harmony.Patch(ExecuteItemActionMethod, new HarmonyMethod(thisType.GetMethod("ExecuteItemAction_Prefix", BindingFlags.NonPublic | BindingFlags.Static)));
+
+            Type TooltipFactoryType = typeof(TooltipFactory);
+            MethodInfo ItemActionsMethod = TooltipFactoryType.GetMethod("ItemActions", BindingFlags.NonPublic | BindingFlags.Static);
+            harmony.Patch(ItemActionsMethod, null, new HarmonyMethod(thisType.GetMethod("ItemActions_Postfix", BindingFlags.NonPublic | BindingFlags.Static)));
+
+            Logger.Log("ItemActionPatcher is done.", LogLevel.Debug);
```
PrimeSonic: We should include an Info log here that logs how many entries were patched.  
And only write this log if we have at least 1 entry.

SMLHelper/Patchers/ItemActionPatcher.cs
AlexejheroYTB: A custom constant value is required to differenciate between normal `ItemAction`s and the custom middle click one. This is checked in another patch. (I know that this is not the best way to handle this)

PrimeSonic: When I said leave a comment, I mean in the source code.
Either a comment or `<summary>` documentation.

We don't need to change this, only leave notes so that we aren't left wondering why we did this months later.

SMLHelper/Handlers/ItemActionHandler.cs
PrimeSonic: Recommendation:  
Use `LanguageHandler.SetLanguageLine($"MCAction_{targetTechType}", tooltip);` to pass the tooltip to the Language Helper

Then, use `(Language.main.Get($"MCAction_{targetTechType}")` to retrieve the patched language line in the Harmony patch method.

AlexejheroYTB:
I also want to add configurable left click actions, but I first want to find how to check if an item already has one

SMLHelper/Patchers/ItemActionPatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+        internal class CustomItemAction
+        {
+            internal Action<InventoryItem> Action;
+            internal string LanguageLine;
+            internal Predicate<InventoryItem> Condition;
+
+            internal CustomItemAction(Action<InventoryItem> action, string tooltip, Predicate<InventoryItem> condition)
+            {
+                Action = action;
+                LanguageLine = tooltip;
```
PrimeSonic: For clarity, we might want to rename this internal field to be `LanguageLineKey` and the incoming parameter to be `tooltipKey`, since this is how they are being used with the Language lines.

PR #87, opened 2019-03-16
Interfaces overhaul

PrimeSonic:
## Interface Overhaul
- All public handler classes are now singletons
- Static methods have been preserved for backwards compatibility
- All handler classes now have a corresponding interface that defines the available methods
- This will allow other modders to apply dependency injection to SMLHelper if they choose to

## ItemStorageHelper
- The ItemsContainerPatcher has been refactored
- All methods that could be made public for mods to consume are exposed through a singleton class like the handlers do
- Additional methods were added to assist modders who want to access some additional info about a container without Reflection

## BugFixes
- We now correctly get the calling mod assembly to all methods where this is needed for the LanguageOverride handling

**Changes are fully backwards compatible**


DingoDjango:
ItemContainer part looks good to me.

PR #88, opened 2019-03-19
Integration of FishFramework

kylinator25:
Initial merging of the Fish Framework into SMLHelper. Let me know if you have any concerns regarding where I've put it in the repo/where I initialized the framework from SMLHelper.


SMLHelper/FishFramework/Extensions.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+
+namespace SMLHelper.V2.FishFramework
+{
+    public static class Extensions
+    {
+        public static T AddOrGet<T>(this GameObject obj) where T : Component
```
AlexejheroYTB: This could probably be implemented in Utilities so other stuff can use it. Also I would suggest renaming it to GetOrAddComponent, since it gets before it adds, and it's about components

SMLHelper/FishFramework/CustomFish.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+using SMLHelper.V2.Handlers;
+
+namespace SMLHelper.V2.FishFramework
+{
+    public class CustomFish
```
AlexejheroYTB: This could become FishHandler

SMLHelper/FishFramework/CustomFishPrefab.cs
```diff
@@ -0,0 +1,133 @@
...
+using SMLHelper.V2.Assets;
+using UnityEngine;
+
+namespace SMLHelper.V2.FishFramework
+{
+    /// <summary>
+    /// Class used by CustomFish for constructing a prefab based on the values provided by the user
+    /// You can use this yourself if you want, but you will need to manually provide a TechType
+    /// </summary>
+    public class CustomFishPrefab : ModPrefab
```
AlexejheroYTB: This could be moved into Assets

SMLHelper/FishFramework/FishSpawner.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+using Harmony;
+using UnityEngine;
+using Random = UnityEngine.Random;
+using System.Reflection;
+
+namespace SMLHelper.V2.FishFramework
+{
+    [HarmonyPatch(typeof(Creature), "Start")]
+    public class FishSpawner
```
AlexejheroYTB: This could become FishPatcher

SMLHelper/FishFramework/MainInit.cs
```diff
@@ -0,0 +1,18 @@
...
+using System.Reflection;
+
+namespace SMLHelper.V2.FishFramework
+{
+    public class MainInit
+    {
+        public static void Init()
+        {
+            HarmonyInstance harm = HarmonyInstance.Create("Kylinator25-FishFramework");
+            harm.PatchAll(Assembly.GetExecutingAssembly());
```
AlexejheroYTB: This can be removed

SMLHelper/FishFramework/ScaleFixer.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+
+namespace SMLHelper.V2.FishFramework
+{
+    public class ScaleFixer : MonoBehaviour
```
AlexejheroYTB: This doesnt have to be a separate component and could be merged with the base one

SMLHelper/Initializer.cs
```diff
@@ -21,6 +22,7 @@ public static void Patch()
             {
                 InitializeOld(); // Some patch methods add values/call methods to V2 patchers, and so they need to called first.
                 Initialize();
+                MainInit.Init(); // <- Initialize FishFramework
```
AlexejheroYTB: Merge this with Initialize()

SMLHelper/FishFramework/CustomFish.cs
```diff
@@ -0,0 +1,80 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using UnityEngine;
+using SMLHelper.V2.Handlers;
```
PrimeSonic: General cleanup for all files:  
Remove any unused `using` statements.  
Move any remaining  `using` statements inside the namespace.

SMLHelper/FishFramework/ScaleFixer.cs
kylinator25: Ok i've renamed the component to be more generic, more functionality will probably be added later on

SMLHelper/FishFramework/CustomFish.cs
kylinator25: I kept this class, and created a new class which will basically do what the Register() function does but externally, in keeping with SMLHelper's style

PrimeSonic: I agree with this approach.

SMLHelper/Utility/PrefabUtils.cs
```diff
@@ -18,5 +18,16 @@ public static void AddBasicComponents(ref GameObject _object, string classId)
             var forces = _object.AddComponent<WorldForces>();
             forces.useRigidbody = rb;
         }
+
+        public static T GetOrAddComponent<T>(this GameObject obj) where T : Component
```
PrimeSonic: I like this addition. Very nicely done.

SMLHelper/Patchers/CustomFishPatcher.cs
```diff
@@ -0,0 +1,68 @@
...
+            if(tag)
+            {
+                if(CustomFishHandler.fishTechTypes.Contains(tag.type))
+                {
+                    return;
+                }
+            }
+            if(Random.value < 0.1f)
+            {
+                Console.WriteLine($"[FishFramework] Selecting fish out of {CustomFishHandler.fishTechTypes.Count} total types");
```
PrimeSonic: Can these Console writes be updated to use our [Logger](https://github.com/SMLHelper/SMLHelper/blob/2.x/SMLHelper/Logger.cs) class?  
Regarding the logging levels:  
- Info = "Always printed"
- Warn = "Only printed for recoverable errors that _may_ result in unintended behavior"
- Error = "Only printed for critical errors that _absolutely_ will result in unintended behavior"
- Debug = "Only printed when debug logging in enabled. Used for things like the normal execution of patches."

SMLHelper/MonoBehaviours/CustomCreature.cs
```diff
@@ -0,0 +1,15 @@
...
+    using UnityEngine;
+
+    public class CustomCreature : MonoBehaviour
+    {
+        public float scale;
+
+        void Update()
+        {
+	     // At the moment all this does is fix the scale, it will likely have more functionality later
+            transform.localScale = new Vector3(scale, scale, scale);
```
PrimeSonic: Is this this necessary to run on every update call?  
Is there any performance penalty to doing so?

SMLHelper/MonoBehaviours/CustomCreature.cs
```diff
@@ -0,0 +1,15 @@
+﻿namespace SMLHelper.V2.MonoBehaviours
+{
+    using UnityEngine;
+
+    public class CustomCreature : MonoBehaviour
```
PrimeSonic: If the class is _not_ intended to be manipulated by other assemblies, make it `internal`.

SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,33 @@
+﻿namespace SMLHelper.V2.Handlers
+{
+    using System;
+    using System.Collections.Generic;
+    using SMLHelper.V2.FishFramework;
+    using SMLHelper.V2.Assets;
+
+    public static class CustomFishHandler
```
PrimeSonic: As another standard practice for anything that is to be used as a public API, make sure to include XML documentation for all public classes, methods, fields, and properties.

SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,132 @@
...
+
+        public CustomFishPrefab(string classId, string prefabFileName, TechType techType = TechType.None) : base(classId, prefabFileName, techType)
+        {
+        }
+
+        public override GameObject GetGameObject()
+        {
+            if (cachedPrefab)
+            {
+                return cachedPrefab;
```
PrimeSonic: Please explain the benefits of this `cachedPrefab` pattern.  
If it's that good, then maybe this pattern should just be added directly to [ModPrefab](https://github.com/SMLHelper/SMLHelper/blob/2.x/SMLHelper/Assets/ModPrefab.cs) itself.

SMLHelper/Initializer.cs
```diff
@@ -46,6 +47,7 @@ internal static void InitializeOld()
 
         internal static void Initialize()
         {
+            CustomFishPatcher.Patch(harmony);
```
PrimeSonic: Is there any risk in patching the custom fish _before_ patching the `TechTypes`?

SMLHelper/Assets/CustomFishPrefab.cs
```diff
@@ -0,0 +1,132 @@
...
+            SplineFollowing spline = mainObj.GetOrAddComponent<SplineFollowing>();
+            spline.locomotion = loco;
+            spline.levelOfDetail = mainObj.GetOrAddComponent<BehaviourLOD>();
+            spline.GoTo(mainObj.transform.position + mainObj.transform.forward, mainObj.transform.forward, 5f);
+            behaviour.splineFollowing = spline;
+
+            if (pickupable)
+            {
+                Console.WriteLine("[FishFramework] Adding pickupable component");
+                mainObj.GetOrAddComponent<Pickupable>();
```
PrimeSonic: If the fish can be picked up, then it can be turned into an `InventoryItem`.  
That means that this entity will need a TechType and an icon for use in containers.

Take a look at the [Spawnable](https://github.com/SMLHelper/SMLHelper/blob/2.x/SMLHelper/Assets/Spawnable.cs) class that inherits from `ModPrefab` and see if there is anything there that can help with ensuring that a custom fish that is picked up will work as intended.

If this is somehow already working, disregard this comment.

SMLHelper/Patchers/CustomFishPatcher.cs
```diff
@@ -0,0 +1,68 @@
+﻿
+
+namespace SMLHelper.V2.FishFramework
```
AlexejheroYTB: Could this be merged with `SMLHelper.V2.Patchers`? The handler is already in the appropriate namespace.

SMLHelper/FishFramework/CustomFish.cs
```diff
@@ -0,0 +1,55 @@
+﻿namespace SMLHelper.V2.FishFramework
+{
+    using System;
+    using System.Collections.Generic;
+    using UnityEngine;
+
+    public class CustomFish
+    {
```
AlexejheroYTB: What is the difference between this and `CustomFishPrefab`, and why would one be used over the other?

PrimeSonic:
@kylinator25 For help with creating XML documentation, I highly recommend using GhostDoc if you don't have it already.  
https://submain.com/download/ghostdoc/community/

SMLHelper/Initializer.cs
kylinator25: All this does is harmony patch Creature.Start() to spawn the fish

SMLHelper/Assets/CustomFishPrefab.cs
kylinator25: Im using it because a lot of fish are spawned, even during gameplay when the player loads new areas, so this is to reduce the potential performance hit when loading new areas with lots of creatures

PrimeSonic: So there are performance benefits.  
Do you think this could be applied to the `ModPrefab` class directly so that all modded prefabs can get the same benefit?  
Heck, that might even be the solution to the issues we've been seeing where modded prefabs load loaded twice when being deserialized.

SMLHelper/Assets/CustomFishPrefab.cs
kylinator25: The player can pick up the fish just fine, but I have yet to integrate icons

SMLHelper/Assets/CustomFishPrefab.cs
kylinator25: If you wanna give it a shot, go ahead. It would be very helpful for things with lots of components on them

PrimeSonic: You are also more than welcome to make that change as part of this PR.  
Since you are already making changes anyways, and you clearly understand how this prefab cache pattern works, it seems only right to allow you to make the update.

That way, the `CustomFishPrefab` class wouldn't need this anymore because the base class would handle it instead.

SMLHelper/FishFramework/CustomFish.cs
kylinator25: I created this class to act as a wrapper for CustomFishPrefab that automatically supplies a TechType and registers the prefab, but if you would prefer to have the prefabs created in the same was as other ModPrefab classes thats fine

PrimeSonic: I think I understand now. So effectively this provides two different ways to register the fish.  

Either through `CustomFishPrefab` which inherits from `ModPrefab`    
or instead from `CustomFish` which is provided to the `RegisterFish` method of the `CustomFishHandler` class.

Having multiple options can be helpful in simplifying the code someone else will need to write to integrate into this API.  
Just make sure things don't get too confusing.  
Since everything is strongly typed, there is no issue with someone incorrectly using the APIs here.  
Right now I don't have any strong opinions on which way this should go.

SMLHelper/Patchers/CustomFishPatcher.cs
kylinator25: I did merge it but forgot to change the namespace -_-

SMLHelper/MonoBehaviours/CustomCreature.cs
kylinator25: I have tested it in Start() and Awake(), and it leads to unexpected results. Namely some fish are the correct scale while others are still unscaled. I will look into doing it dynamically in Update()

SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: It can be left to the modder to register a sprite for their new fish `TechType`

SMLHelper/Assets/CustomFishPrefab.cs
PrimeSonic: Going to test the change for myself before marking this as resolved.

SMLHelper/SMLHelper.csproj
```diff
@@ -67,6 +67,10 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Crafting\NodeFamily.cs" />
+    <Compile Include="FishFramework\CustomFish.cs" />
```
AlexejheroYTB: At the moment, there's only one file in the `FishFramework` folder. Could that be changed, or is it fine that way?

SMLHelper/Handlers/CustomFishHandler.cs
```diff
@@ -0,0 +1,51 @@
...
+        /// It is mainly used by CustomFishPatcher to spawn fish in
+        /// </summary>
+        public static List<TechType> fishTechTypes = new List<TechType>();
+
+        /// <summary>
+        /// Registers a CustomFish object into the game
+        /// </summary>
+        /// <param name="fish">The CustomFish that you are registering</param>
+        /// <returns>The TechType created using the info from your CustomFish object</returns>
+        public static TechType RegisterFish(CustomFish fish)
```
AlexejheroYTB: Would this work with `CustomFishPrefab`?

kylinator25: This method creates a `CustomFishPrefab` which is registered using `PrefabHandler`. If the user created their own `CustomFishPrefab` instead of using `CustomFish`, they will effectively need to do everything in this method manually

AlexejheroYTB: Ah, I see. Thanks for the clarification

PR #92, opened 2019-04-09
Language Override BugFix

### LINKED ISSUE #90: Language Overrides not compatible with string.format
We need to select a different set of wrapping characters than `{` and `}` and these cause issues when the language line starts or ends with a `string.format` placeholder `{0}`, `{1}`,... `{#}`.  
Possibly even look into how this system might be affected when line breaks are added.

PrimeSonic:
Closes #90 **Language Overrides not compatible with string.format**

Right now this only fixes issues caused by a `string.format` placeholders `{#}` being at the start or end of a language string.  
This was easily resolved by using an index/length based trimming instead of a character based trimming.

However, there are still other issues to resolve:
1. Language lines that include a line break `\n` are not properly handled by the LanguageOverride feature, as it depend on explicit line breaks to function.
2. Language lines that include the special character chosen to be the key-value separator (currently `:`) also fail parsing.

It's likely that we need to use a completely different serialization structure, likely using one that won't have issues with common special characters and `string.format` placeholders.

---

Closes #97

### DIFF
--- a/SMLHelper/Patchers/LanguagePatcher.cs
+++ b/SMLHelper/Patchers/LanguagePatcher.cs
@@ -191,7 +191,7 @@
 
         internal static string TrimTextDelimiters(string value)
         {
-            return value.Trim(TextDelimiterOpen, TextDelimiterClose);
+            return value.Substring(1, value.Length - 2);
         }
     }
 }
--- a/SMLHelper/Patchers/LanguagePatcher.cs
+++ b/SMLHelper/Patchers/LanguagePatcher.cs
@@ -16,7 +16,7 @@
         private const char TextDelimiterOpen = '{';
         private const char TextDelimiterClose = '}';
         private const char KeyValueSeparator = ':';
-        private static readonly Regex OverrideRegex = new Regex("(?<key>[\\w]+)\\s*:\\s*{(?<value>([~!@#$%^&*()\\-_=+\\[\\];:\"',<>\\/?]|{\\d+}|[\\w\\s\n])+)}(\n|\r\n)*", RegexOptions.Compiled | RegexOptions.Multiline);
+        private static readonly Regex OverrideRegex = new Regex("(?<key>[\\w]+)\\s*:\\s*{(?<value>([~!@#$%^&*()\\-_=+\\[\\];:\"',<>\\/?]|{\\d+}|[\\w\\s\n])+)}(\n|\r\n)*", RegexOptions.Multiline);
 
         private static readonly Dictionary<string, Dictionary<string, string>> originalCustomLines = new Dictionary<string, Dictionary<string, string>>();
         private static readonly Dictionary<string, string> customLines = new Dictionary<string, string>();
@@ -57,8 +57,6 @@
 
         private static void WriteOriginalCustomLines()
         {
-            Logger.Log("Writing original language files.", LogLevel.Debug);
-
             if (!Directory.Exists(LanguageOrigDir))
                 Directory.CreateDirectory(LanguageOrigDir);
 
--- a/SMLHelper/Patchers/LanguagePatcher.cs
+++ b/SMLHelper/Patchers/LanguagePatcher.cs
@@ -20,8 +20,12 @@
 
         private static readonly Dictionary<string, Dictionary<string, string>> originalCustomLines = new Dictionary<string, Dictionary<string, string>>();
         private static readonly Dictionary<string, string> customLines = new Dictionary<string, string>();
-
         private static Type languageType = typeof(Language);
+
+        internal static string GetCustomLine(string key)
+        {
+            return customLines[key];
+        }
 
         internal static void Postfix(ref Language __instance)
         {
@@ -108,27 +112,34 @@
 
                 Dictionary<string, string> originalLines = originalCustomLines[modName];
 
-                MatchCollection matches = Regex.Matches(text, OverrideRegex, RegexOptions.Multiline);
-
-                int overridesApplied = 0;
-                foreach (Match match in matches)
-                {
-                    string key = match.Groups["key"].Value;
-                    string value = match.Groups["value"].Value;
-
-                    if (!originalLines.ContainsKey(key))
-                    {
-                        Logger.Log($"Key '{key}' in language override file for '{modName}' did not match an original key.", LogLevel.Warn);
-                        continue; // Skip keys we don't recognize
-                    }
-
-                    customLines[key] = value;
-
-                    overridesApplied++;
-                }
+                int overridesApplied = ExtractCustomLinesFromText(modName, text, originalLines);
 
                 Logger.Log($"Applied {overridesApplied} language overrides to mod {modName}.", LogLevel.Info);
             }
+        }
+
+        internal static int ExtractCustomLinesFromText(string modName, string text, Dictionary<string, string> originalLines)
+        {
+            MatchCollection matches = Regex.Matches(text, OverrideRegex, RegexOptions.Multiline);
+
+            int overridesApplied = 0;
+            foreach (Match match in matches)
+            {
+                string key = match.Groups["key"].Value;
+                string value = match.Groups["value"].Value;
+
+                if (!originalLines.ContainsKey(key))
+                {
+                    Logger.Log($"Key '{key}' in language override file for '{modName}' did not match an original key.", LogLevel.Warn);
+                    continue; // Skip keys we don't recognize
+                }
+
+                customLines[key] = value;
+
+                overridesApplied++;
+            }
+
+            return overridesApplied;
         }
 
         internal static void AddCustomLanguageLine(string modAssemblyName, string lineId, string text)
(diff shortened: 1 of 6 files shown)


SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -191,7 +191,7 @@ internal static void AddCustomLanguageLine(string modAssemblyName, string lineId
 
         internal static string TrimTextDelimiters(string value)
         {
-            return value.Trim(TextDelimiterOpen, TextDelimiterClose);
+            return value.Substring(1, value.Length - 2);
```
PrimeSonic: @AlexejheroYTB pinging you for discussion of ideas.  
Note that Unity uses the single line feed character '\n' in its text.
### Important Special Characters
```CSharp
char TextDelimiterOpen = '{';
char TextDelimiterClose = '}';
char KeyValueSeparator = ':';
```

There are two ways I can see to fully support `string.format` placeholders and line breaks.  

#### Option 1
Completely re-write the serialization format into something that can safely ignore the important special characters.  
**Pros**: Written strings will match what is being sent to SMLHelper.  
**Cons:** Windows might auto-fix line breaks. Requires a lot more coding.

#### Option 2
Replace the special characters with proxy characters when writing the original language files and reverse this process when reading from them.
**Pros**: Little code change needed. Allows one entry per line.
**Cons:** Multi-line entries will be on a single line, separate by some special character or sequence, which may be confusing.

In all cases, language files will likely need to be rebuilt.

AlexejheroYTB: I would like to propose another option: escaping `{`, `}`, `:`, `\` with `\`, so they aren't treated as special characters by the parser

PrimeSonic: Adding Escapes would require updating the parser to actually handle them in the first place.  
This isn't a very complex parser. This is all the important code from it:
```CSharp
string[] languageLines = File.ReadAllLines(file, Encoding.UTF8);

Dictionary<string, string> originalLines = originalCustomLines[modName];

int overridesApplied = 0;
for (int lineIndex = 0; lineIndex < languageLines.Length; lineIndex++)
{
    string line = languageLines[lineIndex];
    if (string.IsNullOrEmpty(line))
        continue; // Skip empty lines

    string[] split = line.Split(new[] { KeyValueSeparator }, StringSplitOptions.RemoveEmptyEntries);

    if (split.Length != 2)
        continue; // Not correctly formatter

    string key = split[0];
    if (!originalLines.ContainsKey(key))
        continue; // Skip keys we don't recognize.

    customLines[key] = TrimTextDelimiters(split[1]);
    overridesApplied++;
}
```

PrimeSonic: Starting a new thread:  

With this change that trims out the outer delimiters by index, all that would be left would be to account for `string.Format` tokens `{0}` and Unix-style line break characters (which would be picked up by Windows) `\n`

Regex replacement could be done to convert patterns like `{0} {1} {2}` into `[0] [1] [2]` or something equally simple.  
And the same could be done to replace the line break characters `\n` with something like `[NewLine]`.

This wouldn't be too hard to handle in each direction, and would require a minimal level of updates on behalf of the user to convert their existing files to use these new tokens instead.

AlexejheroYTB: I personally would take a different approach, by not escaping the characters at all, but by changing the way the file is parsed. Instead of checking each file individually we can parse it using regex and checking for the delimiter characters `:{` and `}\n`

PrimeSonic: That should be doable. Here's a regex to try.  
`(?<key>[\w]+):(?<value>{[\w\s{}\n]+}\n\r?) `  
We'll need to make sure that multi-line matching is enabled

PrimeSonic: Better regex: `(?<key>[\w]+):\s*{(?<value>({\d+}|[\w\s\n])+)}(\n|\n\r)*`

AlexejheroYTB: Even better regex: `(?<key>\w+?)\s*?:\s*?{(?<value>[\s\S]+?)}(\n|\n\r)?`

PrimeSonic: `\s\S` is too lazy. This regex fails to handle the string.format tokens.
Sample that doesn't work for this one `key:{val{0}ue}`

PrimeSonic: Until we can find a better way,  
`(?<key>[\w]+)\s*:\s*{(?<value>([~!@#$%^&*()\-_=+\[\];:"',<>\/?]|{\d+}|[\w\s\n])+)}(\n|\n\r)*`  
this regex includes explicit matching on every special character I can easily access from my keyboard.

Matches special characters `[~!@#$%^&*()\-_=+\[\];:"',<>\/?]`  
Matches string.format tokens `{\d+}`
Matches letters, numbers, and line breaks `[\w\s\n]`

AlexejheroYTB: `(?<key>\w+?)\s*?:\s*?{(?<value>[\s\S]+?)}(\n|\n\r)$`

PrimeSonic: That somehow works against my test text
```
key:{value}
key:{valu{0}e}
key:{valu
{0}e}
key:{val-u{0}e}
key:{val__[!]
-u{0}{1}e}
```
Validated on https://regex101.com/

AlexejheroYTB: `$` matches the end of the text/line so we don't have to check for `\n` and `\r` characters

AlexejheroYTB: The groups are also parsed correctly

image

PrimeSonic: Looks promising.  
If you want to try replacing the line-by-line parsing logic with one that applies this regex to the full text and then iterates over the matches, go for it.

PrimeSonic: Still the only regex that passes all tests  
`(?<key>[\w]+)\s*:\s*{(?<value>([~!@#$%^&*()\-_=+\[\];:"',<>\/?]|{\d+}|[\w\s\n])+)}(\n|\r\n)*`

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -100,84 +101,43 @@ private static void ReadOverrideCustomLines()
...
-                    }
-                }
-                else
-                {
-                    return true; // Key not found
+                    Logger.Log($"Key '{key}' in language override file for '{modName}' did not match an original key.", LogLevel.Warn);
+                    continue; // Skip keys we don't recognize
                 }
+
+                customLines[key] = value;
```
PrimeSonic: Before doing this, first check if `key` begins with the prefix "Tooltip_".  
If it does, it will be necessary to override the value for this key in the doctionary `TooltipFactory.techTypeTooltipStrings.valueToString`.

AlexejheroYTB:
What about chinese characters? What about russian characters? For the value, \w is not enough.

AlexejheroYTB:
What if someone wanted to add some text between `{` and `}`? `{\d+}` does not cover that.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -15,12 +16,17 @@ internal class LanguagePatcher
         private const char TextDelimiterOpen = '{';
         private const char TextDelimiterClose = '}';
         private const char KeyValueSeparator = ':';
+        private static readonly Regex OverrideRegex = new Regex("(?<key>[\\w_]+)\\s*:\\s*{(?<value>([~!@#$%^&*()\\-_=+\\[\\];:\"',<>\\/?]|{\\d+}|[\\w\\s\n])+)}(\n|\r\n)*", RegexOptions.Multiline);
```
AlexejheroYTB: In a regex like this, everything should be made lazy. A greedy quantifier could cause the value to also include the next line.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -15,12 +16,17 @@ internal class LanguagePatcher
         private const char TextDelimiterOpen = '{';
         private const char TextDelimiterClose = '}';
         private const char KeyValueSeparator = ':';
```
AlexejheroYTB: What if the user forgets to close the `}`? With this PR, there's no real need to keep the text delimiters. At least not both of them.

PrimeSonic:
> What if someone wanted to add some text between `{` and `}`? `{\d+}` does not cover that.
`{\d+}` is specifically for string.format tokens.

SMLHelper/Patchers/LanguagePatcher.cs
PrimeSonic: Forgetting the end delimiter would break multi-line values.  
Not really an option.

AlexejheroYTB:
Closed in favor of #97

PR #95, opened 2019-04-21
JSON parser for language overrides

AlexejheroYTB:


AlexejheroYTB:
I have a better idea which doesn't involve changing the parser

PR #97, opened 2019-05-19
Language Override BugFix (attempt no. 2)

### LINKED ISSUE #90: Language Overrides not compatible with string.format

AlexejheroYTB:
Closes #90 
Closes #92

### DIFF
--- a/SMLHelper/Patchers/LanguagePatcher.cs
+++ b/SMLHelper/Patchers/LanguagePatcher.cs
@@ -120,6 +120,12 @@
 
                 string key = split[0];
 
+                if (split.Length < 2)
+                {
+                    Logger.Log($"Line '{lineIndex}' in language override file for '{modName}' was incorrectly formatted.", LogLevel.Warn);
+                    continue; // Not correctly formatted
+                }
+
                 if (!originalLines.ContainsKey(key))
                 {
                     Logger.Log($"Key '{key}' on line '{lineIndex}' in language override file for '{modName}' did not match an original key.", LogLevel.Warn);
@@ -150,6 +156,11 @@
             foreach (string line in lines)
             {
                 string[] split = line.Split(new[] { KeyValueSeparator }, 2, StringSplitOptions.RemoveEmptyEntries);
+
+                if (split.Length < 2)
+                {
+                    return true; // Not correctly formatted
+                }
 
                 string lineKey = split[0];
                 string lineValue = split[1].Replace("\\n", "\n").Replace("\\r", "\r");
--- a/SMLHelper/Patchers/LanguagePatcher.cs
+++ b/SMLHelper/Patchers/LanguagePatcher.cs
@@ -12,8 +12,6 @@
         private static readonly string LanguageDir = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Language");
         private static readonly string LanguageOrigDir = Path.Combine(LanguageDir, "/Originals");
         private static readonly string LanguageOverDir = Path.Combine(LanguageDir, "/Overrides");
-        private const char TextDelimiterOpen = '{';
-        private const char TextDelimiterClose = '}';
         private const char KeyValueSeparator = ':';
 
         private static readonly Dictionary<string, Dictionary<string, string>> originalCustomLines = new Dictionary<string, Dictionary<string, string>>();
@@ -66,7 +64,7 @@
             }
 
             if (filesWritten > 0)
-                Logger.Log($"Updated {filesWritten} of {originalCustomLines.Count} original language files.", LogLevel.Info);
+                Logger.Log($"Updated {filesWritten} of {originalCustomLines.Count} original language files.", LogLevel.Debug);
         }
 
         private static void WriteOriginalLinesFile(string modKey)
@@ -75,8 +73,8 @@
             var text = new StringBuilder();
             foreach (string langLineKey in modCustomLines.Keys)
             {
-                string value = modCustomLines[langLineKey];
-                text.AppendLine($"{langLineKey}{KeyValueSeparator}{TextDelimiterOpen}{value}{TextDelimiterClose}");
+                string value = modCustomLines[langLineKey].Replace("\n", "\\n").Replace("\r", "\\r");
+                text.AppendLine($"{langLineKey}{KeyValueSeparator}{value}");
             }
 
             File.WriteAllText($"{LanguageOrigDir}/{modKey}.txt", text.ToString(), Encoding.UTF8);
@@ -87,7 +85,7 @@
             if (!Directory.Exists(LanguageOverDir))
                 Directory.CreateDirectory(LanguageOverDir);
 
-            string[] files = Directory.GetFiles(LanguageOverDir);
+            string[] files = Directory.GetFiles(LanguageOverDir, "*.txt");
 
             if (files.Length == 0)
                 return;
@@ -96,7 +94,7 @@
 
             foreach (string file in files)
             {
-                string modName = Path.GetFileName(file).Replace(".txt", string.Empty);
+                string modName = Path.GetFileNameWithoutExtension(file);
 
                 if (!originalCustomLines.ContainsKey(modName))
                     continue; // Not for a mod we know about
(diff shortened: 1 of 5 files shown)


SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -112,13 +110,7 @@ private static void ReadOverrideCustomLines()
                     if (string.IsNullOrEmpty(line))
                         continue; // Skip empty lines
 
-                    string[] split = line.Split(new[] { KeyValueSeparator }, StringSplitOptions.RemoveEmptyEntries);
-
-                    if (split.Length != 2)
```
PrimeSonic: The file being read is one that was manually edited.  
If we remove this safety check here, then we risk throwing an IndexOutOfRangeException when we do something like this:  
 ```
split[1]
```
Let's leave the safety checks in.

File: SMLHelper/Patchers/LanguagePatcher.cs [hunk 2]
```diff
@@ -12,8 +12,6 @@ internal class LanguagePatcher
         private static readonly string LanguageDir = Path.Combine(Path.GetDirectoryName(Assembly.GetExecutingAssembly().Location), "Language");
         private static readonly string LanguageOrigDir = Path.Combine(LanguageDir, "/Originals");
         private static readonly string LanguageOverDir = Path.Combine(LanguageDir, "/Overrides");
```
AlexejheroYTB: The language patcher currently doesn't work correctly and returns 0 files. Will have to fix that before merging.

PrimeSonic: We can use our last attempt instead for now (which has been proven to work) and change it later.  
https://github.com/SMLHelper/SMLHelper/pull/92

AlexejheroYTB: After 22b8013, works perfectly. The only current issue(?) is that the curly braces are added as part of the text for old files.

AlexejheroYTB: But I don't think that's an "issue". People can update their files.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -152,13 +155,15 @@ private static bool FileNeedsRewrite(string modKey)
             // Confirm if the file actually needs to be updated
             foreach (string line in lines)
             {
-                string[] split = line.Split(new[] { KeyValueSeparator }, StringSplitOptions.RemoveEmptyEntries);
+                string[] split = line.Split(new[] { KeyValueSeparator }, 2, StringSplitOptions.RemoveEmptyEntries);
 
-                if (split.Length != 2)
-                    return true; // Malformatted, likely externally edited
+                if (split.Length < 2)
```
PrimeSonic: Values greater than `2` would also be invalid if we're trying to keep things to a single `Key:Value` entry per line.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -75,11 +71,11 @@ private static void WriteOriginalLinesFile(string modKey)
             var text = new StringBuilder();
             foreach (string langLineKey in modCustomLines.Keys)
             {
-                string value = modCustomLines[langLineKey];
-                text.AppendLine($"{langLineKey}{KeyValueSeparator}{TextDelimiterOpen}{value}{TextDelimiterClose}");
+                string value = modCustomLines[langLineKey].Replace("\n", "\\n").Replace("\r", "\\r");
+                text.AppendLine($"{langLineKey}{KeyValueSeparator}{value}");
```
PrimeSonic: Better backwards compatibility could be had if we maintain the old `TextDelimiterOpen`|`TextDelimiterClose` values.  
They can be trimmed safely by index by removing the first and last characters of the captured value after the `KeyValueSeparator`.

SMLHelper/Patchers/LanguagePatcher.cs
```diff
@@ -105,41 +101,48 @@ private static void ReadOverrideCustomLines()
...
+        {
+            int overridesApplied = 0;
+            for (int lineIndex = 0; lineIndex < languageLines.Length; lineIndex++)
+            {
+                string line = languageLines[lineIndex];
+                if (string.IsNullOrEmpty(line))
+                    continue; // Skip empty lines
 
-                    string key = split[0];
+                string[] split = line.Split(new[] { KeyValueSeparator }, 2, StringSplitOptions.RemoveEmptyEntries);
```
PrimeSonic: For the sake of documentation, replace the literal `2` in all calls into `string.Split` and `split.Length` with a reference to a new constant.
```CSharp 
const int SegmentCount = 2;
```

Good call using [Split(String[], Int32, StringSplitOptions)](https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-4.8#System_String_Split_System_String___System_Int32_System_StringSplitOptions_) to avoid the issues of `:` characters in the value text.

SMLHelper/Patchers/LanguagePatcher.cs
AlexejheroYTB: Values cannot be greater than 2

SMLHelper/Patchers/LanguagePatcher.cs
AlexejheroYTB: They are completely useless. I could implement a conditional check to only remove them if they are present.

SMLHelper/Patchers/LanguagePatcher.cs
PrimeSonic: I know that, and you know that, but for the sake of clarity and self-documentation, we can use `!= SegmentCount` to be very clear about what this code is validating.

SMLHelper/Patchers/LanguagePatcher.cs
AlexejheroYTB: I'm not sure how to respond do this. Why would we need to define a constant _integer_ value which is only used in 1 or 2 places?

PrimeSonic: Using a well named constant can make it effortless to infer the _meaning_ and _intent_ of the literal value its substituting.  
That makes for clean, self-documenting code that is much more easily understood by others.

Like it or not, SMLHelper is an open source project and we should be making every effort to make it easy for other people to jump in, understand what they are looking at, and be able to contribute.

SMLHelper/Patchers/LanguagePatcher.cs
AlexejheroYTB: A conditional check would actually not work. Stuff like `{0}CustomValue{1}` would become `0}CustomValue{1`. We either completely get rid of the text delimiters, or keep them forever. (There is literally no point to keep them except backwards compatibility, and we're talking about Unity 2018 here...)

PrimeSonic: > There is literally no point to keep them except backwards compatibility

Precisely.  
While the Unity 2018 will cause a lot of headaches for many, users should only really need to update their mods as updates are made.  
I am perfectly aware that we no longer need delimiters after this update, but I am making the case that if we can save 99% of users the hassle of having to rebuild or update their override files, it's worth it.

You're welcome to make the case that this isn't necessary. Otherwise, I can add these updates in myself later tonight.

AlexejheroYTB: What if a text delimiter is not found? Do we just throw an error or something? And refuse to load the line even if it's perfectly fine?

PrimeSonic: Resolved without re-writing the original file.  
Entries that use `{`|`}` delimiters wrapping the value will be safely removed before overriding without accidentally removing string.format tokens.

AlexejheroYTB:
If all tests are passed, this pull request is ready for merge.

PrimeSonic:
image

PR #98, opened 2019-10-30
SMLHelper 2.3

PrimeSonic:
## Little Big Update
- Now updated for new Assembly-CSharp and Unity DLLs
- https://unknownworlds.com/subnautica/subnautica-big-little-update-released/

## New Features
- ItemStorageHelper
- StorageHelperExtensions
- CustomFish
- ItemActions
- `GetOrAddComponent` added to PrefabUtils
- Interfaces for all handlers

## Improvements
- Better logging
- Replaced hardcoded file paths with paths relative to the SMLHelper's execution location

## Bug Fixes
- Better parsing of LanguageOverride files when line breaks are involved
- Multiple calls into `KnownTechHandler.AddAnalysisTech` will update correctly instead of overwriting each other
- Cloned prefabs of Buildables now properly disposed of, freeing up resources
- Multiple calls into `SetTechData` will now see the later entry overwriting the earlier entry, instead of discarding both

## Legacy APIs fully deprecated
- The old legacy fields and methods are no longer available
- Please update to V2 APIs


SMLHelper/Assets/ModPrefab.cs
```diff
@@ -62,6 +65,14 @@ protected ModPrefab(string classId, string prefabFileName, TechType techType = T
 
         internal GameObject GetGameObjectInternal()
         {
+            if(cachedPrefab)
+            {
+                // This will avoid GetGameObject() being called every time the object is spawned. Could break things if there is a degree 
+                // of random generation in GetGameObject(), in that case the modder will need to do their random stuff in a MonoBehaviour.Awake()
+                // e.g. in MinecraftFish, the tropical fish have random colours which is done in a MonoBehaviour, rather than the GetGameObject() method
+                return cachedPrefab;
```
PrimeSonic: After some more review, I am suggesting that we make this an optional feature.  
Any mods that conditionally return `null` from `GetGameObject` (intentionally) will break with this change.

I have a couple of mods that do that today and it stands to reason that other mods, either today or in the future, may do the same.
https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/blob/master/CyclopsNuclearReactor/CyNukReactorBuildable.cs#L108
https://github.com/PrimeSonic/PrimeSonicSubnauticaMods/blob/master/CyclopsBioReactor/Items/CyBioReactor.cs#L69

If caching were made optional and default it to "do not cache" to preserve compatibility with the behavior of existing mods,  
I would expect that the feature would almost never be used.

Considering how little impact comes from this one method, I suggest that the sane thing to do is to remove prefab caching entirely.

AlexejheroYTB: My opinion is that prefab caching should be removed.

zorgesho: For static objects that cached prefab is deleting in Fixer component after 1 second anyway.  It now works only for physics enabled objects and only because prefabs are sinking down infinitely.

PrimeSonic: So Buildables aren't being removed by the Fixer component.  
That is something that will need to be resolved separately.

SMLHelper/Handlers/CraftTreeHandler.cs
```diff
@@ -162,11 +307,35 @@ public static void AddTabNode(CraftTree.Type craftTree, string name, string disp
...
+        void ICraftTreeHandler.RemoveNode(CraftTree.Type craftTree, params string[] stepsToNode)
         {
-            Assert.IsTrue(craftTree <= CraftTree.Type.Rocket, $"{nameof(RemoveNode)} is intended for use only with standard crafting trees, not custom ones.");
+            ValidateStandardCraftTree(craftTree);
             CraftTreePatcher.NodesToRemove.Add(new Node(stepsToNode, craftTree));
         }
+
+        #endregion
+
+        private static void ValidateStandardCraftTree(CraftTree.Type craftTree)
```
PrimeSonic: Something about the new validation method is breaking custom crafting trees.

PrimeSonic: Found the issue:  https://github.com/QModManager/SMLHelper/blob/c145c5f7d431fa9aeb8aa77d250ee1d872b774e2/SMLHelper/Assets/Craftable.cs#L44

The `Craftable` class assumes non-modded crafting trees.  
In its current state, the `Craftable` becomes incompatible with modded crafting trees in the 2.3 update.  
This should be fixed if at all possible

PrimeSonic: Current workaround is to use the `Spawnable` or `ModPrefab` class instead of `Craftable`.

AlexejheroYTB:
There are still changes that aren't present here from QModManager/QModManager#93 and QModManager/QModManager#100

SMLHelper/Handlers/CraftTreeHandler.cs
PrimeSonic: After further investigation, it's clear that the `Craftable` class was never intended to be used with a custom crafting tree.  

At this time, there isn't a way to make modded craft trees with the `Craftable` class.

PR #104, opened 2019-11-06
Subnautica 2019 Update

PrimeSonic:
- Updating references and files to Unity2018
- Deleting Legacy support


kylinator25:
The Texture2D.LoadImage() extension method is now in UnityEngine.ImageConversionModule.dll, so don't forget to reference that.

PrimeSonic:
> Looks good to me, but where did everything from `SMLHelper/Legacy/Utility.cs` go?

Those were old utility methods used in patching.  
SMLHelper V2 doesn't use these. And given their function, I don't expect that any mods were using these utility functions.

AlexejheroYTB:
Is this updated for 2018 or 2019?
(I read the name, I'm just making sure)

PrimeSonic:
> Is this updated for 2018 or 2019?
> (I read the name, I'm just making sure)

This is for the 2019 update of Subnautica: the update the brings us to Unity2018.

PR #105, opened 2019-11-26
Support for Below Zero (Unity2018) & Subnautica (Unity2019)

MrPurple6411:

# SMLHelper is coming to Below Zero
**Project now has custom build configurations: `SUBNAUTICA` and `BELOWZERO`**
- The all relevant DLLs for both games are available but the build configuration ensures that the correct ones are included
- The build configurations enable preprocessor directives that allow specific code to only be compiled on a specific configuration
 - The build configuration even switches the .NET Framework version as Subnautica has been updated to .NET 4 while Below Zero will continue on .NET 3.5 until 2020
  - This allows us to support both Subnautica and Below Zero on a single branch

**Support was added for patching CraftData (now TechData) in Below Zero**
- `ICraftDataHandler` will expose a few new methods on Below Zero not available to Subnautica
- `CraftDataHandler` and `CraftDataPatcher` contain different implementations exclusive to Subnautica and Below Zero to handle the differences in the way BZ changed how crafting data is handled

# Other changes

**The `AssetsFolder` property in the `Spawnable` class is now `virtual` with default behavior**
- When default behavior is used, it assumes that the default folder is the same folder where the mod DLL is
**`Spawnable` now exposes a new `virtual` method: `GetItemSprite` that mods can override**
- The default behavior will load a sprite from an image using the `AssetsFolder` and `IconFileName` properties
- This can be overridden to allow mods to use this class while providing their own sprite from another source
- In the event that a sprite cannot be found for the default behavior, a warning will be logged and the `SpriteManager.defaultSprite` will be used
- No exception will be thrown when an icon is not provided

**Previously missing XML documentation now added**

**Enum cache files now save only when the player manually saves the game**
- A separate file will be saved for entries from mods that aren't active
- Will allow those values to be recovered later

**Added a new `Handler` class that makes it easy to discover all of the various handler classes provided by SMLHelper**

**New `CustomFabricator` asset class has been added**
- Streamlines the process of creating custom fabricators

New `IngameMenuHandler` offers simple utilities for patching into the In-game menu**
- Currently offers methods for attaching events to invoke when the player saves the game


Example mod/Example mod.csproj
```diff
@@ -41,7 +41,7 @@
     <Reference Include="System.Data" />
     <Reference Include="System.Xml" />
     <Reference Include="UnityEngine.CoreModule">
-      <HintPath>..\Dependencies\UnityEngine.CoreModule.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\UnityEngine.CoreModule.dll</HintPath>
```
PrimeSonic: **TODO** Separate Unity DLLs from assembly-csharp DLLs.  
The Unity libraries are now the same between both games and can be handled together.

Example mod/Example mod.csproj
```diff
@@ -32,7 +32,7 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Assembly-CSharp">
-      <HintPath>..\Dependencies\Assembly-CSharp.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\Assembly-CSharp.dll</HintPath>
```
PrimeSonic: **TODO** 
Introduce two new [custom build configurations](https://docs.microsoft.com/en-us/visualstudio/ide/how-to-create-and-edit-configurations?view=vs-2019): `SUBNAUTICA`|`BELOWZERO`. 

Use these to [conditionally select](https://stackoverflow.com/questions/1786917/is-there-a-way-to-specify-assembly-references-based-on-build-configuration-in-vi) the correct Assembly-CSharp DLL for the game.

SMLHelper.Tests/SMLHelper.Tests.csproj
```diff
@@ -50,10 +50,10 @@
   </PropertyGroup>
   <ItemGroup>
     <Reference Include="Assembly-CSharp">
-      <HintPath>..\Dependencies\Assembly-CSharp.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\Assembly-CSharp.dll</HintPath>
     </Reference>
     <Reference Include="Assembly-CSharp-firstpass_publicized">
-      <HintPath>..\Dependencies\Assembly-CSharp-firstpass_publicized.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\Assembly-CSharp-firstpass_publicized.dll</HintPath>
```
PrimeSonic: **TODO** Conditional references based on build configuration

SMLHelper/Assets/FishPrefab.cs
```diff
@@ -69,7 +69,7 @@ public sealed override GameObject GetGameObject()
             forces.handleGravity = true;
             forces.underwaterDrag = 1f;
             forces.underwaterGravity = 0;
-            forces.waterDepth = Ocean.main.GetOceanLevel();
+            forces.waterDepth = Ocean.GetOceanLevel();
```
PrimeSonic: **TODO** This will be updated to compile differently for the two [build configurations](https://docs.microsoft.com/en-us/visualstudio/ide/understanding-build-configurations?view=vs-2019) using [preprocessor directives](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives/)
```CS
#if SUBNAUTICA 
forces.waterDepth = Ocean.main.GetOceanLevel();
#elif BELOWZERO
forces.waterDepth = Ocean.GetOceanLevel();
#endif
```

SMLHelper/Assets/PdaItem.cs
```diff
@@ -58,7 +58,7 @@ internal PdaItem(string classId, string friendlyName, string description)
         /// <summary>
         /// This provides the <see cref="TechData"/> instance used to designate how this item is crafted or constructed.
         /// </summary>
-        protected abstract TechData GetBlueprintRecipe();
+        protected abstract RecipeData GetBlueprintRecipe();
```
PrimeSonic: **TODO** Update with conditional compilation. 
Update XML documentation to match code.
```CS
#if SUBNAUTICA 
/// <summary>
/// This provides the <see cref="TechData"/> instance used to designate how this item is crafted or constructed.
/// </summary>
protected abstract TechData GetBlueprintRecipe();
#elif BELOWZERO
/// <summary>
/// This provides the <see cref="RecipeData"/> instance used to designate how this item is crafted or constructed.
/// </summary>
protected abstract RecipeData GetBlueprintRecipe();
#endif
```

SMLHelper/Crafting/RecipeData.cs
```diff
@@ -5,8 +5,8 @@
     /// <summary>
     /// A class that fully describes a recipe for a <see cref="TechType"/> identified item.
     /// </summary>
-    /// <seealso cref="ITechData" />
-    public class TechData : ITechData
+    /// <seealso cref="TechData" />
+    public class RecipeData
```
PrimeSonic: **TODO** Separate SMLHelper's `TechData` and the new `RecipeData` into two separate files.  
These will be wrapped entirely by the preprocessor directives. 

```CS
#if SUBNAUTICA
public class TechData : ITechData
{
....
}
#endif
```
```CS
#if BELOWZERO
public class RecipeData
{
....
}
#endif
```

SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -26,11 +28,11 @@ private CraftDataHandler()
         /// <para>Can be used for existing TechTypes too.</para>
         /// </summary>
         /// <param name="techType">The TechType whose TechData you want to edit.</param>
-        /// <param name="techData">The TechData for that TechType.</param>
-        /// <seealso cref="TechData"/>
-        public static void SetTechData(TechType techType, ITechData techData)
+        /// <param name="recipeData">The TechData for that TechType.</param>
+        /// <seealso cref="RecipeData"/>
+        public static void SetTechData(TechType techType, RecipeData recipeData)
```
PrimeSonic: **TODO**  Determine if it would be best to have two separate files for each version of `CraftDataHandler` to instead to handle the changes within the same file.

SMLHelper/Interfaces/ICraftDataHandler.cs
```diff
@@ -1,26 +1,27 @@
 ﻿namespace SMLHelper.V2.Interfaces
 {
+    using System.Collections.Generic;
     using Crafting;
 
     public interface ICraftDataHandler
     {
         /// <summary>
-        /// <para>Allows you to edit recipes, i.e. TechData for TechTypes.</para>
+        /// <para>Allows you to edit JsonValues Directly for TechTypes.</para>
```
PrimeSonic: **TODO** Same as `CraftDataHandler`.

SMLHelper/Initializer.cs
```diff
@@ -29,6 +29,7 @@ public static void Patch()
 
         internal static void Initialize()
         {
+            TechDataPatcher.Patch(harmony);
```
PrimeSonic: **TODO**

```CS
#if BELOWZERO
TechDataPatcher.Patch(harmony);
#endif
```

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -7,23 +7,6 @@
 
     internal class CraftDataPatcher
     {
-        #region Internal Fields
```
PrimeSonic: **TODO** Wrap all Subnautica specific for in 
```CS
#if SUBNAUTICA
// code here
#endif
```

SMLHelper/Patchers/TechDataPatcher.cs
```diff
@@ -0,0 +1,74 @@
...
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Linq;
+    using System.Text;
+    using Assets;
+    using Harmony;
+
+
+    internal class TechDataPatcher
```
PrimeSonic: **TODO** Wrap this entire file in 
```CS
#if BELOWZERO
// code here
#endif
```

SMLHelper/SMLHelper.csproj
```diff
@@ -30,9 +30,9 @@
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
     <WarningsNotAsErrors>612,618</WarningsNotAsErrors>
-    <DebugType>pdbonly</DebugType>
+    <DebugType>none</DebugType>
     <Optimize>true</Optimize>
-    <OutputPath>bin\Release\</OutputPath>
+    <OutputPath>..\..\..\Program Files %28x86%29\Steam\steamapps\common\SubnauticaZero\QMods\Modding Helper\</OutputPath>
```
PrimeSonic: Revert user specific changes

SMLHelper/SMLHelper.csproj
```diff
@@ -47,17 +47,25 @@
...
   <ItemGroup>
     <Reference Include="0Harmony">
-      <HintPath>..\Dependencies\0Harmony-1.2.0.1.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\0Harmony-1.2.0.1.dll</HintPath>
+      <Private>False</Private>
     </Reference>
     <Reference Include="Assembly-CSharp-firstpass_publicized">
-      <HintPath>..\Dependencies\Assembly-CSharp-firstpass_publicized.dll</HintPath>
+      <HintPath>..\..\DependenciesBZ\Assembly-CSharp-firstpass_publicized.dll</HintPath>
+      <Private>False</Private>
```
PrimeSonic: **TODO** Handle the version specific references based on the build configuration

SMLHelper/mod.json
```diff
@@ -4,7 +4,7 @@
   "Author": "AHK1221 & PrimeSonic & AlexejheroYTB",
   "Version": "2.3.0",
   "Enable": true,
-  "Game": "Subnautica",
+  "Game": "BelowZero",
```
PrimeSonic: **TODO**
Create separate mod.json files for each game.  
Update version to `2.3.1`.

Example mod/Example mod.csproj
PrimeSonic: For now, Unity DLLs are simply in the Dependencies folder

SMLHelper/Patchers/TechDataPatcher.cs
PrimeSonic: This file was removed - All of its patching was moved into CraftDataPatcher to keep related patches in the same space

SMLHelper/Initializer.cs
PrimeSonic: TechDataPatcher was removed - all BZ recipe patching was moved into CraftDataPatcher

SMLHelper/Handlers/CraftDataHandler.cs
PrimeSonic: A single Handler was maintained with different sections for Subnautica and Below Zero.  
The `ICraftDataHandler` interface was updated to have slightly different methods for the two games.

SMLHelper/mod.json
PrimeSonic: After separating this into two files, a simple post build event was added that copies the appropriate file into the output folder with the name `mod.json`

.gitignore
```diff
@@ -263,3 +263,9 @@ Dependencies/Assembly-CSharp-firstpass.dll
 Dependencies/Assembly-CSharp.dll
 /Dependencies/Assembly-CSharp-firstpass_publicized.dll
 /Dependencies/Assembly-CSharp_publicized.dll
+/DependenciesBZ/Assembly-CSharp_publicized.dll
+/DependenciesBZ/Assembly-CSharp-firstpass_publicized.dll
+/DependenciesBZ/Assembly-CSharp.dll
+/DependenciesBZ/Assembly-CSharp_publicized.dll
+/DependenciesBZ/Assembly-CSharp-firstpass.dll
+/DependenciesBZ/Assembly-CSharp-firstpass_publicized.dll
```
AlexejheroYTB: Common dependencies should be stored in `Dependencies`
Subnautica dependencies should be stored in `Dependencies/Subnautica`
Below Zero dependencies should be stored in `Dependencies/BelowZero`

SMLHelper.sln
```diff
@@ -11,22 +11,22 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SMLHelper.Tests", "SMLHelpe
...
+		{418502DD-372D-4EF9-8021-B262552DFEDE}.SUBNAUTICA|Any CPU.ActiveCfg = SUBNAUTICA|Any CPU
+		{418502DD-372D-4EF9-8021-B262552DFEDE}.SUBNAUTICA|Any CPU.Build.0 = SUBNAUTICA|Any CPU
+		{C8FB0981-77D2-47C7-BBEF-A3A9EBACACBF}.BELOWZERO|Any CPU.ActiveCfg = BELOWZERO|Any CPU
+		{C8FB0981-77D2-47C7-BBEF-A3A9EBACACBF}.BELOWZERO|Any CPU.Build.0 = BELOWZERO|Any CPU
+		{C8FB0981-77D2-47C7-BBEF-A3A9EBACACBF}.SUBNAUTICA|Any CPU.ActiveCfg = SUBNAUTICA|Any CPU
+		{C8FB0981-77D2-47C7-BBEF-A3A9EBACACBF}.SUBNAUTICA|Any CPU.Build.0 = SUBNAUTICA|Any CPU
+		{EBC4AAF3-A13B-427E-BF99-3AE23EDEB4B8}.BELOWZERO|Any CPU.ActiveCfg = BELOWZERO|Any CPU
+		{EBC4AAF3-A13B-427E-BF99-3AE23EDEB4B8}.BELOWZERO|Any CPU.Build.0 = BELOWZERO|Any CPU
+		{EBC4AAF3-A13B-427E-BF99-3AE23EDEB4B8}.SUBNAUTICA|Any CPU.ActiveCfg = SUBNAUTICA|Any CPU
+		{EBC4AAF3-A13B-427E-BF99-3AE23EDEB4B8}.SUBNAUTICA|Any CPU.Build.0 = SUBNAUTICA|Any CPU
```
AlexejheroYTB: Something like this will probably need to be done in the QModManager solution as well when the merge happens

SMLHelper/Assets/FishPrefab.cs
```diff
@@ -69,7 +69,11 @@ public sealed override GameObject GetGameObject()
             forces.handleGravity = true;
             forces.underwaterDrag = 1f;
             forces.underwaterGravity = 0;
+#if BELOWZERO
+            forces.waterDepth = Ocean.GetOceanLevel();
+#elif SUBNAUTICA
             forces.waterDepth = Ocean.main.GetOceanLevel();
+#endif
```
AlexejheroYTB:
```cs
#if BELOWZERO
    forces.waterDepth = Ocean.GetOceanLevel();
#else 
    forces.waterDepth = Ocean.main.GetOceanLevel();
#endif
```

SMLHelper/Interfaces/ICraftDataHandler.cs
```diff
@@ -1,9 +1,60 @@
 ﻿namespace SMLHelper.V2.Interfaces
 {
+#if BELOWZERO
+    using System.Collections.Generic;
+#endif
     using Crafting;
 
     public interface ICraftDataHandler
     {
+#if BELOWZERO
```
AlexejheroYTB: Split `ICraftDataHandler.cs` into multiple files for SN and BZ, this is getting messy.

SMLHelper/Patchers/CraftDataPatcher.cs
```diff
@@ -11,6 +11,8 @@ internal class CraftDataPatcher
 
         private static readonly Func<TechType, string> AsStringFunction = (t) => t.AsString();
 
+#if SUBNAUTICA
```
AlexejheroYTB: Split `CraftDataPatcher.cs` into multiple files for SN and BZ, this is getting messy.

SMLHelper/modBZ.json
```diff
@@ -0,0 +1,10 @@
+{
```
AlexejheroYTB: `modSN.json` and `modBZ.json`? Really?
Group these in a single folder like `QModManager/Subnautica.json` and `QModManager/BelowZero.json` or something

SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -21,6 +22,69 @@ private CraftDataHandler()
 
         #region Static Methods
 
+#if BELOWZERO
```
AlexejheroYTB: Split `CraftDataHandler.cs` into multiple files for SN and BZ, this is getting messy.

SMLHelper/Assets/FishPrefab.cs
AlexejheroYTB: If both `SUBNAUTICA` and `BELOWZERO` are false, default to `SUBNAUTICA` instead of throwing a compile error.

SMLHelper/Assets/FishPrefab.cs
```diff
@@ -69,7 +69,11 @@ public sealed override GameObject GetGameObject()
             forces.handleGravity = true;
             forces.underwaterDrag = 1f;
             forces.underwaterGravity = 0;
+#if BELOWZERO
```
AlexejheroYTB: Having different build results for Subnautica and Below Zero would mean that QModManager would have separate build results for Subnautica and Below Zero after the merge. For now, it is fine, but I will make some modifications when the merge occurs, perhaps by using a custom build action to update the code.

```cs
#if BELOWZERO
    forces.waterDepth = Ocean.GetOceanLevel();
#else 
    forces.waterDepth = Ocean.main.GetOceanLevel();
#endif 
```

```cs
if (Patcher.Game == Game.BelowZero)
{
    // do some reflection shit here
}
else
{
	// do some other reflection shit here
}
```

The only (pseudo-)down-side is that the game will take longer to load by 0.1 seconds.

Anyway, I'm not saying that I will do that, since it's complicated as hell, but a similar custom build event could be used to add _dependency injection to all classes that require it without having to bloat up the repo_.

SMLHelper/Assets/FishPrefab.cs
PrimeSonic: There are no other build configurations in the project anymore.  
The default RELEASE and DEBUG build configurations were removed as part of this effort.  
There is no reason to ever encounter the scenario of both of these if-directives returning false.

The reason I choose not to use `#else` is so that it is easy to search the code for Subnautica and BelowZero specific code by searching for "if SUBNAUTICA" and "if BELOWZERO".  
We will lose that if we start using the `#else` directive.

SMLHelper/Assets/FishPrefab.cs
PrimeSonic: If we merge, it will no longer be possible to have the same QModInstaller exe and the same QModManager dll serving both games.  

SMLHelper will want direct references to the correct Assembly-CSharp.  
And we can't have a project attempting to reference all of them for both games at the same time.

This is another reason why merging SMLHelper into QMM had to be postponed.

SMLHelper/Interfaces/ICraftDataHandler.cs
PrimeSonic: This I will insist that we do not.  
This same interface will define both the Subnautica `CraftDataHandler` and the BelowZero `TechDataHandler`.  

Having a single interface file will reduce the chances of someone forgetting to update one of the two handlers whenever a feature is added or modified.

I will also insist that this particular file isn't messy at all.  
The methods exclusive to each game were organized at the time of the interface and are clearly separated, with the common methods following after.

SMLHelper/Patchers/CraftDataPatcher.cs
PrimeSonic: This will be handled by @ahk1221 and @MrPurple6411

SMLHelper/Handlers/CraftDataHandler.cs
PrimeSonic: This will be handled by @MrPurple6411 and @ahk1221

SMLHelper/Assets/FishPrefab.cs
AlexejheroYTB: We load a proxy assembly before Assembly-CSharp.dll loads using Doorstop, we check which game QModManager was installed for, then we load the main assembly with that information already known.

PrimeSonic: I am resolving this discussion because it pertains to QModManager.  
We will continue it there on a new branch when we go back to QMM 3.0

.gitignore
```diff
@@ -259,7 +259,7 @@ paket-files/
...
 __pycache__/
 *.pyc
-Dependencies/Assembly-CSharp-firstpass.dll
-Dependencies/Assembly-CSharp.dll
-/Dependencies/Assembly-CSharp-firstpass_publicized.dll
-/Dependencies/Assembly-CSharp_publicized.dll
+Assembly-CSharp-firstpass.dll
+Assembly-CSharp-firstpass_publicized.dll
+Assembly-CSharp.dll
+Assembly-CSharp_publicized.dll
```
AlexejheroYTB: Okay, what???
What is this supposed to achieve?
This will only ignore the assembly files _if they're found in the root directory_

SMLHelper/Handlers/CraftDataHandlerBZ.cs
```diff
@@ -0,0 +1,319 @@
+﻿#if BELOWZERO
```
AlexejheroYTB: Who came up with `BZ` and `Sub` suffixes?
That's a very poor way to organize the files in my opinion...

.gitignore
AlexejheroYTB: Ok I was wrong

SMLHelper/QModManager/Subnautica.json
```diff
@@ -1,11 +1,10 @@
 {
   "Id": "SMLHelper",
   "DisplayName": "Modding Helper (SMLHelper)",
-  "Author": "AHK1221 & PrimeSonic & AlexejheroYTB",
-  "Version": "2.3.0",
+  "Author": "The QModManager Team",
+  "Version": "2.3.1",
```
PrimeSonic: Recommending that this version (for both games) go to **2.4.0**

SMLHelper/Handler.cs
```diff
@@ -0,0 +1,65 @@
+﻿namespace SMLHelper.V2
+{
+    using Interfaces;
+
+    /// <summary>
+    /// A simple location where every SMLHelper handler class can be accessed.
+    /// </summary>
+    public static class Handler
```
PrimeSonic: Any recommendations on a better name for this?

AlexejheroYTB: `Handlers`

SMLHelper/Initializer.cs
```diff
@@ -5,10 +5,16 @@
     using Harmony;
     using Patchers;
 
+    /// <summary>
+    /// WARNING: This class is for use only by QModManager.
+    /// </summary>
     public class Initializer
```
AlexejheroYTB: Then why is it public?

SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -1,11 +1,16 @@
...
 {
+    using SMLHelper.V2.Assets;
     using UnityEngine;
-    using System.Reflection;
     using Logger = V2.Logger;
 
+    /// <summary>
+    /// This <see cref="MonoBehaviour"/> is <c>automatically</c> added to any <see cref="GameObject"/> created through <see cref="ModPrefab.GetGameObject"/>.
+    /// </summary>
     public class Fixer : MonoBehaviour, IProtoEventListener
```
AlexejheroYTB: Could there be a better name for this? `Fixer` doesn't say much about what this is supposed to do

PrimeSonic: Clone Remover, Clone Cleaner, Instantiate Deduper, Origin Saver,  etc. 


Modders don't have to ever interact with this class directly under normal usage.  
But it's here in case someone finds a need for it.

SMLHelper/Initializer.cs
PrimeSonic: Isn't it necessary for it to be public so it can be accessed by QModManager?

SMLHelper/Handler.cs
PrimeSonic: That name isn't possible because that is already the name of the namespace.

SMLHelper/Initializer.cs
AlexejheroYTB: Not if i remember correctly, let me double check

SMLHelper/Assets/CustomFabricator.cs
```diff
@@ -0,0 +1,323 @@
...
+        /// <summary>
+        /// Safely adds a new crafting node to the custom crafting tree of this fabricator.<para/>
+        /// If the item has not been patched yet, its <see cref="Spawnable.Patch"/> method will first be invoked.
+        /// </summary>
+        /// <typeparam name="T">Any type that inherits from <see cref="Spawnable"/>.</typeparam>
+        /// <param name="item">The <see cref="Spawnable"/> item to craft.</param>
+        /// <param name="parentTabId">Optional. The parent tab of this craft node.<para/>
+        /// When this value is null, the craft node will be added to the root of the craft tree.</param>
+        public void AddCraftNode<T>(T item, string parentTabId = null)
+            where T : Spawnable
```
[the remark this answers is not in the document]
PrimeSonic: Realizing now that this was supposed to accept a new type of Craftable like class.  One that doesn't assume a default Fabricator. 
To be updated.

SMLHelper/Initializer.cs
AlexejheroYTB: Apparently it has to be public. 

_To update in the QMM dev branch_

SMLHelper/Handler.cs
AlexejheroYTB: Oh, I didn't see the namespace.

In that case, why is it in the root folder and not in the Handlers folder?

PrimeSonic: In trying to make this class itself as easy to discover as possible

SMLHelper/Crafting/RecipeData.cs
```diff
@@ -0,0 +1,104 @@
...
+#pragma warning disable IDE1006 // Naming Styles - Ignored for backwards compatibility
+namespace SMLHelper.V2.Crafting
+{
+    using System.Collections.Generic;
+
+    /// <summary>
+    /// A class that fully describes a recipe for a <see cref="TechType"/> identified item.
+    /// </summary>
+    /// <seealso cref="TechData" />
+    public class RecipeData
```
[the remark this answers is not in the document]
MrPurple6411: Yes ..... BUT  I can see no advantage of doing so as that would just end up with 3 places that hold the techdata instead of just 2. Also Subnautica is said to be DONE with changes except for unity updates and the like but they could still change BelowZero's Techdata again at any update and then we would have to split them up again.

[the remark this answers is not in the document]
MrPurple6411: In this one instance due to uncertain future changes to Below Zero I do believe it should remain seperate "for now". After BZ full release if they have not muxed it about more then it should not be an issue to do it but again that would just make 3 places instead of 2 for the handling of techdata.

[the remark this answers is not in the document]
AlexejheroYTB: If both of those inherit from an abstract class we can have one set of methods for that abstract class, getting rid of most file separation clutter.

SMLHelper/Handler.cs
AlexejheroYTB: So we're getting rid of static methods?

SMLHelper/Crafting/RecipeData.cs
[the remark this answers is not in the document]
MrPurple6411: Well except that both games still handle the games side completely differently.

SMLHelper/Handler.cs
PrimeSonic: I wouldn't get rid of the static methods unless we are willing to completely break backwards compatibility.  
This here is just to make it easier for modders to find all the handlers in one place without needing to look up the documentation.

SMLHelper/Handlers/IngameMenuHandler.cs
```diff
@@ -0,0 +1,75 @@
+﻿namespace SMLHelper.V2.Handlers
+{
+    using System;
+    using SMLHelper.V2.Interfaces;
+    using SMLHelper.V2.Patchers;
+
+    /// <summary>
+    /// A handler class that offers simple ways to tap into functionality of the in game menu.
+    /// </summary>
+    public class IngameMenuHandler : IIngameMenuHandler
```
AlexejheroYTB: This could probably include a function to block the saving of the game. I can provide code if you want.

PrimeSonic: If that would be useful, then sure.  
I would suggest making these into method calls (not properties).  
Something like this:
```CS
void BlockSaving();

void AllowSaving();
```
And yes, if you can provide the code, go right ahead.

AlexejheroYTB: I would suggest an implementation like this:

```cs
void RegisterSaveBlockingEvent(Func<bool> function);
```

with an usage like this:

```cs
RegisterSaveBlockingEvent(() =>
{
	bool b = false;
    // do something with b
    return b; // true to block, false to not
});

Comment (PrimeSonic) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: That could also work.  
I can see both of these approaches having their use cases.

If you're able to code for both of them, that would be ideal.

Comment (AlexejheroYTB) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: Where would the modder call `BlockSaving` and `AllowSaving`?

Comment (PrimeSonic) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: At the beginning and end of some event she saving should be blocked.

Or the start and end of some animation sequence.
Ask @MrPurple641 about changing the outcome of the Sunbeam

Comment (AlexejheroYTB) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: I've worked with save blocking in the past.

> At the beginning and end of some event she saving should be blocked.

Comment (PrimeSonic) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: > So, SMLHelper would keep track of assemblies that called that function?
It would be up to the individual mods to re-enable saving if they disable it.

If a modder forgets to call `AllowSaving` at some point after they called `BlockSaving`, it would be no different than if their blocking function always returned true and never changed.

Comment (AlexejheroYTB) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: I meant, how would SMLHelper differentiate between two mods calling `BlockSaving` and `AllowSaving` for their own reasons?

Comment (PrimeSonic) [reply SMLHelper/Handlers/IngameMenuHandler.cs]: Get the calling assembly
```CS
Assembly mod = ReflectionHelper.CallingAssemblyByStackTrace();

Blockers ICollection<Assembly>

BlockSaving() => Blockers.Add(callingAssembly);
AllowSaving() => Blockers.Remove(callingAssembly);

If (Blockers.Count > 0)
// block saving
```

AlexejheroYTB:
```cs
[HarmonyPatch(typeof(IngameMenu))]
[HarmonyPatch("GetAllowSaving")]
static class IngameMenu_GetAllowSaving
{
    static void Postfix(ref bool __result)
    {
        __result = __result && !Mod.disableSaving;
    }
}
```

SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -193,11 +199,38 @@ internal void LoadCache()
...
+
+                string[] allText = File.ReadAllLines(savePathDir);
+                foreach (string line in allText)
+                {
+                    string[] split = line.Split(':');
+                    string name = split[0];
+                    string index = split[1];
+
+                    entriesFromDeactivatedFile.Add(Convert.ToInt32(index), name);
+                    entriesFromFile.Add(Convert.ToInt32(index), name);
```
PrimeSonic: Did you intend to load the deactivated items into the `entriesFromFile` collection?

Also, this chunk of code is duplicated for each file and could be refactored.

MrPurple6411: Yes i loaded it into the entries from file because that is what allows it to not reuse those ID's and is what allows you to add mods back into the game..    although they will need to add the mod then save and reload.

MrPurple6411: unless i add a check somewhere that when things are going into the request list it check the deactivated list and removes it from the deactivate list that way they would not need to reload

MrPurple6411: but cross mod support would still require reloading

MrPurple6411: actually no it would work without reload

MrPurple6411: if you want we could actually keep it all in one file if we add a third item per line and have it as a bool...

PrimeSonic: Using two separate files is fine. It simplifies how the data is stored.

SMLHelper/Handlers/IngameMenuHandler.cs
PrimeSonic: @AlexejheroYTB  The "block game from saving" feature will have to wait for a later update.  
It won't be included as part of this branch's release.

PR #107, opened 2019-11-30
Below Zero Saving Cache

PrimeSonic:
- Adds a new patch that allows methods to be invoked when the game is saving
- This will be used by the EnumCacheManager so that the file is only saved when the player manually saves

### DIFF
--- a/SMLHelper/Handlers/IngameMenuHandler.cs
+++ b/SMLHelper/Handlers/IngameMenuHandler.cs
@@ -36,9 +36,9 @@
         }
 
         /// <summary>
-        /// Registers a simple <see cref="Action"/> method to invoke whenever the player saves the game via the in game menu.
+        /// Registers a simple <see cref="Action"/> method to invoke the <c>first time</c> the player saves the game via the in game menu.
         /// </summary>
-        /// <param name="onSaveAction">The method to invoke.</param>
+        /// <param name="onSaveAction">The method to invoke. This action will not be invoked a second time.</param>
         public static void RegisterOneTimeUseOnSaveEvent(Action onSaveAction)
         {
             Main.RegisterOneTimeUseOnSaveEvent(onSaveAction);
--- a/SMLHelper/Handlers/IngameMenuHandler.cs
+++ b/SMLHelper/Handlers/IngameMenuHandler.cs
@@ -1,0 +1,75 @@
+﻿namespace SMLHelper.V2.Handlers
+{
+    using System;
+    using SMLHelper.V2.Interfaces;
+    using SMLHelper.V2.Patchers;
+
+    /// <summary>
+    /// A handler class that offers simple ways to tap into functionality of the in game menu.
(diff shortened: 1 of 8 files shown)


SMLHelper/Utility/SaveUtils.cs
```diff
@@ -12,5 +15,23 @@ public static string GetCurrentSaveDataDir()
...
+        public static void RegisterOnSaveEvent(Action onSaveAction)
+        {
+            IngameMenuPatcher.OnSaveEvents += onSaveAction;
+        }
+
+        /// <summary>
+        /// Removes a method that would be invoked whenever the game is saved via the in-game menu.
+        /// </summary>
+        /// <param name="onSaveAction">The method invoked.</param>
+        public static void DeregisterOnSaveEvent(Action onSaveAction)
```
AlexejheroYTB: Unregister

SMLHelper/Utility/SaveUtils.cs
```diff
@@ -12,5 +15,24 @@ public static string GetCurrentSaveDataDir()
         {
             return SaveLoadManager.temporarySavePath;
         }
+
+        /// <summary>
+        /// Registers a simple <see cref="Action"/> method to invoke whenever the player saves the game via the in game menu.
+        /// </summary>
+        /// <param name="onSaveAction">The method to invoke.</param>
+        public static void RegisterOnSaveEvent(Action onSaveAction)
```
AlexejheroYTB: This should be moved to IngameMenuHandler

PrimeSonic: Created `IngameMenuHandler` and moved the new methods to this space.

SMLHelper/Handlers/IngameMenuHandler.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// <param name="onSaveAction">The method invoked.</param>
+        public static void UnregisterOnSaveEvent(Action onSaveAction)
+        {
+            Main.UnregisterOnSaveEvent(onSaveAction);
+        }
+
+        /// <summary>
+        /// Registers a simple <see cref="Action"/> method to invoke whenever the player saves the game via the in game menu.
+        /// </summary>
+        /// <param name="onSaveAction">The method to invoke.</param>
```
AlexejheroYTB: The summary needs to reflect the fact that this method is one-time. Also, I think you could merge `RegisterOnSaveEvent` with `RegisterOneTimeUseOnSaveEvent`, and have an optional `bool` parameter that defaults to `false` dictating whether that event is one-time or not.

PrimeSonic: I considered merging the two methods with an optional parameter, but then it complicated the documentation and explanation of the `UnregisterOnSaveEvent` method.

To keep things simple, I thought it best to just make a separate method with clear documentation.

PR #108, opened 2019-12-07
Fragment/loot handling

ahk1221:

  - Added LootDistribution, WorldEntityDatabase, PDA Scanner and PDA Encyclopedia handlers and patcher methods.
  - This allows modders to make prefabs spawn naturally in the world, based on biomes and probabilities.
  - You can also change loot distribution for existing prefabs.

### DIFF
--- a/SMLHelper/Handlers/LootDistributionHandler.cs
+++ b/SMLHelper/Handlers/LootDistributionHandler.cs
@@ -2,6 +2,8 @@
 {
     using Interfaces;
     using Patchers;
+    using System.Collections.Generic;
+    using UWE;
 
     /// <summary>
     /// A handler that manages Loot Distribution.
@@ -25,9 +27,64 @@
             Main.AddCustomLootDistData(data, classId);
         }
 
+        /// <summary>
+        /// Edits Loot Distribution data for existing prefabs, for e.g. original game prefabs.
+        /// </summary>
+        /// <param name="classID">The ClassID of the prefab. If unsure, use CraftData.GetClassIdForTechType.</param>
+        /// <param name="biome">The biome, in which this prefab is spawning, or being set to spawn in.</param>
+        /// <param name="probability">The probability for which this prefab should spawn for. Between 0 and 1.</param>
+        /// <param name="count">The count of the prefab. Usually just 1.</param>
+        public static void EditLootBiomeData(string classID, BiomeType biome, float probability, int count)
+        {
+            Main.EditLootBiomeData(classID, biome, probability, count);
+        }
+
         void ILootDistributionHandler.AddCustomLootDistData(LootDistributionData.SrcData data, string classId)
         {
             LootDistributionPatcher.CustomSrcData.Add(classId, data);
+        }
+
+        void ILootDistributionHandler.EditLootBiomeData(string classid, BiomeType biome, float probability, int count)
+        {
+            LootDistributionData.SrcData srcData;
+            if(!LootDistributionPatcher.CustomSrcData.TryGetValue(classid, out srcData))
+            {
+                LootDistributionPatcher.CustomSrcData[classid] = (srcData = new LootDistributionData.SrcData());
+
+                List<LootDistributionData.BiomeData> biomeDistribution = new List<LootDistributionData.BiomeData>();
+                biomeDistribution.Add(new LootDistributionData.BiomeData()
+                {
+                    biome = biome,
+                    probability = probability,
+                    count = count
+                });
+
+                srcData.distribution = biomeDistribution;
+
+                return;
+            }
+
+            for(int i = 0; i < srcData.distribution.Count; i++)
+            {
+                LootDistributionData.BiomeData distribution = srcData.distribution[i];
+
+                if(distribution.biome == biome)
+                {
+                    distribution.count = count;
+                    distribution.probability = probability;
+
+                    return;
+                }
+            }
+
+            // If we reached this point, that means the srcData is present, but the biome in the distribution is not.
+            // Lets add it manually.
+            srcData.distribution.Add(new LootDistributionData.BiomeData()
+            {
+                biome = biome,
+                probability = probability,
+                count = count
+            });
         }
     }
 }
--- a/SMLHelper/Handlers/LootDistributionHandler.cs
+++ b/SMLHelper/Handlers/LootDistributionHandler.cs
@@ -1,0 +1,33 @@
+﻿namespace SMLHelper.V2.Handlers
+{
+    using Interfaces;
+    using Patchers;
+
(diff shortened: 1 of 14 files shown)


SMLHelper/Handlers/LootDistributionHandler.cs
```diff
@@ -0,0 +1,133 @@
...
+        /// <param name="data">The <see cref="LootDistributionData.SrcData"/> that contains data related to the spawning of a prefab, also contains the path to the prefab.</param>
+        /// <param name="classId">The classId of the prefab.</param>
+        public static void AddCustomLootDistData(LootDistributionData.SrcData data, string classId)
+        {
+            Main.AddCustomLootDistData(data, classId);
+        }
+
+        /// <summary>
+        /// Adds in a custom entry into the Loot Distribution of the game.
+        /// You must also add the <see cref="WorldEntityInfo"/> into the <see cref="WorldEntityDatabase"/> using <see cref="WorldEntityDatabaseHandler"/>.
```
PrimeSonic: Would it be possible to combine all of these methods somehow into something that puts all the necessary data into one place?  
Maybe add new SMLHelper class that has properties for all of these items, just to make it easier to to everything necessary to add new things to the loot distribution.

SMLHelper/Patchers/PDAEncyclopediaPatcher.cs
```diff
@@ -0,0 +1,31 @@
+﻿namespace SMLHelper.V2.Patchers
+{
+    using Harmony;
+    using System;
+    using System.Collections.Generic;
+
+    internal class PDAEncyclopediaPatcher
+    {
+        internal static SelfCheckingDictionary<string, PDAEncyclopedia.EntryData> CustomEntryData = new SelfCheckingDictionary<string, PDAEncyclopedia.EntryData>("CustomEntryData");
```
PrimeSonic: Make this either a `readonly` field or a `{ get; }` only property

SMLHelper/Patchers/PdaPatcher.cs
```diff
@@ -9,6 +10,7 @@ internal class PDAPatcher
     {
         internal static readonly Dictionary<TechType, int> FragmentCount = new Dictionary<TechType, int>();
         internal static readonly Dictionary<TechType, float> FragmentScanTime = new Dictionary<TechType, float>();
+        internal static SelfCheckingDictionary<TechType, PDAScanner.EntryData> CustomEntryData = new SelfCheckingDictionary<TechType, PDAScanner.EntryData>("CustomEntryData");
```
PrimeSonic: Make this either a `readonly` field or a `{ get; }` only property

SMLHelper/Patchers/WorldEntityDatabasePatcher.cs
```diff
@@ -0,0 +1,32 @@
+﻿namespace SMLHelper.V2.Patchers
+{
+    using Harmony;
+    using System;
+    using UWE;
+
+    internal class WorldEntityDatabasePatcher
+    {
+        internal static SelfCheckingDictionary<string, WorldEntityInfo> CustomWorldEntityInfos = new SelfCheckingDictionary<string, WorldEntityInfo>("CustomWorldEntityInfo");
```
PrimeSonic: Make this either a `readonly` field or a `{ get; }` only property

SMLHelper/Patchers/WorldEntityDatabasePatcher.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        private static bool Prefix(string classId, ref WorldEntityInfo info, ref bool __result)
+        {
+            foreach(var entry in CustomWorldEntityInfos)
+            {
+                if(entry.Key == classId)
+                {
+                    __result = true;
+                    info = entry.Value;
+                    return false;
```
PrimeSonic: Why are we prefix patching a dictionary lookup instead of just adding our custom entries to the `infos` dictionary directly?

SMLHelper/Patchers/LootDistributionPatcher.cs
```diff
@@ -0,0 +1,119 @@
...
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(LootDistributionPatcher), "InitializePostfix")));
+
+            Logger.Log("LootDistributionPatcher is done.", LogLevel.Debug);
+        }
+
+        private static void InitializePostfix(LootDistributionData __instance)
+        {
+            foreach(var entry in CustomSrcData)
+            {
+                if(__instance.srcDistribution.ContainsKey(entry.Key))
```
PrimeSonic: `srcDistribution` is a Dictionary.  
Use [TryGetValue](https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.8) instead of a ContainsKey followed by another key lookup when accessing by index.

```CS
if (__instance.srcDistribution.TryGetValue(entry.Key, out LootDistributionData.SrcData srcData))
```

SMLHelper/Patchers/LootDistributionPatcher.cs
```diff
@@ -0,0 +1,119 @@
...
+                                dstData.prefabs = new List<LootDistributionData.PrefabData>();
+                                __instance.dstDistribution.Add(biome, dstData);
+                            }
+
+                            LootDistributionData.PrefabData prefabData = new LootDistributionData.PrefabData();
+                            prefabData.classId = classId;
+                            prefabData.count = count;
+                            prefabData.probability = probability;
+                            dstData.prefabs.Add(prefabData);
+                        }
```
PrimeSonic: To make this easier to read, refactor the code of the if and the else blocks into their own methods

SMLHelper/Patchers/WorldEntityDatabasePatcher.cs
ahk1221: I tried patching the constructor as there is no "initialize" method, so to speak, which didn't work for some reason (maybe the constructor was already called at that point?). And anyways, the dictionary is never looked up by the game's code directly, all calls are made to TryGetInfo.

SMLHelper/Handlers/LootDistributionHandler.cs
ahk1221: Yes, that would be `Fragment` class I was talking to you earlier about. Or are you talking about something else?

SMLHelper/Patchers/WorldEntityDatabasePatcher.cs
PrimeSonic: Yeah, doing a postfix patch on the constructor would be the better option since there are no Initialize methods here.

PR #112, opened 2019-12-15
nameof_Overhaul

PrimeSonic:
- Updates all string literals used in Reflection class with `nameof`
- BugFix: Removing the redundant `ResourcesPatcher` that would break in Below Zero


SMLHelper/Initializer.cs
```diff
@@ -48,7 +48,6 @@ internal static void Initialize()
             CraftTreePatcher.Patch(harmony);
             DevConsolePatcher.Patch(harmony);
             LanguagePatcher.Patch(harmony);
-            ResourcesPatcher.Patch(harmony);
```
AlexejheroYTB: This is also handled in #113. Expect conflicts.

PR #114, opened 2019-12-19
ConsoleParserFix

PrimeSonic:
Fixes the bug that prevented modded techtypes from being spawned in via console


AlexejheroYTB:
Test

PR #115, opened 2019-12-19
Dev December 2019 - SMLHelper v2.5.0

PrimeSonic:
#SMLHelper v2.5.0
  - New handler added `LootDistributionHandler`
  - New handler added `PDAEncyclopediaHandler`
  - New handler added `WorldEntityDatabaseHandler`
  - New methods added to `PDAHandler`
  - Removed redundant/deprecated ResourcePatcher
  - Better compile time checks against changes to the Assembly-CSharp code
  - `SpritePatcher` now capable of patching Subnautica, Below Zero, and Experimental branches
  - Fixed a bug introduced in the previous release that caused modded items to no longer spawn via console commands


SMLHelper/Handlers/LootDistributionHandler.cs
```diff
@@ -0,0 +1,145 @@
...
+        /// </summary>
+        /// <param name="data">The <see cref="LootDistributionData.SrcData"/> that contains data related to the spawning of a prefab, also contains the path to the prefab.</param>
+        /// <param name="classId">The classId of the prefab.</param>
+        /// <param name="info">The WorldEntityInfo of the prefab. For more information on how to set this up, see <see cref="WorldEntityDatabaseHandler"/>.</param>
+        public static void AddLootDistributionData(string classId, LootDistributionData.SrcData data, WorldEntityInfo info = null)
+        {
+            Main.AddLootDistributionData(data, classId);
+
+            if (info != null)
+                WorldEntityDatabaseHandler.AddCustomInfo(classId, info);
```
AlexejheroYTB: Why is this present in `LootDistributionHandler.AddLootDistributionData` but not in `LootDistributionHandler.Main.AddLootDistributionData`? Shouldn't static methods behave _exactly_ the same as instance methods in the case of dependency injection? (So call them without doing any additional stuff)

SMLHelper/Patchers/SpritePatcher.cs
```diff
@@ -1,27 +1,58 @@
...
+    using System.Reflection;
+    using Assets;
 
     internal class SpritePatcher
     {
+        // The groups field is present in Subnautica and BZ Stable.
+        private static readonly FieldInfo groupsInfo = typeof(SpriteManager).GetField("groups", BindingFlags.Static | BindingFlags.NonPublic);
+
+        // In BZ Experimental, it is replaced by the atlases field.
+        private static readonly FieldInfo atlasesInfo = typeof(SpriteManager).GetField("atlases", BindingFlags.Static | BindingFlags.NonPublic);
```
AlexejheroYTB: Publicized assembly?

PrimeSonic: From the main description
> SpritePatcher now capable of patching Subnautica, Below Zero, and Experimental branches

Yes, this was necessary to make that possible.  
It won't affect in-game performance as this only has to run once during patching.

PR #120, opened 2020-01-04
Play .wav files globally - AudioUtils class

AlexejheroYTB:


SMLHelper/Utility/AudioUtils.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// <returns></returns>
+        public static RESULT PlaySound(string path)
+        {
+            Sound newSound = new Sound();
+            ChannelGroup channels = new ChannelGroup();
+            Channel newChannel = new Channel();
+            var system = FMODUnity.RuntimeManager.LowlevelSystem;
+            system.getMasterChannelGroup(out channels);
+            system.createSound(path, MODE.DEFAULT, out newSound);
+            return system.playSound(newSound, channels, false, out newChannel);
```
PrimeSonic: Would it be possible to make methods like
```CS
Sound CreateSoundFromFile(string filePath);
RESULT PlaySound(Sound sound);
```
So that sounds that are expected to be reused often could be created once and reused.

MrPurple6411:
Made a Test mod and it works beautifully.

PR #122, opened 2020-01-05
Enable TooltipPatcher

AlexejheroYTB:


AlexejheroYTB:
Tested and works both for Subnautica and Below Zero

PR #125, opened 2020-01-07
Spawnable updates

PrimeSonic:
- Adds the new `Equipable` class
  - Adds the `SizeInInventory` property to `Spawnable`

### DIFF
--- a/SMLHelper/Assets/Spawnable.cs
+++ b/SMLHelper/Assets/Spawnable.cs
@@ -79,10 +79,11 @@
             this.FriendlyName = friendlyName;
             this.Description = description;
 
-            CorePatchEvents += () => this.PrefabHandler.RegisterPrefab(this);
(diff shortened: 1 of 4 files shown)


SMLHelper/Assets/Spawnable.cs
```diff
@@ -72,6 +81,11 @@ protected Spawnable(string classId, string friendlyName, string description)
 
             CorePatchEvents += () => this.PrefabHandler.RegisterPrefab(this);
             CorePatchEvents += () => this.SpriteHandler.RegisterSprite(this.TechType, GetItemSprite());
+            CorePatchEvents += () =>
```
AlexejheroYTB: Why are these separate?

PrimeSonic: No particular reason

PR #131, opened 2020-02-06
HullPlate asset class

AlexejheroYTB:
Added a `HullPlate` asset class


SMLHelper/Assets/HullPlate.cs
```diff
@@ -0,0 +1,76 @@
...
+    /// <seealso cref="PdaItem" />
+    /// <seealso cref="Spawnable"/>
+    /// <seealso cref="Buildable"/>
+    public abstract class HullPlate : Buildable
+    {
+        /// <summary>
+        /// Gets the <see cref="Texture2D"/> that the hull plate should have
+        /// </summary>
+        /// <returns></returns>
+        public abstract Texture2D GetTexture();
```
PrimeSonic: The base class `Spawnable` from which this class inherits, has a virtual implementation for loading its [icon sprite](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Assets/Spawnable.cs#L131) from a PNG file at run time.

Could this not do the same?  
Could this method have a virtual implementation, also making use of our [ImageUtils](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Utility/ImageUtils.cs), such that it could detect an image file by name?  
If so, allow the texture file to also have a default value that can be overridden.

SMLHelper/Assets/HullPlate.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Gets the prefab game object.
+        /// You are not required to override this method if you are using <see cref="HullPlate"/>
+        /// </summary>
+        /// <returns>The game object to be instantiated into a new in-game entity.</returns>
+        public override GameObject GetGameObject()
+        {
+            GameObject obj = GameObject.Instantiate(CraftData.GetPrefabForTechType(TechType.DioramaHullPlate));
+
+            MeshRenderer meshRenderer = obj.FindChild("Icon").GetComponent<MeshRenderer>();
+            meshRenderer.material.mainTexture = this.GetTexture();
```
PrimeSonic: Consider if we should cache the results of the `GetTexture` method for quick reuse instead of re-running the method every time.

AlexejheroYTB: Should I make a `Cacheable<T>` class in the `Utility` namespace?
We could use that in a lot of places.

SMLHelper/Assets/HullPlate.cs
AlexejheroYTB: I personally don't like SMLHelper's way of making optional properties required or making required properties optional.

This method will most definitely be used in conjunction with `ImageUtils`, but that doesn't mean that I should turn this into a string property to the path of the file.

Maybe the modder wants to get the texture from an asset bundle, or idk. If they want to load the image from a file, what's stopping them form using `ImageUtils` themselves?

(More on this topic later in the Discord)

PR #133, opened 2020-02-26
Auto save dynamic patching

MrPurple6411:

  - Auto Save Caches enabled again.
  -Loading and saving of cache to handle auto saving.
  -Enabled Cache Sorting for easier reading.
  -Dynamic patching for Craft data patching to enable possible real-time changes of crafting recipes.
  -Added an onQuit event Register that "might" be useful.


MrPurple6411:
Nope.

MrPurple6411:
Now working with both games again.

SMLHelper/Assets/ModPrefab.cs
```diff
@@ -22,6 +22,7 @@ internal static void Add(ModPrefab prefab)
         }
 
         internal static IEnumerable<ModPrefab> Prefabs => PreFabsList;
+        internal static int PrefabCount = PreFabsList.Count;
```
PrimeSonic: I believe you intended this to be a get-only property instead of a field that's written to once and  never updated.
```CS
internal static int PrefabCount => PreFabsList.Count;
```

SMLHelper/Handlers/CraftDataHandler_Subnautica.cs
```diff
@@ -206,12 +206,11 @@ void ICraftDataHandler.AddBuildable(TechType techType)
...
         {
-            if (CraftDataPatcher.CustomTechData.TryGetValue(techType, out ITechData moddedTechData))
+            if (!CraftDataPatcher.CustomTechData.TryGetValue(techType, out ITechData moddedTechData))
             {
-                return moddedTechData;
+                moddedTechData = CraftData.Get(techType, true);
             }
-
-            return null;
+            return moddedTechData;
```
PrimeSonic: While I can understand why this could be useful,  
this change effectively is means that `GetModdedTechData` is now allowed to return `TechData` instances that didn't come from mods.  

Rather than change how `ICraftDataHandler.GetModdedTechData` works,  
I would recommend instead creating a new method that does this.

```CS
ITechData GetTechData(TechType techType)
```

SMLHelper/Initializer.cs
```diff
@@ -20,12 +20,14 @@ public class Initializer
...
         {
-            Logger.Initialize();
+            Logger.Initialize(); 
 #if SUBNAUTICA
             Logger.Log($"Loading v{Assembly.GetExecutingAssembly().GetName().Version} for Subnautica", LogLevel.Info);
 #elif BELOWZERO
             Logger.Log($"Loading v{Assembly.GetExecutingAssembly().GetName().Version} for BelowZero", LogLevel.Info);
 #endif
+            TechTypePatcher.cacheManager.LoadCache();
+            CraftTreeTypePatcher.cacheManager.LoadCache();
```
PrimeSonic: Add a Debug level log before these new lines indicating that cache files are being loaded

SMLHelper/Initializer.cs
```diff
@@ -69,6 +71,8 @@ private static void Initialize()
             WorldEntityDatabasePatcher.Patch(harmony);
             IngameMenuPatcher.Patch(harmony);
             TooltipPatcher.Patch(harmony);
+            TechTypePatcher.cacheManager.SaveCache();
+            CraftTreeTypePatcher.cacheManager.SaveCache();
```
PrimeSonic: Add a Debug level log before these new lines indicating that cache files are being saved

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
+            PrefabAltered = 0;
+        }
+
+        private static void CraftDataGetPrefabPrefix()
+        {
+            if (ModPrefab.PrefabCount > 0)
+            {
+                ModPrefab prefab = ModPrefab.Prefabs.FirstOrFallback(null);
+                if (!CraftData.techMapping.ContainsKey(prefab.TechType) || !CraftData.entClassTechTable.ContainsKey(prefab.ClassID))
+                    CraftDataPrefabIDCachePostfix();
```
PrimeSonic: Can you add a comment explaining why the Prefix patch method is invoking the Postfix patch method?

Maybe also explain it here as well.

SMLHelper/Utility/EnumCacheManager.cs
```diff
@@ -27,10 +27,10 @@ internal class EnumCacheManager<T> where T : Enum
...
         private class DoubleKeyDictionary : IEnumerable<KeyValuePair<int, string>>
         {
-            private readonly Dictionary<int, string> MapIntString = new Dictionary<int, string>();
-            private readonly Dictionary<T, string> MapEnumString = new Dictionary<T, string>();
-            private readonly Dictionary<string, T> MapStringEnum = new Dictionary<string, T>(StringComparer.InvariantCultureIgnoreCase);
-            private readonly Dictionary<string, int> MapStringInt = new Dictionary<string, int>(StringComparer.InvariantCultureIgnoreCase);
+            private readonly SortedDictionary<int, string> MapIntString = new SortedDictionary<int, string>();
+            private readonly SortedDictionary<T, string> MapEnumString = new SortedDictionary<T, string>();
+            private readonly SortedDictionary<string, T> MapStringEnum = new SortedDictionary<string, T>(StringComparer.InvariantCultureIgnoreCase);
+            private readonly SortedDictionary<string, int> MapStringInt = new SortedDictionary<string, int>(StringComparer.InvariantCultureIgnoreCase);
```
PrimeSonic: What benefits does the `SortedDictionary` bring to this?  

Would the `SelfCheckingDictionary` have had something useful to offer here?

SMLHelper/Patchers/IngameMenuPatcher.cs
```diff
@@ -7,21 +7,32 @@
...
 
         public static void Patch(HarmonyInstance harmony)
         {
             harmony.Patch(AccessTools.Method(typeof(IngameMenu), nameof(IngameMenu.SaveGame)),
-                postfix: new HarmonyMethod(AccessTools.Method(typeof(IngameMenuPatcher), nameof(InvokeEvents))));
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(IngameMenuPatcher), nameof(InvokeSaveEvents))));
+            harmony.Patch(AccessTools.Method(typeof(IngameMenu), nameof(IngameMenu.QuitGame)),
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(IngameMenuPatcher), nameof(InvokeQuitEvents))));
+            harmony.Patch(AccessTools.Method(typeof(IngameMenu), nameof(IngameMenu.QuitGameAsync)),
+                postfix: new HarmonyMethod(AccessTools.Method(typeof(IngameMenuPatcher), nameof(InvokeQuitEvents))));
```
PrimeSonic: `InvokeQuitEvents` gets invoked for both `QuitGame` and `QuitGameAsync`?

Is it unreliable to patch only one of these?

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
+            if (TechDataAltered < CustomTechData.Keys.Count)
+                AddCustomDataToOriginalDictionaries();
+        }
+
+        private static void GetHarvestOutputPrefix()
+        {
+            if (CustomHarvestOutputList.Count > 0 && HarvestOutputAltered < CustomHarvestOutputList.Count)
+            {
+                PatchUtils.PatchDictionary(CraftData.harvestOutputList, CustomHarvestOutputList);
+                HarvestOutputAltered = CustomHarvestOutputList.Count;
```
PrimeSonic: `HarvestOutputAltered` and the various other `...Altered` fields are being assigned with Count values but are being used more like boolean checks.

Is the numeric value of these new fields actually important?  
Or would be be enough to handle them as `bool`s instead?

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
 
-                var techDataInstance = new CraftData.TechData
+                if (techExists)
                 {
-                    _techType = techType,
-                    _craftAmount = smlTechData.craftAmount
-                };
+                    sameData = smlTechData.craftAmount == techData.craftAmount &&
+                        smlTechData.ingredientCount == techData.ingredientCount &&
+                        smlTechData.linkedItemCount == techData.linkedItemCount;
```
PrimeSonic: This isn't a good assumption to make.  
Just because the number of ingredients, linked items, and crafted amount are the same doesn't necessarily mean that the there wasn't a change.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
+                                sameData = false;
+                                break;
+                            }
+                        }
 
-                var ingredientsList = new CraftData.Ingredients();
+                    if (sameData)
+                        for (int i = 0; i < smlTechData.linkedItemCount; i++)
+                        {
+                            if (smlTechData.GetLinkedItem(i) != techData.GetLinkedItem(i))
```
PrimeSonic: This may be dangerous if the number of linked items in `smlTechData` is larger than the ones in `techData`.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
+                    
+                    if(sameData)
+                        for (int i = 0; i < smlTechData.ingredientCount; i++)
+                        {
+                            if (smlTechData.GetIngredient(i).techType != techData.GetIngredient(i).techType)
+                            {
+                                sameData = false;
+                                break;
+                            }
+                            if (smlTechData.GetIngredient(i).amount != techData.GetIngredient(i).amount)
```
PrimeSonic: This may be dangerous if the number of ingredients in `smlTechData` is larger than the ones in `techData`.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -1,97 +1,301 @@
...
+                    {
+                        CraftData.techData.Remove(techType);
+                        Logger.Log($"{techType} TechType already existed in the CraftData.techData dictionary. Original value was replaced.", LogLevel.Warn);
+                        replaced++;
+                        TechDataAltered++;
+                    }
+                    else
+                    {
+                        added++;
+                        TechDataAltered++;
```
PrimeSonic: Instead of incrementing `TechDataAltered` every time you also increment `replaced` and `added`,  
you can simply set  the value once at the end of the loop.
```CS
TechDataAltered = added + replaced;
```

SMLHelper/Assets/ModPrefab.cs
MrPurple6411: Woops yup

SMLHelper/Handlers/CraftDataHandler_Subnautica.cs
MrPurple6411: Ahh yes this was a change when I was trying to make it put the mod data into the games cache and then wipe the custom so that I could just tell it to patch when there was data in the custom data dictionary. Must have forgot to revert that when I changed from that thought

SMLHelper/Initializer.cs
MrPurple6411: Will do

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: Will fix the naming because the prefix is kind of a verify patch required and the postfix is do it anyway.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: Bool would work better if we added a bool false in the methods that add to those.  Which would have been a far better way to do it.   Will change.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: That's why if goes further and checks the ingredients and linked items one by one but only if that comes back true

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: Which is why it only triggers if they are the same from in the check before this

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: I actually moved away from using that check. Just forgot to remove it. Will do

SMLHelper/Patchers/IngameMenuPatcher.cs
MrPurple6411: Might be on on just quit

SMLHelper/Utility/EnumCacheManager.cs
MrPurple6411: The sorteddictionary makes it so that things in the cache are sorted by the number value of the techdata so things dont jump all over the place when removing and adding

SMLHelper/Assets/ModPrefab.cs
MrPurple6411: Turns out this was not required as the prefab cache is wiped and rewritten every time the game checks for a prefab so our code needs to run at that time too. Kinda makes a cache useless but hey UWE

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: Actually ended up going another way altogether Please RECHECK.

PrimeSonic: The new method looks good, but now you've got all those leftover `int` fields to clean up.

MrPurple6411: 1790ea9

SMLHelper/Patchers/IngameMenuPatcher.cs
MrPurple6411: Just patches quit now.

SMLHelper/Patchers/CraftDataPatcher_BelowZero.cs
```diff
@@ -39,7 +50,7 @@ private static void AddCustomTechDataToOriginalDictionary()
                 bool techDataExists = TechData.entries.ContainsKey(techType);
                 if (techDataExists && TechData.entries[techType] != CustomTechData[techType])
                 {
-                    if (TechData.TryGetValue(techType, out JsonValue originalData))
+                    if (TechData.entries.ContainsKey(techType))
```
PrimeSonic: `(TechData.entries.ContainsKey(techType)` is checked twice here.  
This nested if-block look redundant when `techDataExists` already has the value it is checking.

SMLHelper/Patchers/CraftDataPatcher_BelowZero.cs
```diff
@@ -39,7 +50,7 @@ private static void AddCustomTechDataToOriginalDictionary()
                 bool techDataExists = TechData.entries.ContainsKey(techType);
                 if (techDataExists && TechData.entries[techType] != CustomTechData[techType])
```
PrimeSonic: `CustomTechData[techType]` is used multiple times in this method.  
A local variable could replace the multiple dictionary lookups with a single one.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -22,76 +26,314 @@ internal static void AddToCustomTechData(TechType techType, ITechData techData)
...
+                PatchUtils.PatchDictionary(CraftData.harvestFinalCutBonusList, CustomFinalCutBonusList);
+            }
+        }
+
+        private static void GetItemSizePrefix(TechType techType)
+        {
+            if (CraftData.itemSizes.ContainsKey(techType) && CustomItemSizes.ContainsKey(techType))
+            {
+                if ((CraftData.itemSizes[techType].x != CustomItemSizes[techType].x || CraftData.itemSizes[techType].y != CustomItemSizes[techType].y))
+                    CraftData.itemSizes[techType] = CustomItemSizes[techType];
```
PrimeSonic: Throughout these patching methods, I see the same pattern used several times
```CS
if (CraftData.somedictionary.ContainsKey(techType))
{
    /*do something with*/ CraftData.somedictionary[techType]
}
```

This pattern results in doing the key-lookup twice.  
Instead, use the `TryGetValue` method from `IDictionary` whenever possible.

```CS
if (CraftData.somedictionary.TryGetValue(techType, out var value))
{
    /*do something with*/ value
}
```

SMLHelper/Patchers/CraftDataPatcher_BelowZero.cs
MrPurple6411: yes it is. altered this a bit.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
MrPurple6411: looks like that would work nicely here. forgot that the try get value was only patched in the BZ side so i can use it here without causing an infinite loop.

SMLHelper/Patchers/CraftDataPatcher_BelowZero.cs
```diff
@@ -19,15 +20,28 @@ internal static void AddToCustomTechData(TechType techType, JsonValue techData)
...
-        private static void TechDataCachePostfix()
+        private static void CheckPatchRequired(TechType techType)
         {
-            AddCustomTechDataToOriginalDictionary();
+            if(CustomTechData.TryGetValue(techType, out JsonValue SMLTechData))
+            {
+                if (TechData.entries.ContainsKey(techType))
+                {
+                    if(SMLTechData != TechData.entries[techType])
+                        AddCustomTechDataToOriginalDictionary();
```
PrimeSonic: Another unnecessary key check
```CS
if (TechData.entries.TryGetValue(techType, out var ogTech))
{
    if(SMLTechData != ogTech)
        AddCustomTechDataToOriginalDictionary();
```

PR #138, opened 2020-03-24
SMLHelper 2.6.1

MrPurple6411:
- Fixed a Null Ref Exception when trying to get unmodified/vanilla TechData from `GetTechData`.
- `ModdedTechTypeExists` and `TryGetModdedTechType` no longer return `true` for deactivated TechTypes
- Optional `format` parameter for `LoadSpriteFromFile` in Subnautica now handled

### DIFF
--- a/SMLHelper/Handlers/CraftDataHandler_Subnautica.cs
+++ b/SMLHelper/Handlers/CraftDataHandler_Subnautica.cs
@@ -234,41 +234,38 @@ namespace SMLHelper.V2.Handlers
         /// <returns>Returns TechData if it exists; Otherwise, returns <c>null</c>.</returns>
         TechData ICraftDataHandler.GetTechData(TechType techType)
         {
-
             if (CraftDataPatcher.CustomTechData.TryGetValue(techType, out ITechData iTechData))
             {
-                TechData techData = new TechData() {craftAmount = iTechData.craftAmount};
-
-                for(int i = 0; i < iTechData.ingredientCount; i++)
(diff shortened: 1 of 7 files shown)


SMLHelper/Handlers/CraftDataHandler_Subnautica.cs
```diff
@@ -235,36 +235,36 @@ ITechData ICraftDataHandler.GetModdedTechData(TechType techType)
...
-                    IIngredient ingredient = td.GetIngredient(i);
+                    IIngredient ingredient = iTechData.GetIngredient(i);
                     Ingredient smlingredient = new Ingredient(ingredient.techType, ingredient.amount);
                     techData.Ingredients.Add(smlingredient);
                 }
-                for (int i = 0; i < td.linkedItemCount; i++)
+                for (int i = 0; i < iTechData.linkedItemCount; i++)
                 {
-                    techData.LinkedItems.Add(moddedTechData.GetLinkedItem(i));
+                    techData.LinkedItems.Add(iTechData.GetLinkedItem(i));
```
zorgesho: @MrPurple6411 You had bug there because you copy-pasted code. It is better to just use local function for these two identical pieces of code.

PrimeSonic: Good call. I'm not sure how I didn't see that myself before.

No need to update the release package on Nexus as it will be functionally identical to this refactor.

PR #139, opened 2020-03-24
Simple refactor of Duplicate code in Craftdatahandler for BZ

MrPurple6411:

  - Simple refactor of Duplicate code in Craftdatahandler for BZ
  -
  -

### DIFF
--- a/SMLHelper/Handlers/CraftDataHandler_BelowZero.cs
+++ b/SMLHelper/Handlers/CraftDataHandler_BelowZero.cs
@@ -200,7 +200,7 @@ namespace SMLHelper.V2.Handlers
         {
             if (CraftDataPatcher.CustomTechData.TryGetValue(techType, out JsonValue techData))
             {
-                return ConvertToTechData(techData);
+                return ConvertToRecipeData(techData);
             }
 
             if (!TechData.Contains(TechType.Knife))
@@ -210,13 +210,13 @@ namespace SMLHelper.V2.Handlers
 
             if (TechData.TryGetValue(techType, out techData))
             {
-                return ConvertToTechData(techData);
+                return ConvertToRecipeData(techData);
             }
 
             return null;
         }
 
-        private static RecipeData ConvertToTechData(JsonValue techData)
+        private static RecipeData ConvertToRecipeData(JsonValue techData)
         {
             RecipeData currentRecipeData = new RecipeData();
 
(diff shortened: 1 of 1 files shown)


zorgesho:
Still lot of copypasta in that file

SMLHelper/Handlers/CraftDataHandler_BelowZero.cs
```diff
@@ -195,93 +195,64 @@ void ICraftDataHandler.AddLinkedItems(TechType techType, ICollection<TechType> l
         /// Make sure your mod is loading after the mod whose TechData you are trying to access.
         /// </summary>
         /// <param name="techType">The TechType whose TechData you want to access.</param>
-        /// <returns>The ITechData from the modded item if it exists; Otherwise, returns <c>null</c>.</returns>
+        /// <returns>The RecipeData from the item if it exists; Otherwise, returns <c>null</c>.</returns>
```
PrimeSonic: Opening up a comment thread so it can be resolved later.

I'm basically in agreement with @zorgesho .
If this change is to de-duplicate code, we should take the opportunity to do as much as possible.
It will make things cleaner and easier to maintain in the long run.
Not to mention that it'll save us from copy/paste errors as well.

If you need any help with refactoring, let me know and I can jump on.

AlexejheroYTB:
Don't merge into master

SMLHelper/Handlers/CraftDataHandler_BelowZero.cs
PrimeSonic: Refactors made.

SMLHelper/Handlers/CraftDataHandler_BelowZero.cs
```diff
@@ -347,16 +286,7 @@ RecipeData ICraftDataHandler.GetModdedRecipeData(TechType techType)
...
-                CraftDataPatcher.CustomTechData[techType][TechData.PropertyToID("equipmentType")] = new JsonValue((int)equipmentType);
-            else
-                CraftDataPatcher.CustomTechData[techType] = new JsonValue
-                {
-                    {
-                        TechData.PropertyToID("equipmentType") ,
-                        new JsonValue((int)equipmentType)
-                    }
-                };
+            AddJsonProperty(techType, "equipmentType", new JsonValue((int)equipmentType));
```
zorgesho: I think, for the sake of perfection, you can add intermediate generic method and put `new JsonValue()` there. Other than that, it looks good.

PrimeSonic: That was my first instinct.  
Turns out, it isn't feasible since the `JsonValue` constructor only accepts a specific set of types, which don't really have anything in common to make for a clean constraint.

This would have ended up with a separate overload of the same method for each type that `JsonValue` supports.  
Rather than do that and end up with almost as much nearly duplicate code as before, I opted to just require the entire `JsonValue` instance and leave that to be handled one level up.

zorgesho: Ah, I see.. Is JsonValue is specific to BelowZero? I tried to find it before commenting, but it seems it's not part of Newtonsoft.Json and I don't have BZ yet.

PR #140, opened 2020-03-26
Added simple KeyCode utility methods that behave identically like Gam…

tobeyStraitjacket:

  - Added simple KeyCode utility methods that behave identically like GameInput.GetButton methods, except for KeyCodes and strings.
  -
  -

### DIFF
--- a/SMLHelper/Utility/KeyCodeUtils.cs
+++ b/SMLHelper/Utility/KeyCodeUtils.cs
@@ -135,5 +135,124 @@
                     }
             }
         }
+
+        internal static GameInput.InputState GetInputStateForKeyCode(KeyCode keyCode)
+        {
+            var inputState = default(GameInput.InputState);
+            if (!GameInput.clearInput && !GameInput.scanningInput)
+            {
+                for (var i = 0; i < GameInput.inputs.Count; i++)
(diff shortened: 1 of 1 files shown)


SMLHelper/Utility/KeyCodeUtils.cs
```diff
@@ -135,5 +135,124 @@ public static KeyCode StringToKeyCode(string s)
                     }
             }
         }
+
```
PrimeSonic: This all looks pretty good.

One question: Are there any considerations to be made for the small differences between Subnautica and Below Zero?

There probably aren't but I just want to be sure.

PrimeSonic: Given how there are next to no mods doing anything with controller support, I don't think this will be anything to worry about.

I'll set up new branch for the next release later and get this merged.
There's another open PR that needs my attention in the meantime.

tobeyStraitjacket: Yeah no rush. I can confirm btw that it in my testing, it works as intended in BZ.

AlexejheroYTB:
Don't merge into master

PR #142, opened 2020-03-27
Json utility

tobeyStraitjacket:

* New `SMLHelper.V2.Json` namespace:
    - Added `public interface IJsonFile` to the `Json.Interfaces` namespace - A simple interface for interacting with JSON files.
    - Added `public abstract class ConfigFile : IJsonFile` to the `Json` namespace - A simple implementation of IJsonFile for standardising config files.
    - Added `public class KeyCodeConverter : JsonConverter` to the `Json.Converters` namespace - A `Newtonsoft.Json.JsonConverter` for parsing `KeyCode`s to useful strings in JSON files for legibility.
    - Added `public static class JsonExtensions` to the `Json.ExtensionMethods` namespace - Provides convenience methods for saving/loading JSON data `<T> where T : class` object instances.
* Added `public static class JsonUtils` to the `SMLHelper.V2.Utility` namespace - Provides universal `Load` and `Save` `static` methods for saving/loading JSON data from `<T> where T : class` objects.

Confirmed working in both SN and BZ.


tobeyStraitjacket:
Finished making changes: awaiting review.

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,226 @@
...
+                catch (Exception ex)
+                {   // Ideally, what we want is to pop up a QMM Dialog letting the user know exactly what mod
+                    // and exactly what file is causing the error, with an option to open a JSON validator in
+                    // browser to help them fix the problem.
+                    // We'd also want this issue to stop QMods from considering the mod loaded properly, or
+                    // expose a method for mods to check if any mods had config file errors, so that mods
+                    // which rely on all mods loading properly before performing actions can act accordingly.
+                    // For now, without copying QMM's dialog class into SML, this is the best we can do.
+                    Logger.Announce($"Could not parse JSON file: {path}", LogLevel.Error, true);
+                    throw ex;
```
PrimeSonic: Honestly, I wouldn't bother.

If something is wrong with the config file as read, I would simply replace it with the default config instead.  
Sure, we should log the error, but I wouldn't stop the the mod from loading over that.

If this seems to heavy handed, you could make it another option (defaulting to _yes_).

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,226 @@
...
+        ///         // property on subsequent runs.
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        /// <seealso cref="Load{T}(T, bool)"/>
+        /// <seealso cref="Load(bool)"/>
+        /// <seealso cref="Save"/>
+        public static T Load<T>(bool saveDefaultConfigIfNotExist = true) where T : ModConfig
+            => Load(Activator.CreateInstance<T>(), saveDefaultConfigIfNotExist);
```
PrimeSonic: Add this to the constraints
```CS
where T : ModConfig, new()
```
This enforces the requirement that the class type `T` specified must have a parameterless constructor.

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,226 @@
...
+    ///         string activationKey = KeyCodeUtils.KeyCodeToString(Config.ActivationKey);
+    ///         Console.WriteLine($"[MyMod] LOADED: ActivationKey = {activationKey}");
+    ///         // Will print "[MyMod] LOADED: ActivationKey = Backspace" in Player.log on first run,
+    ///         // and whatever value they have saved in their config.json for the "ActivationKey"
+    ///         // property on subsequent runs.
+    ///     }
+    /// }
+    /// </code>
+    /// </example>
+    public class ModConfig
```
PrimeSonic: Should probably be `abstract`

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,226 @@
...
+            string.IsNullOrEmpty(subfolder) ? string.Empty : subfolder,
+            $"{fileName}.json"
+        );
+
+        /// <summary>
+        /// Creates a new instance of <see cref="ModConfig"/>
+        /// </summary>
+        /// <param name="fileName">The name of the config file, "config" by default.</param>
+        /// <param name="subfolder">Optional subfolder for the config file.</param>
+        public ModConfig(string fileName = "config", string subfolder = null)
```
PrimeSonic: This constructor should be `protected`

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,226 @@
...
+        ///         // property on subsequent runs.
+        ///         Config.ActivationKey = KeyCode.Mouse2;
+        ///         Config.Save();
+        ///         // The "ActivationKey" property in config.json will now read "MouseButtonMiddle"
+        ///     }
+        /// }
+        /// </code>
+        /// </example>
+        /// <seealso cref="Load{T}(T, bool)"/>
+        public void Save()
```
PrimeSonic: How come `Load` is `static` but `Save` is not?

Perhaps it might be worth considering if we even need these methods to live as part of this `ModConfig` class at all.  
It may be better to build out a ModConfig service, with a well defined interface, instead of packing all of the functionality into the class itself.

Open for discussion.

SMLHelper/Options/ModConfig.cs
tobeyStraitjacket: Good catch.

SMLHelper/Options/ModConfig.cs
tobeyStraitjacket: Thanks, I didn't know about this.

SMLHelper/Options/ModConfig.cs
tobeyStraitjacket: I'm honestly happy to have it just load the defaults, have been wanting input on this.

PrimeSonic: Using the default config whenever there is a serialization exception is what I (and I'm sure plenty others) have been doing for a while. :+1:

SMLHelper/Options/ModConfig.cs
tobeyStraitjacket: Fair enough. My interest has been in reducing boilerplate. Being able to simply:

`MyConfig` config = ModConfig.Load<MyConfig>();`
to load a file and then simply:
`config.Save();`
To save it once changed, say from the options menu.

I was interested in writing an instance version of ModConfig.Load() that handles loading the values in, the main issue is that it isn't super clean - as far as I can tell, we'd be deserializing json into a new object, then copying the properties back into the original object, probably using reflection. If there's a better way I'm open to it.

PrimeSonic: Instead of making the `Load` method instance, the `Save` method could be made static easily.
```CS
MyConfig config = ModConfig.Load<MyConfig>();
ModConfig.Save(config);
```
And if these are `static` methods, they could be easily moved into a service model, which would allow for the service interface to be easily replaced whenever needed.
We can still include static versions of those methods (like we do with the various Handlers).
But honestly, that might be overkill right now.
I think the simplest solution at this time would be simply making `Save` a static method as well.

Though do consider if these static methods shouldn't be part of some separate utility class instead of defined within the same `ModConfig` base class.

SMLHelper/Options/ModConfig.cs
```diff
@@ -0,0 +1,186 @@
...
+    /// A framework for handling JSON config files for your mods.
+    /// </summary>
+    /// <example>
+    /// <code>
+    /// using System;
+    /// using SMLHelper.V2.Options;
+    /// using SMLHelper.V2.Utility;
+    /// using UnityEngine;
+    /// 
+    /// public class Config : ModConfig {
```
zorgesho: Funny thing, I came up with exactly same naming in my code (base class - ModConfig, derived - just Config), but later I switched to the opposite and made Config base class. You use derived name just in couple of places when you declare and load config and base class name is used everywhere else.

tobeyStraitjacket: I was mostly just sticking with the same convention as ModOptions, but this is all likely to be changed soon :)

SMLHelper/Options/ModConfig.cs
tobeyStraitjacket: Have made changes which should resolve your issues :)

SMLHelper/Interfaces/IJsonFile.cs
```diff
@@ -0,0 +1,19 @@
...
+    /// <seealso cref="Options.ConfigFile"/>
+    /// <seealso cref="Utility.JsonUtils.Load{T}(T, bool, Oculus.Newtonsoft.Json.JsonConverter[])"/>
+    /// <seealso cref="Utility.JsonUtils.Save{T}(T, Oculus.Newtonsoft.Json.JsonConverter[])"/>
+    public interface IJsonFile
+    {
+        /// <summary>
+        /// The file path at which the JSON file is accessible for reading and writing.
+        /// </summary>
+        [JsonIgnore]
+        string JsonFilePath { get; }
```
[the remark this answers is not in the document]
tobeyStraitjacket: From my testing, they carry over. None of my tests have had properties with the `[JsonIgnore]` attribute declared _only_ in an `Interface` saved to disk.

SMLHelper/Options/ConfigFile.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            var path = Path.GetDirectoryName(Assembly.GetCallingAssembly().Location);
+            if (string.IsNullOrEmpty(fileName))
+            {
+                fileName = "config";
+            }
+            JsonFilePath = Path.Combine(
+                path,
+                string.IsNullOrEmpty(subfolder) ? string.Empty : subfolder,
+                $"{fileName}.json"
```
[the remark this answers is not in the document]
tobeyStraitjacket: Good catch on the whitespace, I'll definitely add this to the docs.

SMLHelper/Interfaces/IJsonFile.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Or, to put that in hopefully an easier to understand way:

When testing, and saving files by using an instance of the `ConfigFile` class, which implements `IJsonFile`, the `JsonFilePath` property has not been saved to disk. As such, I did not bother including the `[JsonIgnore]` attribute in the `ConfigFile` class, and don't believe it's necessary.

SMLHelper/Options/ConfigFile.cs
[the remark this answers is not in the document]
tobeyStraitjacket: I've looked at the documentation for `Path.Combine` which you linked about for some extra information, and found this:

> The parameters are not parsed if they have white space. Therefore, if path2 includes white space (for example, " \file.txt "), the Combine method appends path2 to path1.

So it appears that you _can_ include white space, they merely won't be _parsed_ (to remove extra slashes etc.), but they'll still be _combined_.

I went on to test this by creating a class derived from `ConfigFile`:
```cs
    internal class Config : ConfigFile
    {
        public KeyCode OverrideKey { get; set; } = KeyCode.Escape;
        public AutoLoadMode AutoLoadMode { get; set; } = AutoLoadMode.MostRecentlySaved;
        public KeyCode ToggleAutoLoadMode { get; set; } = KeyCode.LeftControl;

        public Config() : base(subfolder: " Config Files ") { }
    }
```
And it created a file "config.json" at the path "QMods\MyMod\ Config Files\config.json". The trailing white space was omitted - which I assume was enforced by Windows (Explorer does the same thing) - but otherwise it worked as you might expect. As such, it would be misleading to suggest it cannot contain white space.

SMLHelper/Interfaces/IJsonFile.cs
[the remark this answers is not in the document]
tobeyStraitjacket: image

See above example of a file saved from a class implementing the `IJsonFile` interface _without_ specifying the `[JsonIgnore]` attribute in the derived class for properties where it is specified in `IJsonFile`.

```cs
public abstract class ConfigFile : IJsonFile
    {
        /// <summary>
        /// The file path at which the JSON file is accessible for reading and writing.
        /// </summary>
        public string JsonFilePath { get; private set; }
...
```

SMLHelper/Json/ConfigFile.cs
```diff
@@ -0,0 +1,90 @@
...
+    using SMLHelper.V2.Json.Converters;
+    using SMLHelper.V2.Json.ExtensionMethods;
+    using SMLHelper.V2.Json.Interfaces;
+    using Oculus.Newtonsoft.Json;
+    using Oculus.Newtonsoft.Json.Converters;
+
+    /// <summary>
+    /// A simple implementation of <see cref="IJsonFile"/> for use with config files.
+    /// </summary>
+    public abstract class ConfigFile : IJsonFile
```
AlexejheroYTB: Should this class really be abstract?

PrimeSonic: That's the entire point of it.  
It's there to be a base class to the class that contains all the properties that will _actually_ get serialized.  
Just one way to use this new feature.

AlexejheroYTB: Then perhaps it should not be titled Config**File**

AlexejheroYTB:
```cs
public class MyConfig : ConfigFile
```
`MyConfig` is a configuration in this case, not a configuration **file**

PrimeSonic: The entire point of this class is to facilitate serialization of this class to & from a json _file_.

SMLHelper/Json/ConfigFile.cs
```diff
@@ -0,0 +1,90 @@
...
+            new FloatConverter(),
+            new StringEnumConverter(),
+            new VersionConverter()
+        };
+
+        /// <summary>
+        /// The <see cref="JsonConverter"/>s that should always be used when reading/writing JSON data.
+        /// </summary>
+        /// <seealso cref="alwaysIncludedJsonConverters"/>
+        public JsonConverter[] AlwaysIncludedJsonConverters => alwaysIncludedJsonConverters;
```
zorgesho: Also, that can be marked as virtual, so in case user needs to add some custom converters they can be added in the overrided property.

SMLHelper/Json/ConfigFile.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// 
+        /// public static class MyConfig : ConfigFile
+        /// {
+        ///     public KeyCode ActivationKey { get; set; } = KeyCode.Escape;
+        ///     public MyConfig() : base("options", "Config Files") { }
+        ///     // The config file will be stored at the path "QMods\YourModName\Config Files\options.json"
+        /// }
+        /// </code>
+        /// </example>
+        protected ConfigFile(string fileName = "config", string subfolder = null)
```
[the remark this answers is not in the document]
PrimeSonic: > you can add subolder's name to filename
That would require a little more knowledge of how the internal workings of this class than we should expect from modders using it.

> Is subfolder really needed here?
While I can also don't necessarily see myself using a subfolder for a config file,  
having an easy option to use it isn't a bad thing.
Better flexibility can drive adoption as more modders will be able to use this in the way they prefer.

SMLHelper/Json/ConfigFile.cs
PrimeSonic: The way I see this, there should actually be 2 properties.

One `internal` property that returns the `alwaysIncludedJsonConverters`,  
and a second `public` property that the modder can set or override with any additional JsonConverters they wish to use.

SMLHelper/Json/ConfigFile.cs
```diff
@@ -0,0 +1,90 @@
...
+
+        /// <summary>
+        /// Loads the JSON properties from the file on disk into the <see cref="ConfigFile"/>.
+        /// </summary>
+        /// <param name="createFileIfNotExist">Whether a new JSON file should be created with default values if it does not
+        /// already exist.</param>
+        /// <param name="jsonConverters">Optional <see cref="JsonConverter"/>s to be used for serialization.
+        /// The <see cref="AlwaysIncludedJsonConverters"/> will always be used, regardless of whether you pass them.</param>
+        /// <seealso cref="Save(JsonConverter[])"/>
+        public void Load(bool createFileIfNotExist = true, params JsonConverter[] jsonConverters)
```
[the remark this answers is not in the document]
PrimeSonic: > jsonConverters, ... can be added as an attribute to the property

This is true.  
As this is the instance variation of the `Load` method, specifically tailored to the `ConfigFile` class,  
then the JsonConverters can be handled through a property on the class.
They wouldn't need to be visible as parameters on this `Load` method.

SMLHelper/Json/ConfigFile.cs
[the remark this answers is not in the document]
zorgesho: > That would require a little more knowledge of how the internal workings of this class

That can be added to comments section
> having an easy option to use it isn't a bad thing

My point about all this is that adding optional parameter to public interface is easier than removal.

SMLHelper/Json/ConfigFile.cs
zorgesho: > modder can set or override with any additional JsonConverters they wish to use.
Or maybe modder wants to use their own custom VersionConverter? :)

tobeyStraitjacket: They can, quite easily. JsonConverters added directly as attributes on properties will take precedence over any other JsonConverters. And if they want to *entirely* exclude a JsonConverter that is in the always included, they can also make their own implementation of IJsonFile, or their own derived class of ConfigFile (yeah, I should make it possible to override the AlwaysIncludedJsonConverters property).

tobeyStraitjacket: The point of ConfigFile class is to have a basic, simple implementation of IJsonFile that covers the majority of use cases for working specifically with a commonly implemented config.json file. Currently, modders are reinventing the wheel every single time they want to do this (or they're reinventing the wheel once for themselves and using a Shared Project or something to use it again and again in any of their mods that use a config.json). What I am attempting to do here is make a simple, standardised API so that people don't *need* to keep reinventing the wheel, that covers the majority of use cases for someone who just wants to load and save properties to a config.json. An API that is simple and easy enough for someone who is making their very first mod to use.

If you want to go deeper with it, that's where you either write your own IJsonFile implementation, or make use of JsonUtils.Save() and JsonUtils.Load() directly. There's _nothing_ to stop people from not using the JSON API provided here at all and use Unity's own, or LitJson, or Newtonsoft.Json directly if they want. This is just to make it so you don't have to think about any of that in the majority of cases.

zorgesho: > JsonConverters added directly as attributes on properties will take precedence over any other JsonConverters

Yeah, I mean with that overrided property. You can add them with overrided property, you can add them with attributes and you can add them via method parameters. Maybe simplify that?

tobeyStraitjacket: I do agree it doesn't *need* to be visible as parameters on the Load() and Save() instance methods of ConfigFile or IJsonFile. They're mostly there for greater flexibility of usage - to allow for more use cases- but I could see myself conceding on removing them as parameters from ConfigFile.Load()/Save() instance methods. I absolutely want to keep them on the JsonUtils.Load()/Save() static methods, however.

zorgesho: > I absolutely want to keep them on the JsonUtils.Load()/Save() static methods, however.

Maybe just add JsonSerializerSettings instead? I think, it's possible to include converters there.

PrimeSonic:
Is there anything left before this can be merged?

PR #146, opened 2020-05-01
Mod Options Fixes for Below Zero

MrPurple6411:

  - Fixed Mod Options not finding Mods Tab
  - Fixed Null Key Get in ModOptionsHeadingsToggle at StatesConfig.get
  - Updated some dependencies.


SMLHelper/Patchers/OptionsPanelPatcher.cs
```diff
@@ -132,7 +142,17 @@ private class StatesConfig
 
                     public HeadingState this[string name]
                     {
-                        get => states.TryGetValue(name, out HeadingState state)? state: HeadingState.Expanded;
+                        get
+                        {
+                            if(string.IsNullOrEmpty(name))
```
zorgesho, line 147: Why this check is needed here?

SMLHelper/Options/SliderModOption.cs
```diff
@@ -6,6 +6,9 @@
     using UnityEngine;
     using UnityEngine.UI;
     using UnityEngine.Events;
+#if BELOWZERO
```
zorgesho, line 9: Can it be something like this?
```cs
#if SUBNAUTICA
    using Text = UnityEngine.UI.Text;
#endif
#if BELOWZERO
    using Text = TMPro.TextMeshProUGUI;
#endif
```

PrimeSonic: This should also work. Personally I don't have a preference for either style.

SMLHelper/Patchers/OptionsPanelPatcher.cs
MrPurple6411: somehow in BZ a null is getting passed into this getter.

SMLHelper/Options/SliderModOption.cs
MrPurple6411: yup this works as well. and as I have no issue either way I put that in.

zorgesho: Less duplicated code this way

SMLHelper/Patchers/OptionsPanelPatcher.cs
PrimeSonic: https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=netframework-4.0

An `ArgumentNullException` can be thrown if it's passed into `Dictionary.TryGetValue`.  
This seems like a sensible check to make.

MrPurple6411: exactly what was happening.

zorgesho: Yeah, probably `text` field in new BZ Text component can be null if it is empty.

@MrPurple6411 Let it be something like this then:
`states.TryGetValue(name ?? "", out HeadingState state)? state: HeadingState.Expanded;`

MrPurple6411: ahhhh ok thats the issue... Text is no longer used in BZ  its now TextMeshProUGUI   that would do it. meaning i need to go find your ref to Text that is getting passed to this.

zorgesho: It's there https://github.com/SubnauticaModding/SMLHelper/blob/f59e6ec623fd39fe3697a6e676306c0a4125d994/SMLHelper/Patchers/OptionsPanelPatcher.cs#L219

PR #150, opened 2020-05-03
More precise order

K07H:
  - Updated SetQuickSlotType comment.

### DIFF
--- a/SMLHelper/Handlers/CraftDataHandler.cs
+++ b/SMLHelper/Handlers/CraftDataHandler.cs
@@ -33,7 +33,7 @@
 
         /// <summary>
         /// <para>Allows you to edit QuickSlotType for TechTypes. Can be used for existing TechTypes too.</para>
-        /// <para>Careful: This has to be called after the prefab registration.</para>
+        /// <para>Careful: This has to be called after <see cref="SetTechData(TechType, Crafting.RecipeData)"/> and <see cref="SetTechData(TechType, JsonValue)"/>.</para>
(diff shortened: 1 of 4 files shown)


SMLHelper/Handlers/CraftDataHandler.cs
```diff
@@ -33,7 +33,7 @@ public static void SetEquipmentType(TechType techType, EquipmentType equipmentTy
 
         /// <summary>
         /// <para>Allows you to edit QuickSlotType for TechTypes. Can be used for existing TechTypes too.</para>
-        /// <para>Careful: This has to be called after the prefab registration.</para>
+        /// <para>Careful: This has to be called after <see cref="SetTechData(TechType, Crafting.RecipeData)"/> and <see cref="SetTechData(TechType, JsonValue)"/>.</para>
```
PrimeSonic, line 36: `JsonValue` is a Below Zero only type.  
Any line that would reference is needs to be between the preprocessor directives for Below Zero.
```CS
#if BELOWZERO
#endif
```

K07H: Sure Prime you can go ahead for the edition.
I asked MrPurple if his fix on 2.6.3 solved the fact that I had to call SetQuickSlotType after SetTechData but I got no answer yet.

K07H: Btw it's just a comment it's not breaking anything anyway.

PR #152, opened 2020-05-04
Enhancing abstracts further

MrPurple6411:

  - Loot Distribution in Spawnable
  - Custom Scanner entries in PDAITEN
  - Custom Encyclopedia entries in PDAITEM

### DIFF
--- a/SMLHelper/Assets/Spawnable.cs
+++ b/SMLHelper/Assets/Spawnable.cs
@@ -1,22 +1,26 @@
 ﻿namespace SMLHelper.V2.Assets
 {
-    using System;
-    using System.IO;
-    using System.Reflection;
     using Handlers;
     using SMLHelper.V2.Interfaces;
     using SMLHelper.V2.Utility;
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Reflection;
+    using UWE;
 
     /// <summary>
     /// An item that can be spawned into the game.
     /// </summary>
     /// <seealso cref="ModPrefab"/>
-    public abstract class Spawnable : ModPrefab
+    public abstract class Spawnable: ModPrefab
     {
         internal IPrefabHandler PrefabHandler { get; set; } = Handlers.PrefabHandler.Main;
         internal ISpriteHandler SpriteHandler { get; set; } = Handlers.SpriteHandler.Main;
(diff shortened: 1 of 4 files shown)


SMLHelper/Assets/Spawnable.cs
```diff
@@ -70,22 +86,33 @@ public abstract class Spawnable : ModPrefab
...
+                {
+                    CraftDataHandler.SetItemSize(TechType, SizeInInventory);
+                }
+            };
 
-                if (!this.SizeInInventory.Equals(defaultSize))
-                    this.CraftDataHandler.SetItemSize(this.TechType, this.SizeInInventory);
+            OnFinishedPatching += () =>
+            {
+                if(EntityInfo != null && BiomesToSpawnIn != null)
```
PrimeSonic, line 111: Is there any reason why this needs to be done in `OnFinishedPatching` and can't be done from `CorePatchEvents`

SMLHelper/Assets/Spawnable.cs
```diff
@@ -128,14 +157,14 @@ public void Patch()
...
-            this.IsPatched = true;
+            IsPatched = true;
 
             OnFinishedPatching?.Invoke();
         }
 
         internal virtual void PatchTechType()
         {
-            this.TechType = this.TechTypeHandler.AddTechType(ModName, this.ClassID, this.FriendlyName, this.Description, false);
+            TechType = TechTypeHandler.AddTechType(ModName, ClassID, FriendlyName, Description, false);
```
PrimeSonic, line 167: Be careful with Code Cleanup. Filling the diff with "non-changes" like makes it harder to spot the changes that actually matter.

SMLHelper/Assets/Spawnable.cs
MrPurple6411: Actually no. I don't think so.

PrimeSonic: Then it would be best to put it inside the `CorePatchEvents` since that is `internal` and managed by SMLHelper while `OnFinishedPatching` is `public` and handled by the individual mods.

MrPurple6411: Will do.

PR #153, opened 2020-05-17
Ability to register ping types via SML

brett-taylor:
Ability to register new ping types via SML helper

- PingHandler which has a `RegisterNewPingType` and `GetPingType` methods
- `RegisterNewType` is responsible for taking a name of a PingType and a Sprite and creating a PingType enum. It also inserts the Sprite into a collection which when `SpriteManager.Get` is called with `SpriteManager.Group.Pings` will check that collection.
- `GetPingType` is just a wrapper around `Enum.Parse` which handles if you give it an invalid name.

### DIFF
--- a/SMLHelper/Handlers/PingHandler.cs
+++ b/SMLHelper/Handlers/PingHandler.cs
@@ -1,9 +1,12 @@
-﻿using SMLHelper.V2.Patchers;
-using UnityEngine;
-
-namespace SMLHelper.V2.Handlers
+﻿namespace SMLHelper.V2.Handlers
 {
+    using Patchers;
     using Interfaces;
+#if SUBNAUTICA
+    using Sprite = Atlas.Sprite;
+#elif BELOWZERO
+    using Sprite = UnityEngine.Sprite;
+#endif
 
     /// <summary>
     /// A handler related to PingTypes
@@ -18,37 +21,57 @@ namespace SMLHelper.V2.Handlers
         private PingHandler()
         {
         }
-
+        
         /// <summary>
         /// Registers a ping type for use when creating a beacon
         /// </summary>
         /// <param name="pingName">The name of the new ping type</param>
         /// <param name="sprite">The sprite that is associated with the ping</param>
         /// <returns>The newly registered PingType</returns>
-        public PingType RegisterNewPingType(string pingName, Sprite sprite)
+        public static PingType RegisterNewPingType(string pingName, Sprite sprite)
         {
-            return RegisterNewPingType(pingName, new Atlas.Sprite(sprite));   
+            return Main.RegisterNewPingType(pingName, sprite);
         }
 
+        /// <summary>
+        /// Safely looks for a modded ping type in the SMLHelper PingTypeCache and outputs its <see cref="PingType"/> value when found.
+        /// </summary>
+        /// <param name="pingTypeString">The string used to define the modded PingType</param>
+        /// <param name="moddedPingType">The PingType enum value. Defaults to <see cref="PingType.None"/> when the PingType was not found.</param>
+        /// <returns><c>True</c> if the PingType was found; Otherwise <c>false</c></returns>
+        public static bool TryGetModdedPingType(string pingTypeString, out PingType moddedPingType)
+        {
+            return Main.TryGetModdedPingType(pingTypeString, out moddedPingType);
+        }
+        
         /// <summary>
         /// Registers a ping type for use when creating a beacon
         /// </summary>
         /// <param name="pingName">The name of the new ping type</param>
         /// <param name="sprite">The sprite that is associated with the ping</param>
         /// <returns>The newly registered PingType</returns>
-        public PingType RegisterNewPingType(string pingName, Atlas.Sprite sprite)
+        PingType IPingHandler.RegisterNewPingType(string pingName, Sprite sprite)
         {
-            return PingTypePatcher.AddPingType(pingName, sprite);
+            return PingTypePatcher.AddPingType(pingName, sprite);   
         }
-
+        
         /// <summary>
-        /// Returns nullable containing a ping type associated with a name, if it exists.
+        /// Safely looks for a modded ping type in the SMLHelper PingTypeCache and outputs its <see cref="PingType"/> value when found.
         /// </summary>
-        /// <param name="pingName">The name associated with the PingType you want.</param>
-        /// <returns>A nullable containing a PingType if one was associated with the given pingName</returns>
-        public PingType? GetPingType(string pingName)
+        /// <param name="pingTypeString">The string used to define the modded PingType</param>
+        /// <param name="moddedPingType">The PingType enum value. Defaults to <see cref="PingType.None"/> when the PingType was not found.</param>
+        /// <returns><c>True</c> if the PingType was found; Otherwise <c>false</c></returns>
+        bool IPingHandler.TryGetModdedPingType(string pingTypeString, out PingType moddedPingType)
         {
-            return PingTypePatcher.GetPingType(pingName);
+            var cache = PingTypePatcher.cacheManager.RequestCacheForTypeName(pingTypeString, false);
+            if (cache != null)
+            {
+                moddedPingType = (PingType) cache.Index;
+                return true;
+            }
+
+            moddedPingType = PingType.None;
+            return false;
         }
     }
 }
--- a/SMLHelper/Handlers/PingHandler.cs
+++ b/SMLHelper/Handlers/PingHandler.cs
@@ -1,0 +1,54 @@
+﻿using SMLHelper.V2.Patchers;
+using UnityEngine;
+
+namespace SMLHelper.V2.Handlers
+{
+    using Interfaces;
+
+    /// <summary>
+    /// A handler related to PingTypes
+    /// </summary>
+    public class PingHandler : IPingHandler
+    {
+        /// <summary>
+        /// Main entry point for all calls to this handler.
+        /// </summary>
+        public static IPingHandler Main { get; } = new PingHandler();
+
+        private PingHandler()
+        {
+        }
+
+        /// <summary>
+        /// Registers a ping type for use when creating a beacon
+        /// </summary>
+        /// <param name="pingName">The name of the new ping type</param>
+        /// <param name="sprite">The sprite that is associated with the ping</param>
+        /// <returns>The newly registered PingType</returns>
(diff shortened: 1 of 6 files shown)


brett-taylor:
One
Two

The following code for this would be similar to:

```c#
var sprite = SMLHelper.V2.Utility.ImageUtils.LoadSpriteFromFile("fileDir");
var pingType = SMLHelper.V2.Handlers.PingHandler.Main.RegisterNewPingType("TechFabricator", sprite);
Object.FindObjectsOfType<Fabricator>().ForEach(fab =>
{
    var pingInstance = fab.gameObject.AddComponent<PingInstance>();
    pingInstance.pingType = pingType;
    pingInstance.maxDist = 15;
    pingInstance.minDist = 20;
    pingInstance.origin = fab.transform;
    pingInstance.SetLabel("Hey Github PR!");

    PingManager.Unregister(pingInstance);
    PingManager.Register(pingInstance);
});

```

brett-taylor:
For below zero it works on Sprite rather than Atlas.Sprite, I'll add support for BZ tomorrow.

Also right now the PingType enum cache starts from 10. I should probably up that incase SN ever gets a new ping type.

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+using Harmony;
+using SMLHelper.V2.Handlers;
+using SMLHelper.V2.Utility;
+
+namespace SMLHelper.V2.Patchers
+{
+    internal class PingTypePatcher
+    {
+        private const string TechTypeEnumName = "PingType";
+        internal static readonly int startingIndex = 10;
```
PrimeSonic, line 13: Should start over the max value of `PingType` which we can use as a constant.
```CS
internal const int startingIndex = (int)PingType.Sunbeam + 1;
```

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+            Logger.Log($"Successfully added PingType: '{name}' to Index: '{cache.Index}'", LogLevel.Debug);
+            return pingType;
+        }
+
+        internal static PingType? GetPingType(string pingName)
+        {
+            try
+            {
+                var ping = Enum.Parse(typeof(PingType), pingName);
+                return (PingType) ping;
```
PrimeSonic, line 65: Can we use use [Enum.TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.enum.tryparse?view=netframework-4.0) instead of a try-catch block?  
Also use the case-insensitive version.

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+            }
+            catch (ArgumentException e)
+            {
+                return new PingType?();
```
PrimeSonic, line 69: Return `null` if the parsing failed.

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                var pingType = GetPingType(name);
+                if (pingType.HasValue && cachedSprites.ContainsKey(pingType.Value))
+                {
+                    __result = cachedSprites[pingType.Value];
+                    return false;
+                }
+            }
+
+            return true;
```
PrimeSonic, line 108: We shouldn't prefix access to the SpriteManager.  
Instead, do how we have in the [SpritePatcher](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Patchers/SpritePatcher.cs) and simply inject these sprites into the correct locations.

This may mean that the `cachedSprites` field goes away or is used only for working with the SpritePatcher.

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+                Index = cacheManager.GetNextAvailableIndex()
+            };
+
+            if (cacheManager.IsIndexAvailable(cache.Index))
+                cache.Index = cacheManager.GetNextAvailableIndex();
+
+            var pingType = (PingType) cache.Index;
+            cacheManager.Add(pingType, cache.Index, cache.Name);
+            
+            cachedSprites.Add(pingType, sprite);
```
PrimeSonic, line 49: This should probably interact with the [`SpritePatcher`](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Patchers/SpritePatcher.cs) instead of handling the sprites here.

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
...
+            {
+                __result = __result.Cast<PingType>().Concat(cacheManager.ModdedKeys).ToArray();
+            }
+        }
+        
+        private static bool Prefix_IsDefined(Type enumType, object value, ref bool __result)
+        {
+            var result = enumType == typeof(PingType) && cacheManager.ContainsKey((PingType) value);
+            __result |= result;
+            return !result;
```
PrimeSonic, line 123: I can see what you're doing here, but it's less readable and harder to follow when "result" is used in multiple ways like this.  
Can we go with a simpler style?
```CS
if (enumType == typeof(PingType) && cacheManager.ContainsKey((PingType) value))
{
    __result = true;
    return false;
}

return true;
```

SMLHelper/Patchers/PingTypePatcher.cs
```diff
@@ -0,0 +1,151 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
```
PrimeSonic, line 3: Just a personal preference: but get rid of this if you can.
~~using System.Linq;~~

SMLHelper/Handlers/PingHandler.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// <summary>
+    /// A handler related to PingTypes
+    /// </summary>
+    public class PingHandler : IPingHandler
+    {
+        /// <summary>
+        /// Main entry point for all calls to this handler.
+        /// </summary>
+        public static IPingHandler Main { get; } = new PingHandler();
+
```
PrimeSonic, line 17: @AlexejheroYTB @MrPurple6411  
Are we okay with not having static methods on this new handler?

SMLHelper/Interfaces/IPingHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// </summary>
+    public interface IPingHandler
+    {
+        /// <summary>
+        /// Registers a ping type for use when creating a beacon
+        /// </summary>
+        /// <param name="pingName">The name of the new ping type</param>
+        /// <param name="sprite">The sprite that is associated with the ping</param>
+        /// <returns>The newly registered PingType</returns>
+        PingType RegisterNewPingType(string pingName, UnityEngine.Sprite sprite);
```
PrimeSonic, line 16: If all `UnityEngine.Sprite`s are simply going to be converted into `Atlas.Sprite`s, then why bother offering it in the first place?

Could we instead just use `Atlas.Sprite` for Subnautica and `UnityEngine.Sprite` for Below Zero?

SMLHelper/Interfaces/IPingHandler.cs
```diff
@@ -0,0 +1,33 @@
...
+        /// <param name="pingName">The name of the new ping type</param>
+        /// <param name="sprite">The sprite that is associated with the ping</param>
+        /// <returns>The newly registered PingType</returns>
+        PingType RegisterNewPingType(string pingName, Atlas.Sprite sprite);
+
+        /// <summary>
+        /// Returns nullable containing a ping type associated with a name, if it exists.
+        /// </summary>
+        /// <param name="pingName">The name associated with the PingType you want.</param>
+        /// <returns>A nullable containing a PingType if one was associated with the given pingName</returns>
```
PrimeSonic, line 30: Documentation should probably read more like this.
```CS
/// <returns>The PingType value associated with the given pingName if it exists; otherwise returns null.</returns>

File: SMLHelper/Handlers/PingHandler.cs [hunk 11]
```diff
@@ -0,0 +1,54 @@
+﻿using SMLHelper.V2.Patchers;
+using UnityEngine;
+
+namespace SMLHelper.V2.Handlers
+{
+    using Interfaces;
```
PrimeSonic, line 6: Keep `using` statements together.

SMLHelper/Patchers/PingTypePatcher.cs
brett-taylor: PingType doesn't seem to rely on consecutive numbers so we could start from e.g. 1000 to give some breathing room to below zero incase they ever add a new ping type

SMLHelper/Patchers/PingTypePatcher.cs
brett-taylor: We're calling at this point a patched version of the Parse method so ignoreCase wouldnt do anything. Im not using TryParse here as TryParse isn't patched only Parsed. Have you got an example of harmony patching TryParse?

PrimeSonic: I see what you're doing here now. I was confused because there isn't an analogous method to this in either `TechTypePatcher` or `CraftTreeTypePatcher`.

The problem here is that this implementation relies on a Harmony patch to work, but any mod that requests this info will be using it _before_ SMLHelper has made its Harmony patches.

If you want to expose functionality that lets external consumers get the `PingType` enum value for a given ping name string, use something like the `TryGetModdedTechType` and `ModdedTechTypeExists` methods of the [TechTypeHandler](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Handlers/TechTypeHandler.cs), which make use of the enum cache.  
If you really want to be able to handle vanilla values as well, then you can use a switch statement like [ValidateStandardCraftTree](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Handlers/CraftTreeHandler.cs#L367) to go through all the known types before checking the cache on the `default` case.

SMLHelper/Patchers/PingTypePatcher.cs
PrimeSonic: That works too.

brett-taylor:
@PrimeSonic PR is updated with your changes. 

Also have tested on BZ now and it works.
image

PrimeSonic:
Merging now. Will address remaining comments on the `dev` branch.

PR #155, opened 2020-06-10
Ability to add buttons inside in-game menu.

K07H:

  -  Added possibility to add buttons inside in-game menu.


K07H:
> 
> 
> Looks like a cool feature. 👍

If you want to look at what I need to do when out of SML Helper context lol:
https://github.com/K07H/DecorationsMod/blob/master/DecorationsMod/ConfigOptions.cs#L96
Basically creating a checkbox to then replace it by a button with reflection, so it's not optimized and not so easy to read for someone not used to reflection :p

PR #156, opened 2020-06-11
Added volume control to PlaySound

gurrenm3:

  - Ability to change volume level of custom played sounds

### DIFF
--- a/SMLHelper/Utility/AudioUtils.cs
+++ b/SMLHelper/Utility/AudioUtils.cs
@@ -32,6 +32,18 @@
         }
 
         /// <summary>
+        /// Plays a <see cref="Sound"/> globally from a path. Must be a .wav file. Has overload for controlling volume
+        /// </summary>
+        /// <param name="path">The path of the sound. Relative to the base game folder.</param>
+        /// <param name="mode"></param>
+        /// <param name="volumeControl">Which volume control to adjust sound levels by. How loud sound is.</param>
+        /// <returns>The channel on which the sound was created</returns>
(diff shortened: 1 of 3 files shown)


gurrenm3:
Added volumeControl overload to PlaySound method that uses a path instead of Sound

SMLHelper/Utility/AudioUtils.cs
```diff
@@ -42,5 +54,55 @@ public static Channel PlaySound(Sound sound)
             RuntimeManager.LowlevelSystem.playSound(sound, channels, false, out Channel channel);
             return channel;
         }
+
+        /// <summary>
+        /// The a list the different volume controls in the game
+        /// </summary>
+        public enum VolumeControl
```
AlexejheroYTB, line 61: Move the `VolumeControl` enum out of the `AudioUtils` class.
Also, I think the name `VolumeChannel` would fit it a bit better than the current one.

PrimeSonic: The newly renamed `SoundChannel` enum is still inside the `AudioUtils` class.  
Put it into its own file in the same folder and namespace.

AlexejheroYTB: I know, but it doesn't really make sense to put it outside of the AudioUtils class, since it's only used in there.

PrimeSonic: It's a lot cleaner to use `SoundChannel.Master` instead of `AudioUtils.SoundChannel.Master` every time.

PR #157, opened 2020-06-12
Consolidated Enum Harmony patching

PrimeSonic:
- Combines the 3 separate sets of Harmony patches to `Enum` into a single set.
- Converted to PatchUtils

### DIFF
--- a/SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
+++ b/SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
@@ -27,6 +27,8 @@ namespace SMLHelper.V2.Patchers
         private static void PatchForSubnautica(HarmonyInstance harmony)
         {
             PatchUtils.PatchClass(harmony);
+
+            Logger.Log("CraftDataPatcher is ready.", LogLevel.Debug);
         }
 
         [PatchUtils.Prefix]
--- a/SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
+++ b/SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
@@ -27,8 +27,6 @@ namespace SMLHelper.V2.Patchers
         private static void PatchForSubnautica(HarmonyInstance harmony)
         {
             PatchUtils.PatchClass(harmony);
-
-            Logger.Log("CraftDataPatcher is ready.", LogLevel.Debug);
         }
 
         [PatchUtils.Prefix]
--- a/SMLHelper/Patchers/EnumPatcher.cs
+++ b/SMLHelper/Patchers/EnumPatcher.cs
@@ -1,0 +1,113 @@
+﻿namespace SMLHelper.V2.Patchers
+{
+    using System;
+    using System.Collections.Generic;
+    using Harmony;
+    using SMLHelper.V2.Utility;
+
+    internal class EnumPatcher
+    {
+        internal static void Patch(HarmonyInstance harmony)
+        {
+            PatchUtils.PatchClass(harmony);
+
+            Logger.Log("EnumPatcher is done.", LogLevel.Debug);
+        }
+
+        [PatchUtils.Postfix]
+        [HarmonyPatch(typeof(Enum), nameof(Enum.GetValues))]
+        private static void Postfix_GetValues(Type enumType, ref Array __result)
+        {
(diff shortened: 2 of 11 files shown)


SMLHelper/Patchers/EnumPatcher.cs
```diff
@@ -0,0 +1,113 @@
...
+            return list.ToArray();
+        }
+
+        [PatchUtils.Prefix]
+        [HarmonyPatch(typeof(Enum), nameof(Enum.IsDefined))]
+        private static bool Prefix_IsDefined(Type enumType, object value, ref bool __result)
+        {
+            if (IsDefined(TechTypePatcher.cacheManager, enumType, value) ||
+                IsDefined(CraftTreeTypePatcher.cacheManager, enumType, value) ||
+                IsDefined(PingTypePatcher.cacheManager, enumType, value))
```
zorgesho, line 53: It's not about `PatchUtils`, but maybe put these three managers into some dictionary by enum type? They can even add themselves to `EnumPatcher` in `EnumCacheManager` constructor, for example, so there will be no concrete enum types in `EnumPatcher`.

PrimeSonic: I thought about that. I decided to keep it simple for this first iteration and return to it in the event that we ever add a 4th enum to be patched.

zorgesho: Well, I think even with two enum types, more general approach would be useful in this case. But anyway, one `EnumPatcher` is better than having three similar patchers :)

PrimeSonic: While I do agree that a general approach would be the most helpful, it would require abstracting the 3 original enum patching classes in a way they could be used without an explicit reference to the specific enum type they handle.  
The generics involved make this harder than it looks since each of 3 `EnumCacheManager<T>` references are actually completely different types.

In short, it would be a lot of over engineering for not only little gain in terms of code readability but also would add additional overhead to these enum patch methods.

zorgesho: Ah, for some reason I thought that something like `List<EnumCacheManager<Enum>>` is possible. Yeah, you right, it's not worth it then.

SMLHelper/Patchers/CraftDataPatcher_Subnautica.cs
```diff
@@ -27,6 +27,8 @@ internal static void AddToCustomTechData(TechType techType, ITechData techData)
         private static void PatchForSubnautica(HarmonyInstance harmony)
         {
             PatchUtils.PatchClass(harmony);
+
+            Logger.Log("CraftDataPatcher is ready.", LogLevel.Debug);
```
AlexejheroYTB, line 31: This is already called in [`CraftDataPatcher.cs#L88`](https://github.com/SubnauticaModding/SMLHelper/blob/EnumPatchingCleanup/SMLHelper/Patchers/CraftDataPatcher.cs#L88)


```suggestion
```

SMLHelper/Patchers/EnumPatcher.cs
```diff
@@ -0,0 +1,113 @@
...
+                __result = GetValues(TechTypePatcher.cacheManager, __result);
+            }
+            else if (enumType == typeof(CraftTree.Type))
+            {
+                __result = GetValues(CraftTreeTypePatcher.cacheManager, __result);
+            }
+            else if (enumType == typeof(PingType))
+            {
+                __result = GetValues(PingTypePatcher.cacheManager, __result);
+            }
```
AlexejheroYTB, line 32: `EnumPatcher` should have an `internal` method which is called by other patchers that use enums. These values should not be hardcoded here IMO.

AlexejheroYTB: Nevermind, I saw your conversation with zorgesho about this.

SMLHelper/Patchers/TechTypePatcher.cs
```diff
@@ -90,90 +90,13 @@ private static List<int> PreRegisteredTechTypes()
             return bannedIndices;
         }
```
AlexejheroYTB, line 92: Maybe move these enum patchers into their own folder? ...or not?

PR #169, opened 2020-07-28
Add Config File to Example Mod

tobeyStraitjacket:

Updated ExampleMod to take advantage of ConfigFile utility.

### DIFF
--- a/Example mod/Mod.cs
+++ b/Example mod/Mod.cs
@@ -9,7 +9,7 @@ namespace SMLHelper.V2.Examples
 {
     public static class ExampleMod
     {
-        public static Config Config = new Config();
+        public static Config Config { get; } = new Config();
 
         public static void Patch()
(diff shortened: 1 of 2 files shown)


Example mod/Mod.cs
```diff
@@ -1,35 +1,29 @@
...
 using SMLHelper.V2.Options;
-using SMLHelper.V2.Utility;
 using UnityEngine;
 using UnityEngine.UI;
 
 namespace SMLHelper.V2.Examples
 {
     public static class ExampleMod
     {
+        public static Config Config = new Config();
```
PrimeSonic, line 12: Should be a `readonly` field or a `get` only property.

PrimeSonic: That is assuming this is supposed to follow the Singleton pattern.

PR #173, opened 2020-07-31
Attribute-based config and mod options

tobeyStraitjacket:
- **0 breaking changes!**

- Added various Attributes for automatically generating a `ModOptions` menu from the `public` fields/properties/methods of a `ConfigFile` subclass:

  - [`MenuAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/MenuAttribute.cs): Used to specify various options regarding the generated menu, including the display name, whether unattributed members should be included when generating options (see also [`IgnoreMemberAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/IgnoreMemberAttribute.cs)), when to automatically save the config file to disk, etc.

  - [`ButtonAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/ButtonAttribute.cs): Used to specify the attributed method should generate a `ModButtonOption` which calls the attributed method on click.

  - [`ChoiceAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/ChoiceAttribute.cs): Used to specify the attributed field/property should generate a `ModChoiceOption`. Works with index-based, string-based, or enum-based `ModChoiceOption`s.

  - [`KeybindAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/KeybindAttribute.cs): Used to specify the attributed field/property should generate a `ModKeybindOption`.

  - [`SliderAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/SliderAttribute.cs): Used to specify the attributed field/property should generate a `ModSliderOption`, as well as set its range, default value, etc.

  - [`ToggleAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/ToggleAttribute.cs): Used to specify the attributed field/property should generate a `ModToggleOption`.

  - [`OnChangeAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/OnChangeAttribute.cs): Used to specify a method that should be called when the value of the `ModOption` changes - method is specified by name and must be a member of the same class.

  - [`OnGameObjectCreatedAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/OnGameObjectCreatedAttribute.cs): Used to specify a method that should be called when the `ModOption`'s `GameObject` is created - method is specifed by name and must be a member of the same class.

  - [`IgnoreMemberAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/IgnoreMemberAttribute.cs): Used to signify the specified member should be ignored when generating the options menu. See also: [`MenuAttribute.MemberProcessing`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/MenuAttribute.cs#L182-L187)

  - [`ConfigFileAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Json/ConfigFileAttribute.cs): Used to optionally specify the filename and subfolder of the `ConfigFile` without needing to explicitly create an empty constructor which calls the `base(string fileName, string subfolder)` constructor, making for less tedious boilerplate

 - All of the `ModOption` related `EventArgs` now implement the new [`IModOptionEventArgs`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Interfaces/IModOptionEventArgs.cs) interface, this allows us to have both generic events for any kind of `OnChange`, as well as easy coercion from the interface back to the original EventArgs via pattern matching.

- As requested by @MrPurple6411, `ModSliderOption` now has a `Step` property which allows us to set the step (ie. round to nearest X) via either the standard [`AddSliderOption`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/SliderModOption.cs#L78-L93) or the new [`SliderAttribute`](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/SMLHelper/Options/Attributes/SliderAttribute.cs). Only fires OnChange events for values rounded to the step!

- [Example mod](https://github.com/SubnauticaModding/SMLHelper/blob/AttributeBasedConfigAndModOptions/Example%20mod/Mod.cs) has now been updated to make use of all the *shiny new things*.
  <details><summary>Example mod code:</summary>

  ```cs
  [QModCore]
  public static class ExampleMod
  {
      internal static Config Config { get; } = OptionsPanelHandler.RegisterModOptions<Config>();

      [QModPatch]
      public static void Patch()
      {
          Logger.Log(Logger.Level.Info, "Patched successfully!");
      }
  }

  public enum CustomChoice { One, Two, Three }

  [Menu("SMLHelper Example Mod")]
  public class Config : ConfigFile
  {
      [Choice("My index-based choice", "One", "Two", "Three"), OnChange(nameof(MyGenericValueChangedEvent))]
      public int ChoiceIndex;

      [Choice("My string-based choice", "Foo", "Bar"), OnChange(nameof(MyGenericValueChangedEvent))]
      public string ChoiceValue = "Foo";

      [Choice("My enum-based choice"), OnChange(nameof(MyGenericValueChangedEvent))]
      public CustomChoice ChoiceEnum;

      [Choice("My customised enum-based choice", "1", "2", "3"), OnChange(nameof(MyGenericValueChangedEvent))]
      public CustomChoice ChoiceCustomEnum;

      [Keybind("My keybind"), OnChange(nameof(MyGenericValueChangedEvent))]
      public KeyCode KeybindKey;

      [Slider("My slider", 0, 50, DefaultValue = 25), OnChange(nameof(MyGenericValueChangedEvent))]
      public int SliderValue;

      [Slider("My stepped slider", 0, 100, Step = 10), OnChange(nameof(MyGenericValueChangedEvent))]
      public int SteppedSliderValue;

      [Slider("My float-based slider", 0, 100, Format = "{0:F2}"), OnChange(nameof(MyGenericValueChangedEvent))]
      public float FloatSliderValue;

      [Toggle("My checkbox"), OnChange(nameof(MyCheckboxToggleEvent)), OnChange(nameof(MyGenericValueChangedEvent))]
      public bool ToggleValue;

      [Button("My button"), OnGameObjectCreated(nameof(MyGameObjectCreatedEvent))]
      public void MyButtonClickEvent(ButtonClickedEventArgs e)
      {
          Logger.Log(Logger.Level.Info, "Button was clicked!");
          Logger.Log(Logger.Level.Info, $"{e.Id}");
      }

      private void MyCheckboxToggleEvent(ToggleChangedEventArgs e)
      {
          Logger.Log(Logger.Level.Info, "Checkbox value was changed!");
          Logger.Log(Logger.Level.Info, $"{e.Value}");
      }

      private void MyGenericValueChangedEvent(IModOptionEventArgs e)
      {
          Logger.Log(Logger.Level.Info, "Generic value changed!");
          Logger.Log(Logger.Level.Info, $"{e.Id}: {e.GetType()}");

          switch (e)
          {
              case KeybindChangedEventArgs keybindChangedEventArgs:
                  Logger.Log(Logger.Level.Info, keybindChangedEventArgs.KeyName);
                  break;
              case ChoiceChangedEventArgs choiceChangedEventArgs:
                  Logger.Log(Logger.Level.Info, $"{choiceChangedEventArgs.Index}: {choiceChangedEventArgs.Value}");
                  break;
              case SliderChangedEventArgs sliderChangedEventArgs:
                  Logger.Log(Logger.Level.Info, sliderChangedEventArgs.Value.ToString());
                  break;
              case ToggleChangedEventArgs toggleChangedEventArgs:
                  Logger.Log(Logger.Level.Info, toggleChangedEventArgs.Value.ToString());
                  break;
          }
      }

      private void MyGameObjectCreatedEvent(GameObjectCreatedEventArgs e)
      {
          Logger.Log(Logger.Level.Info, "GameObject was created");
          Logger.Log(Logger.Level.Info, $"{e.Id}: {e.GameObject}");
      }
  }
  ```
  </details>


SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,680 @@
...
+    using Text = TMPro.TextMeshProUGUI;
+#endif
+
+    /// <summary>
+    /// An internal derivative of <see cref="ModOptions"/> for use in auto-generating a menu based on attributes
+    /// declared in a <see cref="ConfigFile"/>.
+    /// </summary>
+    /// <typeparam name="TConfigFile">The type of the class derived from <see cref="ConfigFile"/> to use for
+    /// loading to/saving from disk.</typeparam>
+    internal class ConfigModOptions<TConfigFile> : ModOptions where TConfigFile : ConfigFile, new()
```
AlexejheroYTB, line 26: This file is pretty big. Consider using `#region`

tobeyStraitjacket: I've added a couple of regions for grouping methods used for similar things together. What do you think?

SMLHelper/Options/ChoiceAttribute.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <summary>
+        /// Attribute used to signify the specified member should be represented in the mod's options menu as a
+        /// <see cref="ModChoiceOption"/>. Works for either <see cref="int"/> index-based, <see cref="string"/>-based, or
+        /// <see cref="Enum"/>-based members.
+        /// </summary>
+        /// <remarks>
+        /// <see cref="Enum"/> choices can also be parsed from their values by merely omitting the <see cref="ChoiceAttribute"/>.
+        /// </remarks>
+        /// <param name="options">The list of options for the user to choose from.</param>
+        public ChoiceAttribute(params string[] options) => Options = options;
```
PrimeSonic, line 53: Style preference: don't use lambs notation for constructors.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,685 @@
...
+            // Instantiate and load the config
+            Config = new TConfigFile();
+            Config.Load();
+
+            menuAttribute = typeof(TConfigFile).GetCustomAttributes(typeof(MenuAttribute), true).SingleOrDefault() as MenuAttribute
+                ?? new MenuAttribute(Name);
+
+            // Process members
+            var bindingFlags = BindingFlags.Instance | BindingFlags.Public;
+            foreach (PropertyInfo property in typeof(TConfigFile).GetProperties(bindingFlags)
```
PrimeSonic, line 50: If you're going to have a huge Linq statement to provide you the collection to iterate over, put that into a new variable outside so the main `foreach` condition doesn't span multiple lines.

tobeyStraitjacket: I see your point, I just like the semantic syntax of the declarative `foreach` in this case: it is very clear what is being looped over.

Moving that out into a new variable ahead of the `foreach` *might* allow that variable declaration to only span one line, but imo will sacrifice some readability in terms of being able to easily see "okay this is looping over the types where they are declared in a subclass of `ConfigFile` and where..."

And if your goal wasn't to make it so the variable declaration doesn't span only one line, just that the `foreach` condition doesn't... Isn't that making it harder to understand what's being looped over just for the sake of perceived tidyness?

PrimeSonic: How about we just add braces `{ }` around the one line statement being executed? (to more cleanly separate what is part of the `foreach` declaration and what is the action)?

tobeyStraitjacket: Sure, I'm on board with that. Tbh, I usually always use braces - only haven't bc I have noticed in many places in SML/QMM code they aren't being used and figured that was preferred style so I started ditching them in some cases. If I was mistaken, I will start adding them back.

PrimeSonic: I usually prefer no-braces on single-line execution statements, but if the control statement above it is going to span multiple lines, then I would stick with the braces there for clarity.

tobeyStraitjacket: Fair enough then :) ty for that clarification

tobeyStraitjacket:
Converting to draft as I want to give this a once-over re. performance optimisations with all the heavy-lifting/usage of LINQ & Reflection etc.

Example mod/Mod.cs
```diff
@@ -1,123 +1,95 @@
...
+        public KeyCode KeybindKey;
 
-        public override void BuildModOptions()
+        [Label("My slider"), Slider(0, 50, DefaultValue = 25, Format = "{0:F2}"), OnChange(nameof(MyGenericValueChangedEvent))]
+        public int SliderValue;
+
+        [Label("My stepped slider"), Slider(0, 100, Step = 10), OnChange(nameof(MyGenericValueChangedEvent))]
+        public int SteppedSliderValue;
+
+        [Label("My checkbox"), OnChange(nameof(MyCheckboxToggleEvent)), OnChange(nameof(MyGenericValueChangedEvent))]
```
PrimeSonic, line 51: For the sake of consistency, shouldn't there be a `Toggle` attribute here before the `OnChange` attribute?  
There is already an attribute for `Slider` and `Choice`.  
It likely isn't strictly necessary for it to function on a `bool` option but having a `Toggle` attribute would continue the pattern established above.

```CS
[Label("My checkbox"), Toggle, OnChange(nameof(MyCheckboxToggleEvent)), OnChange(nameof(MyGenericValueChangedEvent))]
```

Example mod/Mod.cs
```diff
@@ -1,123 +1,95 @@
...
-        {
-            if (e.Id != "exampleToggle") return;
-            ExampleMod.Config.ToggleValue = e.Value;
-            ExampleMod.Config.Save();
-        }
+        [Label("My customised enum-based choice"), Choice("1", "2", "3"), OnChange(nameof(MyGenericValueChangedEvent))]
+        public CustomChoice ChoiceEnumCustomValues;
+
+        [Label("My keybind"), OnChange(nameof(MyGenericValueChangedEvent))]
+        public KeyCode KeybindKey;
```
PrimeSonic, line 43: Also here, there could be an attribute for Keybinds.  
If it doesn't need any parameters to function, it can be as simple as this:
```CS
[Label("My keybind"), Keybind, OnChange(nameof(MyGenericValueChangedEvent))]
```

SMLHelper/Handlers/OptionsPanelHandler.cs
```diff
@@ -32,11 +35,45 @@ public static void RegisterModOptions(ModOptions options)
...
+        {
+            var configModOptions = new ConfigModOptions<T>();
+            RegisterModOptions(configModOptions);
+
+            var menuAttribute = typeof(T).GetCustomAttributes(typeof(MenuAttribute), true).SingleOrDefault() as MenuAttribute
+                ?? new MenuAttribute();
+
+            if (menuAttribute.SaveOn.HasFlag(MenuAttribute.SaveEvents.SaveGame))
+                IngameMenuHandler.RegisterOnSaveEvent(() => configModOptions.Config.Save());
+            if (menuAttribute.SaveOn.HasFlag(MenuAttribute.SaveEvents.QuitGame))
```
PrimeSonic, line 73: Keep an empty line before new if-statements, to more clearly separate them from the previous execution statement.

SMLHelper/Json/ConfigFile.cs
```diff
@@ -33,6 +39,26 @@ public abstract class ConfigFile : IJsonFile
         /// <seealso cref="alwaysIncludedJsonConverters"/>
         public JsonConverter[] AlwaysIncludedJsonConverters => alwaysIncludedJsonConverters;
 
+        /// <summary>
+        /// Creates a new instance of <see cref="ConfigFile"/>, parsing the filename and subfolder from a
+        /// <see cref="ConfigFileAttribute"/> if declared, or with default values otherwise.
+        /// </summary>
+        public ConfigFile()
+        {
+            if (GetType().GetCustomAttributes(typeof(ConfigFileAttribute), true).FirstOrDefault() is ConfigFileAttribute configFile)
```
PrimeSonic, line 48: Are we able to use the Generic form of `GetCustomAttributes` here? Or is that not available in this version of .NET?
```CS
GetType().GetCustomAttributes<ConfigFileAttribute>(true)...
```

SMLHelper/Options/ChoiceAttribute.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// <seealso cref="LabelAttribute"/>
+    /// <seealso cref="ModChoiceOption"/>
+    /// <seealso cref="Json.ConfigFile"/>
+    [AttributeUsage(AttributeTargets.Property | AttributeTargets.Field, AllowMultiple = false)]
+    public sealed class ChoiceAttribute : Attribute
+    {
+        /// <summary>
+        /// The list of options that will be displayed.
+        /// </summary>
+        public string[] Options { get; }
```
PrimeSonic, line 42: Remind me: Does this have to be an array? Or could it be an `IEnumerable<string>`?

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,700 @@
...
+            }
+            else if (memberType == typeof(KeyCode))
+            {   // Generate a ModKeybindOption
+                AddModOptionMetadata<ModKeybindOption>(memberInfo, memberType);
+            }
+            else if (memberType.IsEnum || memberInfo.GetCustomAttributes(typeof(ChoiceAttribute), true).Any())
+            {   // Generate a ModChoiceOption
+                AddModOptionMetadata<ModChoiceOption>(memberInfo, memberType);
+            }
+            else if (new[] { typeof(float), typeof(double), typeof(int) }.Contains(memberType)
```
PrimeSonic, line 171: You wouldn't be the first person I see to use `Contains` instead of chaining `||`s.  
Just keep it simple.
```CS
else if (memberType ==  typeof(float) ||
         memberType ==  typeof(double) ||
         memberType ==  typeof(int))
```

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,700 @@
...
+        /// <see cref="modOptionsMetadata"/> dictionary.
+        /// </summary>
+        /// <typeparam name="TModOption">The type of the <see cref="ModOption"/> to generate for this member.</typeparam>
+        /// <param name="memberInfo">The <see cref="MemberInfo"/> of the member.</param>
+        /// <param name="memberType">The underlying <see cref="Type"/> of the member.</param>
+        private void AddModOptionMetadata<TModOption>(MemberInfo memberInfo, Type memberType) where TModOption : ModOption
+        {
+            try
+            {
+                var modOptionType = typeof(TModOption); // cache the ModOption type for comparisons
```
PrimeSonic, line 202: You don't need to cache this.  
You are using a Generic, so the type is handled at compile time.  
`typeof` is not a method, it's a keyword. Using it doesn't incur a a runtime cost.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,700 @@
...
+            {
+                // Set the value in the Config
+                switch (modOptionMetadata.MemberInfo)
+                {
+                    case PropertyInfo property when property.PropertyType.IsEnum && modOptionMetadata.ChoiceAttribute == null:
+                        property.SetValue(Config, Enum.Parse(property.PropertyType, e.Value), null);
+                        break;
+                    case PropertyInfo property when property.PropertyType.IsEnum:
+                        property.SetValue(Config,
+                            Enum.Parse(property.PropertyType, Enum.GetNames(property.PropertyType)[e.Index]), null);
```
PrimeSonic, line 321: Either have all parameters of `SetValue` on their own lines (and aligned) or put them all on one line.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,700 @@
...
+                {
+                    e.GameObject.GetComponentInChildren<Text>().gameObject.AddComponent<ModOptionTooltip>().Tooltip
+                        = modOptionMetadata.TooltipAttribute.Tooltip;
+                }
+
+                if (modOptionMetadata.OnGameObjectCreatedAttributes == null)
+                    return; // Skip attempting to invoke events if there are no OnGameObjectCreatedAttributes set for the member.
+
+                // cache types used in comparisons
+                var objectType = typeof(object);
```
PrimeSonic, line 536: Same here, you don't need to create a variable to store the value of a `typeof` expression.  
It's not a method.

SMLHelper/Options/KeybindModOption.cs
```diff
@@ -15,7 +14,7 @@
     /// <summary>
     /// Contains all the information about a keybind changed event.
     /// </summary>
-    public class KeybindChangedEventArgs : EventArgs
+    public class KeybindChangedEventArgs : EventArgs, Interfaces.IModOptionEventArgs
```
PrimeSonic, line 17: Could we get `using Interfaces` up top to clean this up?

SMLHelper/Options/LabelAttribute.cs
```diff
@@ -0,0 +1,57 @@
...
+        /// internal comparisons, you can do so here.
+        /// </summary>
+        public string Id { get; set; } = Guid.NewGuid().ToString().Replace('-', '_');
+
+        /// <summary>
+        /// The order in which to display fields in the mod's option menu, in ascending order. If none is specified, the order will be
+        /// automatically set.
+        /// </summary>
+        public int Order { get; set; } = i++;
+        private static int i = 0;
```
PrimeSonic, line 47: This static counter here has me nervous.  
And given that you are still allowing `Order` to be changed via the `set`, I can only imagine that this will cause trouble down the line.

Example mod/Mod.cs
tobeyStraitjacket: > **NOTE:** There is no `KeybindAttribute` or `ToggleAttribute` as these are merely inferred by the field/property type:
`KeyCode` -> `ModKeybindOption`, `bool` -> `ModToggleOption`

I don't mind adding them if we really want them, it just seems like unnecessary code.

Example mod/Mod.cs
tobeyStraitjacket: > **NOTE:** There is no `KeybindAttribute` or `ToggleAttribute` as these are merely inferred by the field/property type:
`KeyCode` -> `ModKeybindOption`, `bool` -> `ModToggleOption`

I don't mind adding them if we really want them, it just seems like unnecessary code.

Example mod/Mod.cs
tobeyStraitjacket: Bc. if you are not ignoring unattributed members, then any public bool will be picked up as a toggle. If you are ignoring unattributed methods, just giving it a label will pick it up as a toggle, etc. Again, I can add them, it just seems like a waste of bytes.

SMLHelper/Json/ConfigFile.cs
tobeyStraitjacket: Not available.

SMLHelper/Options/ChoiceAttribute.cs
tobeyStraitjacket: The pre-built `ModChoiceOption` takes either an `object[]` array or `string[]` array in the constcutor so it will be coerced to that in the end regardless, so I declared it as one here.

SMLHelper/Options/ConfigModOptions.cs
tobeyStraitjacket: Good to know, thank you

SMLHelper/Options/ConfigModOptions.cs
tobeyStraitjacket: I'm currently rewriting this (see my latest comment about how this is currently a **draft** bc I'm working on optimisations)

Example mod/Mod.cs
PrimeSonic: Then maybe `Label` is the actual redundant attribute.  
If you need the common functionality, you can make `Slider`, `Toggle`, `KeyBind`, and `Choice` all inherit from a `Label` class or implement an `ILabel` interface.  

Then you'd get something a little cleaner
```CS
[Keybind(My keybind"), OnChange(nameof(MyGenericValueChangedEvent))]
public KeyCode KeybindKey;

[Toggle("My checkbox"), OnChange(nameof(MyCheckboxToggleEvent)), OnChange(nameof(MyGenericValueChangedEvent))]
public bool ToggleValue;
```

SMLHelper/Options/LabelAttribute.cs
tobeyStraitjacket: So far it does the job nicely. If you want to set the order manually, it's easy enough to document that you need to set the order for all of the members.

As for whether it has knock-on effects in other mods, this has yet to really be tested extensively. Ideally, I would be able to use the `[CallerLineNumber]` as the order in the constructor and then it's a very easily implied order based on the order of declaration - however that won't work across assemblies and definitely not in this version of .NET.

My simple take is this:
If you're using auto-order, it's not guaranteed to be the same every time.
If you want guaranteed order, set it for every member. Easy.

SMLHelper/Handlers/OptionsPanelHandler.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/aeda1362ca368e4531a7cfd93cb76712353e8d99

SMLHelper/Json/ConfigFile.cs
tobeyStraitjacket: https://docs.microsoft.com/en-us/dotnet/api/system.reflection.customattributeextensions.getcustomattributes?view=netcore-3.1&viewFallbackFrom=netframework-4.0#System_Reflection_CustomAttributeExtensions_GetCustomAttributes__1_System_Reflection_MemberInfo_System_Boolean_

SMLHelper/Options/ChoiceAttribute.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Options/ChoiceModOption.cs#L88

SMLHelper/Options/ConfigModOptions.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/88fbdffba0748285567429b4fdb7d1981ad5d731

SMLHelper/Options/ConfigModOptions.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/88fbdffba0748285567429b4fdb7d1981ad5d731

SMLHelper/Options/KeybindModOption.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/5aac61157d36e9f5703e88746e8a3b0fe08ef79c

SMLHelper/Options/ChoiceAttribute.cs
PrimeSonic: @AlexejheroYTB  Do you think it makes sense to update all these array parameters to `IEnumerable`s?  
Was there a reason they were arrays?

Example mod/Mod.cs
PrimeSonic: Resolving to maintain a single discussion  https://github.com/SubnauticaModding/SMLHelper/pull/173#discussion_r463896532

Example mod/Mod.cs
tobeyStraitjacket: > Then maybe `Label` is the actual redundant attribute.
> If you need the common functionality, you can make `Slider`, `Toggle`, `KeyBind`, and `Choice` all inherit from a `Label` class or implement an `ILabel` interface.
> 
> Then you'd get something a little cleaner
> 
> ```cs
> [Keybind(My keybind"), OnChange(nameof(MyGenericValueChangedEvent))]
> public KeyCode KeybindKey;
> 
> [Toggle("My checkbox"), OnChange(nameof(MyCheckboxToggleEvent)), OnChange(nameof(MyGenericValueChangedEvent))]
> public bool ToggleValue;
> ```

Good shout, don't know why I didn't think of this :D

SMLHelper/Options/ChoiceAttribute.cs
tobeyStraitjacket: Trawling through the code-base, it looks like eventually it leads here:
https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Options/ChoiceModOption.cs#L157
```cs
    public class ModChoiceOption : ModOption
    {
        /// <summary>
        /// The array of readable string options to choose between in the <see cref="ModChoiceOption"/>.
        /// </summary>
        public string[] Options { get; }
```

And from finding all references in VS, that property is used when called out to this method within `Assembly-CSharp.dll`:
`public uGUI_Choice AddChoiceOption(int tabIndex, string label, string[] items, int currentIndex, UnityAction<int> callback = null);`, passed as the `string[] items` parameter, as well as used for the `UnityAction`:

```cs
var choice = panel.AddChoiceOption(tabIndex, Label, Options, Index,
    new UnityAction<int>((int index) => parentOptions.OnChoiceChange(Id, index, Options[index])));
```

So it'll end up being coerced down to the a `string[]` array in the end.

SMLHelper/Options/ConfigModOptions.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/f1a5ef4766d9b25f06dec66050cd266f6ff9f5ae

Note: after refactor of the attributes (still pending) the Attribute.IsDefined() || chain happening here can potentially be changed to just a single Attribute.IsDefined for the base abstract type. And again, when we get updated to .NET 4.6.2, we can convert all of this to generics.

Example mod/Mod.cs
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/49f679b6b5125ba4bd0fe15cf0dfea2811557838

SMLHelper/Options/ChoiceAttribute.cs
PrimeSonic: Then I guess we can leave the explicit string array. No sense in upending everything over such a small detail.

SMLHelper/Options/ConfigModOptions.cs
PrimeSonic: Resolving now. Will look it over again after the rewrite.

SMLHelper/Options/LabelAttribute.cs
PrimeSonic: I'm more concerned about what happens when we get multiple mods using this method.  
I suppose that might not matter given that base on this code from `ConfigModOptions`, this is being used more as a relative ordering rather than an index.
```CS
public override void BuildModOptions()
{
    foreach (var entry in modOptionsMetadata
        .OrderBy(x => x.Value.ModOptionAttribute.Order)
        .ThenBy(x => x.Value.MemberInfo.Name))
```

Personally, I would have these ordered in the same way they appear in the class.  
`Type.GetProperties()` and `Type.GetFields()` will both return their respective members the the order they appear in the class.  
Don't know if that helps.

SMLHelper/Options/ModOptionEventAttribute.cs
```diff
@@ -0,0 +1,32 @@
...
+    using System;
+
+    /// <summary>
+    /// Abstract base attribute used to signify a method to call whenever the derivative event is invoked for the attributed member.
+    /// </summary>
+    /// <remarks>
+    /// The method must be a member of the same class.
+    /// </remarks>
+    [AttributeUsage(AttributeTargets.Field | AttributeTargets.Property | AttributeTargets.Method, AllowMultiple = true)]
+    public abstract class ModOptionEventAttribute : Attribute, IModOptionEventAttribute
```
PrimeSonic, line 13: Would limiting these attributes to only be used on strictly Properties (or strictly Fields) help simplify the rest of the backing code?

SMLHelper/Options/LabelAttribute.cs
tobeyStraitjacket: > I'm more concerned about what happens when we get multiple mods using this method.
> I suppose that might not matter given that, based on this code from `ConfigModOptions`, this is being used more as a relative ordering rather than an index.
> 
> ```cs
> public override void BuildModOptions()
> {
>     foreach (var entry in modOptionsMetadata
>         .OrderBy(x => x.Value.ModOptionAttribute.Order)
>         .ThenBy(x => x.Value.MemberInfo.Name))
> ```
> 
> Personally, I would have these ordered in the same way they appear in the class.
> `Type.GetProperties()` and `Type.GetFields()` will both return their respective members the the order they appear in the class.
> Don't know if that helps.

Yeah the issue with that is that if you have both fields and properties (and currently the design is to allow for both), then we have to make an arbitrary decision on which get listed first... fields or properties. And I don't like that. I would prefer them both to be listed in original order.

And yes, I've considered the case of multiple mods using it. It doesn't matter that mod #2 and mod #5 both use it, bc the number is just a relative order - it is used for ordering only in the local scope of the menu, and is just ordered in terms of "the lowest number is listed first". Whether it starts at 0 or 1000000 in your menu, it doesn't matter.

For what it's worth, from my testing so far, this method *does* list them in the correct order - the order they appear in the class, as the JIT is going through the class, one by one, evaluating the `[Toggle]` and then three lines down, you have a `[Keybind]` and it's processing those. Where it becomes not guaranteed is when you don't specify those yourself - then it becomes, automatically, order of the ones you declared attributes, then order of properties, then order of fields. And I personally think that's good enough?

SMLHelper/Options/ModOptionEventAttribute.cs
tobeyStraitjacket: Yes, but also no. I'm moving to `Traverse` for most of the getting/setting, so the backing code becomes automatically super simplified in that regard. I specifically want to allow for properties *and* fields because ConfigFile is based off of Newtonsoft.Json which also allows for both. And that means having to change the member design of your ConfigFile if you already use both in most cases, and I think that's not ideal.

Yes, the code-base for this is big. Of course it is - it's a big task it undertakes. But how "complex" it is is being worked on. Don't jump the gun on wanting to strip out half of it.

tobeyStraitjacket: PS: Side-benefit of moving to `Traverse` for the getting/setting is that it also caches the reflection lookups for us under-the-hood, so that's nice.

SMLHelper/Options/ModOptions.cs
```diff
@@ -14,7 +15,7 @@ public abstract partial class ModOptions
         /// <summary>
         /// The name of this set of configuration options.
         /// </summary>
-        public readonly string Name;
+        public string Name { get; protected set; }
```
zorgesho, line 18: Same as `AddSliderOption` here, it'll break any mod that uses this field. Though I don't think that anyone is using it outside of SMLHelper, so it's probably safe in this case.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+            => configModOptionsMetadata.ModOptionsMetadata;
+
+        /// <summary>
+        /// Process metadata for members of <typeparamref name="T"/>.
+        /// </summary>
+        private void ProcessMetadata()
+        {
+            var bindingFlags = BindingFlags.Instance | BindingFlags.Public;
+            foreach (PropertyInfo property in typeof(T).GetProperties(bindingFlags)
+                .Where(MemberIsDeclaredInConfigFileSubclass) // Only care about members declared in a subclass of ConfigFile
```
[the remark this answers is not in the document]
tobeyStraitjacket: I want to allow inheritance, but ignore the members from the `ConfigFile` itself.

SMLHelper/Options/ModOptionAttribute.cs
```diff
@@ -0,0 +1,50 @@
...
+        /// internal comparisons, you can do so here.
+        /// </summary>
+        public string Id { get; set; } = Guid.NewGuid().ToString().Replace('-', '_');
+
+        /// <summary>
+        /// The order in which to display fields in the mod's option menu, in ascending order. If none is specified, the order will be
+        /// automatically set.
+        /// </summary>
+        public int Order { get; set; } = i++;
+        private static int i = 0;
```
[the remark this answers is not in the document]
tobeyStraitjacket: I looked at this already and decided that:

1. since we upgrading to .NET Standard soon because of changes to Subnautica and Below Zero, hacking in support for features not available in .NET Framework 4.0 is pointless to do now

1. As mentioned in the resolved conversation, `[CallerLineNumber]` is a **compile time** feature. It is *not* available at *runtime*. This means that it *cannot* work across assemblies. A mod using the attribute will live in another assembly to SMLHelper, and therefore this class will have no way to access the *line numbers* of the *original file*, which result in all 0s, as seen in this example:
image

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+        {
+            if (Attribute.IsDefined(memberInfo, typeof(IgnoreMemberAttribute)))
+            {
+                return false;
+            }
+            else if (!menuAttribute.IgnoreUnattributedMembers)
+            {
+                if (memberInfo is MethodInfo)
+                {
+                    if (Attribute.IsDefined(memberInfo, typeof(ButtonAttribute), true))
```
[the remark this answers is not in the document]
tobeyStraitjacket: Specifying it to be explicit, so that everyone knows without having to go and look it up.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+        }
+        #endregion
+
+        #region Click, Change and GameObjectCreated Events
+        /// <summary>
+        /// Conditionally binds events, ie. OnButtonClicked, OnKeybindChanged, etc.
+        /// </summary>
+        private void BindEvents()
+        {
+            if (modOptionsMetadata.Values.Any(x => x.ModOptionAttribute is ButtonAttribute))
```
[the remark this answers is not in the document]
tobeyStraitjacket: Multiple empty events being called over and over and over again for every mod using this feature for no reason is heavier than a conditional check that happens once per mod.

SMLHelper/Options/SliderModOption.cs
```diff
@@ -86,9 +87,10 @@ protected void AddSliderOption(string id, string label, float minValue, float ma
         /// <param name="defaultValue">The default value for the slider. If this is null then 'value' used as default.</param>
         /// <param name="valueFormat"> format for value, e.g. "{0:F2}" or "{0:F0} %"
         /// (more on this <see href="https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings">here</see>)</param>
-        protected void AddSliderOption(string id, string label, float minValue, float maxValue, float value, float? defaultValue, string valueFormat = null)
+        /// <param name="step">Step for the slider, ie. round to nearest X</param>
+        protected void AddSliderOption(string id, string label, float minValue, float maxValue, float value, float? defaultValue, string valueFormat = null, float step = 0)
```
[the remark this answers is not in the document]
tobeyStraitjacket: Yes, that's worth considering - but they are all about to break, anyway. Have you seen beep keeping up-to-date with the updates coming down the pipe?

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+                    Timestamp = timestamp
+                };
+                Name = configModOptionsMetadata.MenuAttribute.Name;
+
+                ProcessMetadata();
+                SaveCache();
+            }
+        }
+
+        private void LoadCache()
```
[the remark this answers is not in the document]
tobeyStraitjacket: Tbh, I have wondered this. I mostly implemented it bc @PrimeSonic was concerned about LINQ+Reflection being too heavy, but IMO this caching approach is likely overzealous optimisation that might not actually be the "optimal" way anyway.

We could consider implementing a stopwatch to time how long each approach takes, but until we have numbers to work with its kind of impossible to know whether this serialized cache is worth it. It might turn out to be a worthwhile thing to turn on later, if users begin to report slow startup times.

Perhaps we could even make it an option. "Is startup a little slow for you? Try turning on the cache."

SMLHelper/Options/SliderModOption.cs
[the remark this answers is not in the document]
tobeyStraitjacket: However, fair point I'll just add an overload or smth.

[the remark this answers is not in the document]
tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173/commits/d983fa2f51c9ebcd4b7ae6d6cb3e9bf5a758ce7d

SMLHelper/Options/ModOptionAttribute.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Actually, for point 2, I was mistaken, and I must apologise - my testing methodology was imprecise and this can actually work.

I've come around on this - but I'd still rather wait until we officially move to .NET Standard to implement, as if we implement it before then, we'll have a namespace conflict.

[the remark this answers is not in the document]
zorgesho: Yeah, that line parameter should be in the derived attribute. And yeah, considering that exp branch uses 4.6.2, it's better to wait for update, but no one knows when it'll be.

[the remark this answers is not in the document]
tobeyStraitjacket: .NET Standard now actually, not even just .NET 4.6.2.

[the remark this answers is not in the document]
tobeyStraitjacket: I don't mind waiting for it to come to stable for the ordering to be perfect. It's better that than conflicting with the namespace so mods have yet another reason to be broken when the update comes.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: Those events are separate for each `ModOptions` instance, if you don't have particular option type in your config, event for that option type will not be raised for that instance.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: Yeah, I thought about that, but I couldn't come up with any example when it would be needed.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: `ConfigModOptions<T> where T : ConfigFile, new()` adds an event listener *conditionally* for each of those events, if `T` has a need for them. If we don't conditionally add the event listener, then whether you do or don't have that specific option type in your config, our internal event will be raised to loop through your applicable options, empty or not, to raise it. And `ConfigModOptions<T>` event will be raised whether or not you have any. Which, if there are none, is *more* wasteful than checking before we add the listener.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: Well, I don't think that everyone even know what that parameter is for. I'm using that `Attribute.IsDefined` in my code and I needed to "go and look it up" to find out why I don't use that parameter and should I use it :)

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Say we didn't check, and we added the event listeners on `ConfigModOptions<T>` like this:

```cs
private void BindEvents() {
  ButtonClicked += ConfigModOptions_ButtonClicked;
  ChoiceChanged += ConfigModOptions_ChoiceChanged;
  KeybindChanged += ConfigModOptions_KeybindChanged;
  SliderChanged += ConfigModOptions_SliderChanged;
  ToggleChanged += ConfigModOptions_ToggleChanged;
  GameObjectCreated += ConfigModOptions_GameObjectCreated;
}

private void ConfigModOptions_ChoiceChanged(object sender, ChoiceChangedEventArgs e)
{
  if (modOptionsMetadata.TryGetValue(e.Id, out var modOptionMetadata)
      && modOptionMetadata.MemberInfoMetadata.MethodParameterTypes is Type[] parameterTypes)
  {
      // do stuff
  }
}
// etc.
```

Then, any time any choice is changed, our ChoiceChanged event is fired up, and needlessly checks the dictionary for a value we already know won't be there. Every time. It is clearly therefore better to just not listen to the event if it's not needed.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: There is the potential that someone might want to derive from their own previous class, and I don't want to just say "no you can't do that" for an arbitrary reason when it's not that difficult to just determine whether the member is declared in a subclass of `ConfigFile`. Why limit for no good reason?

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: image

I'm talking about people working on the actual codebase. I'd rather be explicit than not. Just because it is the default does not mean you **should not define it,** it just means it is the default if you don't. So I'm not sure why you thought it necessary to leave a review telling me that it is default true as if you think I should remove it?

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: Hmm, I completely lost, is it possible to add option to the same `ModOptions` instance that handled by `ConfigFile` bypassing `ConfigFile` ?

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: It have overload without that param and that overload uses `inherit = true`. It's just harder to read, but if you think that that `true` absolutely needed to be there, ok then.

SMLHelper/Options/ModOptionAttribute.cs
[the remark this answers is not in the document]
zorgesho: It'll probably be more compiler warnings about conflicting types than actual crashes, but yeah, it's better to not add new hacks anyway.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Are you asking if its possible to add a modoption manually to the `ConfigModOptions<T>` instance? The answer is no, not through the publicly exposed API, because you never interact with the `ConfigModOptions<T> : ModOptions` instance directly, therefore you don't have access to the `ModOptions` methods for it. Furthermore, it is an `internal` class, so it also cannot be be publicly extended.

If you *did* have access to the `ModOptions` instance, even then you can't add an option to it from outside of it, as `AddChoiceOption` methods etc. are all `protected`.

So no, it is not possible to at runtime add more options to the menu. Because of this, once you register it to the menu, we already know everything we need to know, so we know whether or not we need to listen to the events.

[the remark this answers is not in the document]
zorgesho: >  any time any choice is changed in the menu for T, our ChoiceChanged event is fired up

Ok, so how `ChoiceChanged` would be fired up if you didn't add choice option with `ProcessMetadata()`?

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: I'm using profilers often with my reflection/linq/harmony stuff. LINQ/reflection can be slow, but I think it needed to be really heavy LINQ stuff to be even near disk operations. I mean, simple logging in the middle can change numbers a lot.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Ah, I see your logic now, and that part is fair enough. Still don't see the logic in subscribing to an event we don't need to subscribe to.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+            MemberInfoMetadata memberInfoMetadata, SliderAttribute sliderAttribute)
+        {
+            float value;
+            if (memberInfoMetadata.ValueType == typeof(float))
+            {
+                value = memberInfoMetadata.GetValue<float>(Config);
+            }
+            else if (memberInfoMetadata.ValueType == typeof(double))
+            {
+                value = Convert.ToSingle(memberInfoMetadata.GetValue<double>(Config));
```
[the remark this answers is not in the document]
tobeyStraitjacket: Since I'm the one who wrote the `GetValue` I am very aware that I have a version that works with `object`, and I do also know that `Convert.ToSingle` works with `object`. Please start assuming that there is a reason for my choices other than my just not knowing what I am doing, it's getting rather irksome.

The types are checked in `ProcessFieldOrProperty`, yes, so the chances of it falling apart when we reach this piece of code are impossible, so yes, technically it would be okay to call `Convert.ToSingle` on an `object`. However, this is more about bulletproofing it against future changes to the codebase. The logic is easier to deciper for someone else reading than if you're just working with an `object`. Furthermore, since there's no need to call `Convert.ToSingle` when we know the type is a float, I'd rather just compare the types.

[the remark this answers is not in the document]
zorgesho: Well, I think that
```
float value = Convert.ToSingle(memberInfoMetadata.GetValue(Config));
```
is easier to read and understand than all those lines, but maybe I'm wrong. If that was an intentional decision, ok then.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: > Still don't see the logic in subscribing to an event

To avoid running that LINQs and also less code -> more readable.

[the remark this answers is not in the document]
tobeyStraitjacket: If you want to avoid running LINQ then I could easily write a method to do it without the LINQ expression.

Less code is often (but not always) more readable, but it is not always better. This is getting subjective again.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: @PrimeSonic when you have a chance, I'd ask you to weigh in on this.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: That code that you wrote above (without checks) is easier to read and understand and it's also faster. And I think subscribing to the event that will never fire is a small price for that.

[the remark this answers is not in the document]
tobeyStraitjacket: @PrimeSonic Can I get a second opinion

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
zorgesho: On my machine using config with around 40 mixed members:
```
[SMLHelper/Info] LoadCache: 50.385 ms
[SMLHelper/Info] ProcessMetadata: 6.6284 ms
[SMLHelper/Info] SaveCache: 20.2819 ms
```
Also:
`[SMLHelper/Info] BindEvents: 0.6095 ms` - with LINQ checks
`[SMLHelper/Info] BindEvents: 0.1359 ms` - without checks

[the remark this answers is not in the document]
PrimeSonic: Worse case scenario, we add a couple seconds to the initial game load time.  
And from what we are seeing in Experimental, that is going to be going up significantly regardless of mods.  
So I will be easing off on any performance concerns, especially if it involves code that only runs once at start anyways.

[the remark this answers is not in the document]
tobeyStraitjacket: Yeah, I did think it might be the case that in many scenarios, the IO is going to be slower. Thank you for taking the time to confirm that.

As for the with/without LINQ checks for bind events, you might be onto something. I would prefer to also test it with alternative, non-LINQ checks that I have yet to write.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
PrimeSonic: I haven't had enough time (or mental energy) to fully take this all in.  
So I can't offer any meaningful feedback right now.  
After we go over it together in more detail, I may have something useful to contribute.

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,952 @@
...
+
+        #region Build ModOptions Menu
+        /// <summary>
+        /// Adds options to the menu based on the <see cref="modOptionsMetadata"/> dictionary, generated by the constructor.
+        /// </summary>
+        public override void BuildModOptions()
+        {
+            foreach (var entry in modOptionsMetadata
+                .OrderBy(x => x.Value.ModOptionAttribute.Order)
+                .ThenBy(x => x.Value.MemberInfoMetadata.Name))
```
PrimeSonic, line 809: There are some concessions I am willing to make in the name of simplifying some of this code.  

Consider allowing yourself to make these assumptions (_and then document them in the guide for this feature_):  
1. All relevant members of the class will be either Fields _or_ properties, but never a mix of both
2. Entries show up in the options menu in the same order as they appear in the class
   - There would be no way for the modders to explicitly specify an order
   - The class must have its fields/properties set up in the desired order

Take these as suggestions.  
I understand wanting to provide modders with the most flexible tool possible, but at some point, constraining their options is actually better for everyone.

tobeyStraitjacket: You're forgetting methods (for buttons). There's no reason to make it so that they will always be at the bottom (or top).

When we get the update to .NET Standard, we'll be able to use `[CallerLineNumber]` to have the order work in order of declaration regardless of whether it's fields, properties, etc, quite easily.

Once we have that switch, we can get rid of the `static int` increment. As for setting the order yourself, it won't be too much effort for me to make it so that there is always a gap of `1` or `5` or `10` between declared members (rather than however many lines there are), so it becomes easy to insert things before auto-ordered ones.

However, that all said, I do agree that order of declaration is simplest implementation and I find it hard to justify why you would need to manually set the order in any other fashion, so I don't really mind the loss of the feature.

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Resolved by 5259a2e

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Resolved by 4596875

SMLHelper/Options/ConfigModOptions.cs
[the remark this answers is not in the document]
tobeyStraitjacket: Resolved by c26bcdd

SMLHelper/Options/ModOptions.cs
tobeyStraitjacket: Yeah, I strongly doubt anyone is using this field outside of SMLHelper and if they are... probably a badly written mod in the first place? Not that I'm saying that makes it okay to cause them a breaking change (all they'd need to do is recompile) but... eh...

SMLHelper/Options/SliderModOption.cs
```diff
@@ -118,37 +137,56 @@ public class ModSliderOption : ModOption
...
         public float Step { get; } = 0.05f;
-#endif
 
         /// <summary> Format for value field (<see cref="ModOptions.AddSliderOption(string, string, float, float, float, float?, string)"/>) </summary>
         public string ValueFormat { get; }
 
         private SliderValue sliderValue = null;
 
+#if SUBNAUTICA
+        private float previousValue;
```
zorgesho, line 152: Btw, why is this needed? And only for SN?

tobeyStraitjacket: It is used here:
https://github.com/SubnauticaModding/SMLHelper/blob/88fc4806fd8b76cf5319a6fbc7d72df1965baefd/SMLHelper/Options/SliderModOption.cs#L162-L166

It tracks the previous value inside the custom SliderValue in SN (where we are implementing our own version of Slider step which hasn't reached SN yet) for triggering OnSliderChange, bc without it, we are trigging the change even when the *stepped to* value is the same bc the *original value* has changed ie.:
You have a slider that is set to step 5.
You move the slider from 0 to 5.
It triggers for 1, 2, 3, 4, 5, instead of just 5.
This behaviour is not what you would expect from a slider with a step of 5. You would expect it to trigger at 0, 5, 10, etc.
Tracking the previous value here mitigates that.

tobeyStraitjacket: Sorry, I remember it slightly wrong. It was actually triggering multiple times for the same, stepped, value, bc the underlying, non-stepped value has changed (0, 1, 2, 3 becomes 0, 0, 0, 5). So I am checking that the stepped-to value has actually changed before we trigger the OnSliderChange event.

zorgesho: Ah, I see

tobeyStraitjacket: Once the stepped slider comes down the pipe from BZ to SN we can discard this.

We could also potentially just not offer a stepped slider in SN at all until the official one comes along - I'm fine with that too.

zorgesho: Well, I added that step to example project because it's was easy way to show how `SliderValue` works, I don't think it's used for real. And I'm slightly worried that not calling `OnSliderChange` on the same value might break some code that expected that behaviour before.

tobeyStraitjacket: > And I'm slightly worried that not calling `OnSliderChange` on the same value might break some code that expected that behaviour before.

But they weren't using it before, because this is a new feature we are adding, so how can it? This is only used for stepped sliders - which don't exist in SN via SMLHelper until this update, unless you're using a custom slider, in which case you won't be using this code.

tobeyStraitjacket: I would also argue that a mod that is expecting to receive multiple repeats of the same value from a slider over and over and is *depending* on that... is an extremely strange (read: bad) mod for depending on such a thing.

zorgesho: I mean, there was no check on value before, `OnSliderChange` was called always, even for the same value. On the other hand, I'm not sure that `UnityAction` even called for the same values, so it's probably fine.

tobeyStraitjacket: Oh, normal sliders *don't* send the same value over and over again. btw. I tested that. It was only an issue with this custom stepped implementation - so I made it work the same as normal - *not* sending the same value multiple times for no reason.

SMLHelper/Options/MenuAttribute.cs
```diff
@@ -0,0 +1,142 @@
...
+    public sealed class MenuAttribute : Attribute
+    {
+        /// <summary>
+        /// Specifies after which events the config file should be saved to disk automatically.
+        /// </summary>
+        /// <remarks>
+        /// This enumeration has a <see cref="FlagsAttribute"/> that allows a bitwise combination of its member values.
+        /// </remarks>
+        [Flags]
+        public enum SaveEvents : short
```
MrPurple6411, line 95: As we discussed in discord. would like a LoadEvents added as well.

tobeyStraitjacket: 7cf2892 :)

tobeyStraitjacket: For your use-case, I recommend setting `LoadOn = MenuAttribute.LoadEvents.MenuRegistered | MenuAttribute.LoadEvents.MenuOpened` as this will load the config both on game start and whenever the options menu is opened.

SMLHelper/Options/ModOptions.cs
tobeyStraitjacket: Turns out someone was using it, so we've reverted it back to a `public field` (dropped the `readonly`).

In testing, this is working fine. It _does_ have the side-effect that now people could technically change their menu names on-the-fly - and even I suppose the names of other peoples' menus.

Technically, they could already do that with some reflection, but with this change they don't even need to know how to do that. So I'm a little worried that suddenly modders will be tempted to screw with each others' menu names... but probably not?

For anyone reading this comment and wondering what the purpose of this change is:
This PR is aiming to allow you to use Attributes to set the data needed for creating your menu. In the past, the only way to set the name of your menu was by passing it as a parameter to the base `ModOptions` constructor:

```cs
public class Options : ModOptions
{
    public Options() : base("My options menu") { }
}
```
And the `Name` field is then set in the `ModOptions` constructor from this parameter.

Since the `Name` field is `readonly`, this means that _only_ the original `ModOptions` constructor can set it - no derivatives can, not even in their constructors. And since we now want to set the `Name` from a derivative after parsing a `MenuAttribute` for its `Name` property, this means we're stuck unless we use reflection to set it (inefficient, feels dirty, not supported in .NET Core) or drop the `readonly` attribute from the field.

It's rather a shame it was chosen to be a `public readonly string` field. A `public` getter with a `protected` setter would have made for a much easier to extend API.

SMLHelper/Json/ConfigFile.cs
tobeyStraitjacket: We are not on .NET Framework 4.7.2, so I've implemented this change :) 5b5fe4e

SMLHelper/Options/ConfigModOptions.cs
```diff
@@ -0,0 +1,857 @@
...
+                var id = entry.Key;
+                var modOptionMetadata = entry.Value;
+
+                Logger.Debug($"[ConfigModOptions] {modOptionMetadata.MemberInfoMetadata.Name}: " +
+                    $"{modOptionMetadata.ModOptionAttribute.GetType()}");
+                Logger.Debug($"[ConfigModOptions] Label: {modOptionMetadata.ModOptionAttribute.Label}");
+
+                var label = modOptionMetadata.ModOptionAttribute.Label;
+
+                if (modOptionMetadata.ModOptionAttribute is ButtonAttribute)
```
PrimeSonic, line 737: This can be converted to a `switch` statement
```CS
switch (modOptionMetadata.ModOptionAttribute)
{
    case ButtonAttribute _:
        BuildModButtonOption(id, label);
        break;
    case ChoiceAttribute choiceAttribute:
        BuildModChoiceOption(id, label, modOptionMetadata.MemberInfoMetadata, choiceAttribute);
        break;
    case KeybindAttribute _:
        BuildModKeybindOption(id, label, modOptionMetadata.MemberInfoMetadata);
        break;
    case SliderAttribute sliderAttribute:
        BuildModSliderOption(id, label, modOptionMetadata.MemberInfoMetadata, sliderAttribute);
        break;
    case ToggleAttribute _:
        BuildModToggleOption(id, label, modOptionMetadata.MemberInfoMetadata);
        break;
}
```

tobeyStraitjacket: I prefer it too, I just hate the discard, but I guess its the same in the end. I'll convert.

tobeyStraitjacket: bbb3a18

SMLHelper/Options/Attributes/MenuAttribute.cs
```diff
@@ -0,0 +1,200 @@
...
+            /// </summary>
+            /// <remarks>
+            /// In normal usage, this option is equivalent to loading when the game is launched.
+            /// </remarks>
+            MenuRegistered = 1,
+            /// <summary>
+            /// Load when the menu is opened by the player.
+            /// </summary>
+            /// <remarks>
+            /// Useful for allowing players to edit their config files manually without restarting the game for their changes to take effect.
```
MrPurple6411, line 139: Can I get either an option to have this load on register as well when set to load on menu open or another option that does both as not everyone opens the menu when they go to start a new game if they have just changed the config file.

tobeyStraitjacket: It's a flags enum. You can do both by specifying to do both ;)

tobeyStraitjacket: https://github.com/SubnauticaModding/SMLHelper/pull/173#discussion_r484490882
^ as I said to you here ;)

MrPurple6411: I am an idiot.... proceed with yourself and ignore that i ever said this

MrPurple6411 (approved):

SMLHelper/Options/Attributes/ButtonAttribute.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// Signifies the attributed method should be represented in the mod's options menu as a <see cref="ModButtonOption"/>
+        /// and sets its label.
+        /// </summary>
+        /// <param name="label">The label for the button. If none is set, the name of the method will be used.</param>
+        public ButtonAttribute(string label = null) : base(label) { }
+
+        /// <summary>
+        /// Signifies the attributed method should be represented in the mod's options menu as a <see cref="ModButtonOption"/>.
+        /// </summary>
+        public ButtonAttribute() { }
```
PrimeSonic, line 47: There appears to be some ambiguity between these constructors.  
One is _parameterless_, the other has a single, _optional_ parameter.

_The same happens across many other attribute classes here_

Since the base class `ModOptionAttribute` only has a single constructor with an optional parameter, do these child classes need to have a default constructor?

SMLHelper/Options/Attributes/ConfigFileMetadata.cs
```diff
@@ -0,0 +1,628 @@
...
+            foreach (FieldInfo field in typeof(T).GetFields(bindingFlags)
+                .Where(memberIsDeclaredInConfigFileSubclass) // Only care about members declared in a subclass of ConfigFile
+                .Where(memberIsNotIgnored)) // Filter out explicitly ignored members
+            {
+                processFieldOrProperty(field, MemberType.Field, field.FieldType);
+            }
+
+            foreach (MethodInfo method in typeof(T).GetMethods(bindingFlags | BindingFlags.Static)
+                .Where(memberIsDeclaredInConfigFileSubclass) // Only care about members declared in a subclass of ConfigFile
+                .Where(memberIsNotIgnored)) // Filter out explicitly ignored members
```
PrimeSonic, line 81: `BindingFlags.Static`  
Will this only pick up static methods?

SMLHelper/Options/Attributes/MenuAttribute.cs
```diff
@@ -0,0 +1,200 @@
...
+            /// <summary>
+            /// Only <see langword="public"/> members attributed with a <see cref="ModOptionAttribute"/> derivative such as 
+            /// <see cref="SliderAttribute"/>, <see cref="ChoiceAttribute"/> etc. will be processed.
+            /// </summary>
+            OptIn,
+            /// <summary>
+            /// All <see langword="public"/> members will be processed, and where they are not attributed, a default attribute
+            /// will be generated for them.
+            /// </summary>
+            OptOut
```
PrimeSonic, line 158: Consider alternatives to `OptOut` since it doesn't exactly describe what this is for.

`IncludeAll` or `Implicit` or maybe something around there?

SMLHelper/Options/Attributes/ButtonAttribute.cs
PrimeSonic: Found `new TAttribute()`

SMLHelper/Options/Attributes/ConfigFileMetadata.cs
tobeyStraitjacket: With BindingFlags, only flags that are provided will be looked at.

```cs
var bindingFlags = BindingFlags.Instance | BindingFlags.Public;
```
```cs
foreach (MethodInfo method in typeof(T).GetMethods(bindingFlags | BindingFlags.Static)
```
Resulting in a combination of `Instance`, `Public` and `Static`, so only public methods will be looked at, and they can be either Instance or Static.

SMLHelper/Options/Attributes/ButtonAttribute.cs
tobeyStraitjacket: Yup, was about to say. It's currently solely because of this constraint, because we need/want to be able to create a default instance in some cases. I wish that `new()` took optional parameters into consideration.

SMLHelper/Options/Attributes/MenuAttribute.cs
tobeyStraitjacket: How about `OptIn` = `Explicit` & `OptOut` = `Implicit`?

tobeyStraitjacket: But my reasons for `OptOut` is that you can still opt out - it will imply that this is an attribute unless you opt-out of it by using the `IgnoreMemberAttribute`. Similar in design to Newtonsoft.Json's `OptIn` vs `OptOut` + `JsonIgnore`.

tobeyStraitjacket: With `OptIn` & `OptOut`, I do feel like you are right in that if you're looking at it, you have to remind yourself which one results in which behaviour. `Explicit` & `Implicit` I think is likely easier to grasp. So I'm on-board with the change.

PrimeSonic: I like `Explicit`/`Implicit` 👍

tobeyStraitjacket: 81d410e :)

PrimeSonic (approved):
While I would prefer to be able to fully wrap my head around how this all works, for now, I am on board with the usage of this new tool.  
The way it's showcased in the example mod is quite nice and I could get behind it.  
We may later find small optimizations to make or other code organization, but it's good enough for now.

_Reminder to create a new wiki page for this before release._

PR #176, opened 2020-08-11
Move Prefab Non Database Patches to PreInitialize.

MrPurple6411:
Was driving myself crazy trying to make my new Customize Your Spawn mod working with modded techtypes when I realized that the reason it fails is because we don't patch these 3 prefab getters until after postPatch meaning I would have had to make my mod run AFTER SMLHelper for the prefabs to work.  

This keeps the current functionality of adding the prefabs to the base games code in meta post patch while still allowing mods to get the required information through the normal game code means...

  - Created a Prefix patch for the lookup TryGetPrefabFilename to allow mods to get the required information to add modded items to the spawning system.
  - Moved the harmony patching of the getter methods to PrePatch so that mods trying to use them will correctly search the modded prefabs as well as the games prefabs

### DIFF
--- a/SMLHelper/Initializer.cs
+++ b/SMLHelper/Initializer.cs
@@ -6,6 +6,7 @@
     using Patchers;
     using Patchers.EnumPatching;
     using QModManager.API.ModLoading;
+    using QModManager.Utility;
 
     /// <summary>
     /// WARNING: This class is for use only by QModManager.
@@ -14,6 +15,8 @@
     [Obsolete("This class is for use only by QModManager.", true)]
     public class Initializer
     {
+        private static Harmony harmony = new Harmony("com.ahk1221.smlhelper");
+
         /// <summary>
         /// WARNING: This method is for use only by QModManager.
         /// </summary>
@@ -32,6 +35,15 @@
             TechTypePatcher.cacheManager.LoadCache();
             Logger.Debug("Loading CraftTreeType Cache");
             CraftTreeTypePatcher.cacheManager.LoadCache();
+
+            try
+            {
+                PreInitialize();
+            }
+            catch(Exception e)
+            {
+                Logger.Error($"Caught exception while trying to initialize SMLHelper{Environment.NewLine}{e}");
+            }
         }
 
         /// <summary>
@@ -43,7 +55,7 @@
         {
             try
             {
-                Initialize();
+                PostInitialize();
             }
             catch (Exception e)
             {
@@ -51,10 +63,13 @@
             }
         }
 
-
-        private static void Initialize()
+        private static void PreInitialize()
         {
-            var harmony = new Harmony("com.ahk1221.smlhelper");
+            PrefabDatabasePatcher.PrePatch(harmony);
+        }
+
+        private static void PostInitialize()
+        {
 
             FishPatcher.Patch(harmony);
 
@@ -67,7 +82,7 @@
             CraftTreePatcher.Patch(harmony);
             DevConsolePatcher.Patch(harmony);
             LanguagePatcher.Patch(harmony);
-            PrefabDatabasePatcher.Patch(harmony);
+            PrefabDatabasePatcher.PostPatch(harmony);
             SpritePatcher.Patch();
             KnownTechPatcher.Patch(harmony);
             BioReactorPatcher.Patch();
--- a/SMLHelper/Patchers/PrefabDatabasePatcher.cs
+++ b/SMLHelper/Patchers/PrefabDatabasePatcher.cs
@@ -29,6 +29,16 @@
             return true;
         }
 
+        internal static bool TryGetPrefabFilename_Prefix(string classId, ref string filename)
+        {
+            if (ModPrefab.TryGetFromClassId(classId, out ModPrefab prefab))
+            {
+                filename = prefab.PrefabFileName;
+                return false;
+            }
+            return true;
+        }
+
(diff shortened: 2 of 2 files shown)


SMLHelper/Initializer.cs
```diff
@@ -14,6 +15,8 @@
     [Obsolete("This class is for use only by QModManager.", true)]
     public class Initializer
     {
+        private static Harmony harmony = new Harmony("com.ahk1221.smlhelper");
```
PrimeSonic, line 18: make field `readonly`

SMLHelper/Initializer.cs
```diff
@@ -32,6 +35,15 @@ public static void PrePatch()
             TechTypePatcher.cacheManager.LoadCache();
             Logger.Debug("Loading CraftTreeType Cache");
             CraftTreeTypePatcher.cacheManager.LoadCache();
+
+            try
+            {
+                PreInitialize();
```
PrimeSonic, line 41: If you're going to create a new method to add prepatch orchestration, move everything into either the `PreInitialize` method or into `PrePatch` directly.

SMLHelper/Initializer.cs
```diff
@@ -32,6 +35,15 @@ public static void PrePatch()
...
+            }
+            catch(Exception e)
+            {
+                Logger.Error($"Caught exception while trying to initialize SMLHelper{Environment.NewLine}{e}");
```
PrimeSonic, line 45: Remove the try-catch.
Let QModManager handle any exceptions that happen during patching.

SMLHelper/Initializer.cs
```diff
@@ -43,18 +55,21 @@ public static void PostPatch()
         {
             try
             {
-                Initialize();
+                PostInitialize();
             }
             catch (Exception e)
             {
                 Logger.Error($"Caught exception while trying to initialize SMLHelper{Environment.NewLine}{e}");
```
PrimeSonic, line 62: For consistency with current recommended best practices, remove the try-catch from this method as well.  
QModManager will catch the exception, log it, and also inform the player that something went wrong.

Catching and logging like this can hide errors from the player.

SMLHelper/Patchers/PrefabDatabasePatcher.cs
```diff
@@ -42,18 +53,26 @@ internal static bool GetPrefabAsync_Prefix(ref IPrefabRequest __result, string c
...
         {
-            harmony.Patch(AccessTools.Method(typeof(PrefabDatabase), nameof(PrefabDatabase.LoadPrefabDatabase)),
-                postfix: new HarmonyMethod(AccessTools.Method(typeof(PrefabDatabasePatcher), nameof(PrefabDatabasePatcher.LoadPrefabDatabase_Postfix))));
-
             harmony.Patch(AccessTools.Method(typeof(PrefabDatabase), nameof(PrefabDatabase.GetPrefabForFilename)), 
                 prefix: new HarmonyMethod(AccessTools.Method(typeof(PrefabDatabasePatcher), nameof(PrefabDatabasePatcher.GetPrefabForFilename_Prefix))));
 
+            harmony.Patch(AccessTools.Method(typeof(PrefabDatabase), nameof(PrefabDatabase.TryGetPrefabFilename)),
+                prefix: new HarmonyMethod(AccessTools.Method(typeof(PrefabDatabasePatcher), nameof(PrefabDatabasePatcher.TryGetPrefabFilename_Prefix))));
+
```
PrimeSonic, line 63: For performance, this Harmony pre-patch should be unpatched after `LoadPrefabDatabase_Postfix()` has run.  
All modded prefabs should be in the database once completed.

SMLHelper/Initializer.cs
MrPurple6411: WOOPS YUP

SMLHelper/Patchers/PrefabDatabasePatcher.cs
MrPurple6411: Guess I am still learning lol ..... How exactly do you Unpatch a patch?

PrimeSonic: Ask @tobeyStraitjacket or look at how he unpatches the QModManager Initialize method.
https://github.com/SubnauticaModding/QModManager/blob/master/QMMLoader/QMMLoader.cs#L62

PrimeSonic: > https://github.com/SubnauticaModding/SMLHelper/pull/176/commits/b066fdaf2c028b221a472ccd00de3cf49dea41a0

No no, this isn't right.  
The new `TryGetPrefabFilename_Prefix` should be unpatched as part of the _execution_ of `LoadPrefabDatabase_Postfix`, so that this prefix patch is still there until we can be certain that all mod prefabs have been copied into the prefab database.  
Once copied, then the prefix patch is no longer needed.

MrPurple6411: Fixed as of cd9968f

tobeyStraitjacket: I would say it might also be worth saving a reference to the targets so you don't need to call `AccessTools.Method` multiple times, just for a bit of extra performance, as I don't think `AccessTools` caches the lookups. But it might not really be that much of a difference.

PR #177, opened 2020-08-16
Fix for Fixer

zorgesho:

  - `Fixer` is not adding to actual prefabs (e.g. loaded from asset bundles)
  - Prefab game object is not registered as `LargeWorldEntity` (registering only builded object now)
  - `IProtoEventListener` and its methods are removed from `Fixer`
  - `SceneCleanerPreserve` component is not adding to the prefab now

### DIFF
--- a/SMLHelper/Assets/ModPrefab.cs
+++ b/SMLHelper/Assets/ModPrefab.cs
@@ -81,28 +81,25 @@
             go.transform.position = new Vector3(-5000, -5000, -5000);
             go.name = this.ClassID;
 
-            /* Make sure prefab doesn't get cleared when quiting game to menu. */
-            SceneCleanerPreserve scp = go.AddComponent<SceneCleanerPreserve>();
-            scp.enabled = true;
-
             if (this.TechType != TechType.None)
             {
-                go.AddComponent<Fixer>().techType = this.TechType;
+                if (go.activeInHierarchy) // don't add Fixer to inactive prefabs
+                    go.AddComponent<Fixer>();
 
-                if (go.GetComponent<TechTag>() != null)
+                if (go.GetComponent<TechTag>() is TechTag tag)
                 {
-                    go.GetComponent<TechTag>().type = this.TechType;
+                    tag.type = this.TechType;
                 }
 
-                if (go.GetComponent<Constructable>() != null)
+                if (go.GetComponent<Constructable>() is Constructable cs)
                 {
-                    go.GetComponent<Constructable>().techType = this.TechType;
+                    cs.techType = this.TechType;
                 }
             }
 
-            if (go.GetComponent<PrefabIdentifier>() != null)
+            if (go.GetComponent<PrefabIdentifier>() is PrefabIdentifier pid)
             {
-                go.GetComponent<PrefabIdentifier>().ClassId = this.ClassID;
+                pid.ClassId = this.ClassID;
             }
 
             return go;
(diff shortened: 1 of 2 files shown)


SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -7,61 +7,35 @@ namespace SMLHelper.V2.MonoBehaviours
     /// <summary>
     /// This <see cref="MonoBehaviour"/> is <c>automatically</c> added to any <see cref="GameObject"/> created through <see cref="ModPrefab.GetGameObject"/>.
     /// </summary>
-    public class Fixer : MonoBehaviour, IProtoEventListener
```
zorgesho, line 10: I think it's safe to remove `IProtoEventListener` here and also make `Fixer` internal, because its lifetime is just one second. But maybe I'm missing something ?

SMLHelper/Assets/ModPrefab.cs
```diff
@@ -81,28 +81,25 @@ internal GameObject GetGameObjectInternal()
...
 
-            /* Make sure prefab doesn't get cleared when quiting game to menu. */
-            SceneCleanerPreserve scp = go.AddComponent<SceneCleanerPreserve>();
-            scp.enabled = true;
-
             if (this.TechType != TechType.None)
             {
-                go.AddComponent<Fixer>().techType = this.TechType;
+                if (go.activeInHierarchy) // don't add Fixer to inactive prefabs
+                    go.AddComponent<Fixer>();
```
zorgesho, line 87: If prefab is not active (e.g. it's loaded from asset bundle) then `Fixer` will not destroy it (because it'll not updated) and next `Fixers` will be added to the same prefab (e.g. BaseClocks mod works that way).

MrPurple6411: I understand the first part but why remove the assigning of the techtype?

zorgesho: Maybe I'm wrong here, but I think techtype was needed when Fixer used ProtoSerialize methods. Prefabs already have `TechTag` and `Constructable` with techtypes and techtype in Fixer is not used.

SMLHelper/MonoBehaviours/Fixer.cs
PrimeSonic: Can you confirm that without `IProtoEventListener`, the cloned objects don't just return upon loading the save again?

SMLHelper/MonoBehaviours/Fixer.cs
```diff
@@ -7,61 +7,35 @@ namespace SMLHelper.V2.MonoBehaviours
...
+    internal class Fixer : MonoBehaviour
```
PrimeSonic, line 10: Since this has been `public` for so long, it's hard to know for sure if there aren't mods out there that are adding their own `Fixer` components manually.

For the sake of not breaking anything, let's keep it `public` just to be safe.  
There's no harm in allowing this to be `public` anyways.

SMLHelper/MonoBehaviours/Fixer.cs
zorgesho: No, they shouldn't. I did a lot of tests with different mods and loading/saving (including quitting to main menu and loading again). And Fixer either removing prefab or removing itself from object, so it won't be saved anyway.
While loading, Fixers are created the same way, by ModPrefab.
I think that maybe `IProtoEventListener` was needed when Fixer didn't remove itself from object.

PrimeSonic: Works for me

SMLHelper/Assets/ModPrefab.cs
PrimeSonic: If they're no longer needed, for the same reason that ProtoSerialize isn't needed anymore, I see no reason not to simplify the code.

zorgesho:
> Will this be backwards compatible?

Strictly speaking no, it won't be backward compatible. I'm going to revert `Fixer` to public, but couple of public members that were used in `ProtoSerializer` will be removed. But because of one-second lifetime of that component I don't think they were actually used.

PR #181, opened 2020-08-25
Configurations for experimental branches

zorgesho:

- four folders with dependencies for each branch now (no common dependencies).
- solution/projects configurations are updated to SN.STABLE, SN.EXP, BZ.STABLE, BZ.EXP.
- new conditional symbols are added - `SUBNAUTICA_STABLE`, `SUBNAUTICA_EXP`, `BELOWZERO_STABLE`, `BELOWZERO_EXP`.
`SUBNAUTICA` and `BELOWZERO` are remained and mean both branches for corresponding game.


MrPurple6411:
1.  why have periods in the folder name?  please no??
2. harmony is the same for all branches because it comes from qmm not the game meaning it has no reason to be duplicated.

AlexejheroYTB (commented):
I agree to MrPurple about the folder and configuration names. They should contain `_` instead of `.`.

zorgesho:
> why have periods in the folder name?
what's wrong with periods?

> harmony is the same for all branches, it has no reason to be duplicated.
yeah, that just for consistency. And with addition of new DLLs there will be no need to think, is it gonna change on some branch or not (e.g. like it was with Json).

SMLHelper/SMLHelper.csproj
```diff
@@ -20,113 +21,99 @@
   <PropertyGroup>
     <StartupObject />
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'SUBNAUTICA|AnyCPU'">
-    <OutputPath>bin\SUBNAUTICA\</OutputPath>
-    <DefineConstants>TRACE;DEBUG;SUBNAUTICA</DefineConstants>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'SN.STABLE|AnyCPU'">
```
zorgesho, line 24: @MrPurple6411 @AlexejheroYTB So, why you don't want periods in names? I'm using them quite often, both in folders and configuration names, didn't have any problems with them. There is also SMLHelper.Tests folder right in the root.

PR #183, opened 2020-08-27
SMLHelper 2.8.4 Different Sprite issue Hotfix.

MrPurple6411:
- Sprite Key already Exists patch.
  - Added Debug logging when sprite key already exists for both modsprites and game sprite.


SMLHelper/Assets/SubnauticaModSprite.cs
```diff
@@ -19,6 +19,8 @@ internal static void Add(SpriteManager.Group group, string name, Atlas.Sprite sp
             if (!ModSprites.ContainsKey(group))
                 ModSprites.Add(group, new Dictionary<string, Atlas.Sprite>(StringComparer.InvariantCultureIgnoreCase));
 
+            if(ModSprites[group].ContainsKey(name))
+                Logger.Debug($"ModSprite already registered for {group}/{name}.  Old sprite will be overwritten.");
```
AlexejheroYTB, line 23: In the case of a conflict, shouldn't both entries be thrown out, similarly to what happens with other data types?

MrPurple6411: That would 1. alter the effect that was already present. and 2. cause whatever mods added those sprites to fail to find them when they looked causing the ? icon instead.

MrPurple6411: this way just adds logging to the debug level to show WHEN one mod is overwriting another mods sprite.

PrimeSonic: @AlexejheroYTB  We no longer discard both entries when in conflict (to avoid worse problems).  
We keep the latest entry and log when an overwrite took place..

tobeyStraitjacket (commented):
Why are we splitting `0Harmony.dll` across exp/stable & SN/BZ? It's the same for all 4 game versions and that is unlikely to change.

EDIT: nvm

tobeyStraitjacket (approved):

PR #184, opened 2020-08-28
Asynchronous loading for ModPrefab

zorgesho:

**public changes:**
- asynchronous virtual method `IEnumerator GetGameObjectAsync(IOut<GameObject> gameObject)` added to the `ModPrefab`.
It is supposed to be used like `GetGameObject`, only all sync operations with game prefabs should be changed to async (examples will be added later).
- `AddPrefabCopy` method added to the `ModPrefabCache` class. It can be used for instantiating prefab and adding that copy to the cache at the same time. That way instantiated copy will not run `Awake()`, `Start()` and `Update()` in both sync and async versions of `GetGameObject`.

**internal changes:**
- `ModPrefabRequest` class added to the `SMLHelper.V2.Assets` namespace. It is used to get modded prefabs asynchronously (it calls `ModPrefab.GetGameObjectInternalAsync`)
- various changes in `PrefabDatabasePatcher` (new patch for object deserialization, using of `ModPrefabRequest` and some refactoring)

All new code supposed to work on both SN branches. Modders can add `GetGameObjectAsync` methods to their prefabs along with `GetGameObject`. For now, on the stable branch, `GetGameObjectAsync` doesn't do/break anything. And exp branch for now can be used to test `GetGameObjectAsync` (`GetGameObject` shouldn't break anything on exp branch).
So mods can be updated to use async prefabs ahead of update and when async update will hit stable branch, mods will not be broken (at least because of that).

### DIFF
--- a/SMLHelper/Assets/ModPrefab.cs
+++ b/SMLHelper/Assets/ModPrefab.cs
@@ -1,6 +1,7 @@
 ﻿namespace SMLHelper.V2.Assets
 {
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using UnityEngine;
 
@@ -71,12 +72,28 @@
         internal GameObject GetGameObjectInternal()
         {
             GameObject go = GetGameObject();
-
-            if (go == null)
-            {
+            if (!go)
                 return null;
-            }
 
+            ProcessPrefab(go);
+            return go;
+        }
+
+        internal IEnumerator GetGameObjectInternalAsync(IOut<GameObject> gameObject)
+        {
+            var taskResult = new TaskResult<GameObject>();
+            yield return GetGameObjectAsync(taskResult);
+
+            GameObject go = taskResult.Get();
+            if (!go)
+                yield break;
+
+            ProcessPrefab(go);
+            gameObject.Set(go);
+        }
+
+        private void ProcessPrefab(GameObject go)
+        {
             if (go.activeInHierarchy) // inactive prefabs don't need to be removed by cache
                 ModPrefabCache.AddPrefab(go);
 
@@ -99,15 +116,17 @@
             {
                 pid.ClassId = this.ClassID;
             }
-
-            return go;
         }
+
 
         /// <summary>
         /// Gets the prefab game object. Set up your prefab components here.
         /// The <see cref="TechType"/> and ClassID are already handled.
         /// </summary>
         /// <returns>The game object to be instantiated into a new in-game entity.</returns>
-        public abstract GameObject GetGameObject();
+        public virtual GameObject GetGameObject() => null; // SUBNAUTICA_EXP TODO: remove (or make obsolete) for SN after async update
+
+        /// <summary> TODO </summary>
+        public virtual IEnumerator GetGameObjectAsync(IOut<GameObject> gameObject) => null;
     }
 }
--- a/SMLHelper/Assets/ModPrefab.cs
+++ b/SMLHelper/Assets/ModPrefab.cs
@@ -72,7 +72,7 @@
         internal GameObject GetGameObjectInternal()
         {
             GameObject go = GetGameObject();
-            if (!go)
+            if (go == null)
                 return null;
 
             ProcessPrefab(go);
@@ -85,7 +85,7 @@
             yield return GetGameObjectAsync(taskResult);
 
             GameObject go = taskResult.Get();
-            if (!go)
+            if (go == null)
                 yield break;
 
             ProcessPrefab(go);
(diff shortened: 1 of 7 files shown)


SMLHelper/Assets/ModPrefabRequest.cs
```diff
@@ -0,0 +1,52 @@
...
+        public bool TryGetPrefab(out GameObject result)
+        {
+            result = taskResult.Get();
+            return result != null;
+        }
+
+        public bool MoveNext()
+        {
+            Init();
+            return task.GetResult() == null; // TODO: possible infinite loop ?
```
AlexejheroYTB, line 47: Should be fixed before merge.

SMLHelper/Assets/ModPrefabCache.cs
```diff
@@ -59,9 +59,23 @@ private static void Init()
         public static void AddPrefab(GameObject prefab, bool autoremove = true)
         {
             Init();
-
             prefab.transform.parent = prefabRoot.transform;
 
+            AddPrefabInternal(prefab, autoremove);
+        }
+
+        /// <summary> TODO </summary>
```
AlexejheroYTB, line 67: Should be done before merge

SMLHelper/Assets/ModPrefab.cs
```diff
@@ -99,15 +116,17 @@ internal GameObject GetGameObjectInternal()
...
         }
 
+
         /// <summary>
         /// Gets the prefab game object. Set up your prefab components here.
         /// The <see cref="TechType"/> and ClassID are already handled.
         /// </summary>
         /// <returns>The game object to be instantiated into a new in-game entity.</returns>
-        public abstract GameObject GetGameObject();
+        public virtual GameObject GetGameObject() => null; // SUBNAUTICA_EXP TODO: remove (or make obsolete) for SN after async update
```
AlexejheroYTB, line 127: What's going on with the TODOs from #181?
Also, this should be done before merge.

AlexejheroYTB (changes requested):
I dont think we should merge any PRs which have TODOs.

SMLHelper/Assets/ModPrefab.cs
```diff
@@ -71,12 +72,28 @@ protected ModPrefab(string classId, string prefabFileName, TechType techType = T
         internal GameObject GetGameObjectInternal()
         {
             GameObject go = GetGameObject();
-
-            if (go == null)
-            {
+            if (!go)
```
PrimeSonic, line 75: Please don't use this syntax for null checks.  
Keep it as `go == null` or `go != null`.

There's no need to add this potentially confusing syntax for what is likely no real benefit to the compiler.

SMLHelper/Assets/ModPrefab.cs
zorgesho: This one (and others that marked with SUBNAUTICA_EXP TODO) should be fixed after SN async update, it's just a reminder. Others TODOs I'll fix before merge of course, I'm still testing all that on my mods.
I can switch it to draft, but with exceptions of TODOs all code is working now.

SMLHelper/Patchers/PrefabDatabasePatcher.cs
```diff
@@ -1,82 +1,134 @@
...
             {
-                filename = prefab.PrefabFileName;
-                __result = true;
-                return false;
+                // trying sync method for backward compatibility
+                if (prefab.GetGameObjectInternal() is GameObject go) // SUBNAUTICA_EXP TODO: remove for SN after async update
+                    return new LoadedPrefabRequest(go);
             }
-            return true;
+            catch (Exception) {}
```
PrimeSonic, line 67: Empty catch block?  
Shouldn't we be at least logging something here?

SMLHelper/Patchers/PrefabDatabasePatcher.cs
```diff
@@ -1,82 +1,134 @@
...
+            if (GetModPrefabAsync(gameObjectData.ClassId) is IPrefabRequest request)
+            {
+                yield return request;
 
-                return false;
+                if (request.TryGetPrefab(out GameObject prefab))
+                {
+                    result.Set(UnityEngine.Object.Instantiate(prefab).GetComponent<UniqueIdentifier>());
+                    yield break;
+                }
```
PrimeSonic, line 103: What happens if this `TryGetPrefab` returns false?

SMLHelper/Patchers/PrefabDatabasePatcher.cs
zorgesho: It's for the time after async update, when sync methods will not work. Before update that catch is not supposed to catch anything, and after update there will be probably lot of exceptions here. And after exception this patch will try async request.
Probably not a perfect solution and this block is supposed to be removed after async update anyway.

SMLHelper/Patchers/PrefabDatabasePatcher.cs
zorgesho: Then `ProtobufSerializer.InstantiatePrefabAsync` in line 106 supposed to be used.

SMLHelper/Assets/ModPrefab.cs
PrimeSonic: Could you at least at a little more description to the `TODO` comments?  
Something like this
```CS
// TODO - Remove this block after SN1 Experimental changes are in Stable
```

PrimeSonic: As long as the `TODO` comment explains what we need to do and when, I'm fine with these.

`TODO`s in XML documentation should be updated as soon as they can (assuming we aren't expecting drastic changes to the documentation.

AlexejheroYTB (approved):

SMLHelper/Assets/ModPrefab.cs
zorgesho: After some testing I realized that sync `GetGameObject()` still can be useful in exp branch if it used with asset bundles, for example. If it doesn't require access to the game prefabs, it can work synchronously. So I remove this specific TODO and couple of others. For now, it's just one TODO left, for the `PrefabDatabase.GetPrefabForFilename` patch, that method is removed in the exp branch.

PR #199, opened 2020-12-20
BasicText - easy-to-use text items to add to Subnautica screen

Cattlesquat:
Super-simple text items (literally 2 lines to get a basic one going -- see main comments for examples). Default is to "clone" the SN in-game text so that you get attractive-looking and SN-suitable text by default w/o having to figure out e.g. what Material to use and what font and etc, etc.  BUT you can set those things too if you want, and it supports "fade out after X seconds".

I don't know how to make the comments the way it seems to want me to (I come from a JavaDoc world and formatted them that way). But anyway have a look.

I have learned SN engine would be more happy if we set layer to "31" (its highest layer) than any other value, though that DOES mean it wouldn't appear on top of the "black screen" when you die. Of course most users probably don't need that anyway.

### DIFF
--- a/SMLHelper/Utility/BasicText.cs
+++ b/SMLHelper/Utility/BasicText.cs
@@ -369,7 +369,7 @@ namespace SMLHelper.V2.Utility
                 case TextAnchor.LowerLeft:
                 case TextAnchor.LowerCenter:
                 case TextAnchor.LowerRight:
-                    displayY = y - height / 2;
+                    displayY = y + height / 2;
                     break;
 
                 default:
--- a/SMLHelper/Utility/BasicText.cs
+++ b/SMLHelper/Utility/BasicText.cs
@@ -1,0 +1,436 @@
+﻿/**
+ * BasicText -- Cattlesquat
+ * 
+ * Places a simple text object on the screen and keeps it there until either hidden (or a designated fade-to-black timer has passed). 
+ * By default uses the same font/size/color as the "Press Any Button To Begin" message at the beginning of the game, and appears 
+ * centered about 1/3 down the screen, but all parameters can be reconfigured.
+ * 
+ * The idea of the defaults is that new modders don't have to bootstrap a bunch of irritating Unity stuff -- don't have to understand
+ * what a "Material" is or how to make one, don't have to know to initialize a font, or even a color. Can just start adding text and
+ * then can always custom & configure on further revision.
+ * 
+ * SIMPLE USAGE EXAMPLE:
+ * BasicText message = new BasicText();
+ * message.ShowMessage("This Message Will Fade In 10 Seconds", 10);
+ * 
+ * COMPLEX USAGE EXAMPLE:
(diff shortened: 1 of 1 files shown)


SMLHelper/Utility/BasicText.cs
```diff
@@ -0,0 +1,435 @@
...
+        protected TextAnchor align { get; set; }      // text alignment
+        protected Color color { get; set; }           // text color
+        protected int size { get; set; }              // text size
+        protected Font font { get; set; }             // text font
+        protected FontStyle style { get; set; }       // text font style
+        protected Material material { get; set; }     // text material
+        protected GameObject textObject { get; set; } = null;          // Our game object
+        protected uGUI_TextFade textFade { get; set; } = null;         // Our text fader
+        protected Text textText { get; set; } = null;                  // Our text object
+        protected ContentSizeFitter textFitter { get; set; } = null;   // Our content size fitter
```
PrimeSonic, line 431: Is there expectation that we will be extending/inheriting the `BasicText` class?

If not, these could be made `internal`

SMLHelper/Utility/BasicText.cs
```diff
@@ -0,0 +1,435 @@
...
+
+        static int index = 0; // For giving unique names to the game objects
```
PrimeSonic, line 433: If this is to be hidden from public use, make it `internal` explicitly.

SMLHelper/Utility/BasicText.cs
```diff
@@ -0,0 +1,435 @@
...
+ * ... // other things happen, time goes by
+ * message.Hide(); // Hides the message
+ */
+using System;
+using UnityEngine;
+using UnityEngine.UI;
+
+namespace SMLHelper.V2.Utility
+{
+    public class BasicText
```
PrimeSonic, line 32: TODO: Add XML documentation for Visual Studio.
Simply start typing three forward slashes `///` above any member definition and Visual Studio should auto-complete all the necessary XML tags.

SMLHelper/Utility/BasicText.cs
Cattlesquat: Made them internal

SMLHelper/Utility/BasicText.cs
Cattlesquat: Made it internal

SMLHelper/Utility/BasicText.cs
Cattlesquat: I tried what you say in Visual Studio, but nothing happened. I must not have something set up right, but Visual Studio is of course a maze of options.

PrimeSonic: I forgot that's actually not built-in.
Here's the tool that enables that.
https://submain.com/download/ghostdoc/community/

Cattlesquat: image
I have Visual Studio 2019 :O

PR #203, opened 2021-02-05
Button Patcher Removal

MrPurple6411:

  - Removed the Extra button patcher and implimented Zorgesho's fix.


SMLHelper/Patchers/OptionsPanelPatcher.cs
```diff
@@ -51,6 +52,18 @@ internal static void AddTab_Postfix(string label, int __result)
                 modsTabIndex = __result;
         }
 
+        [PatchUtils.Prefix]
+        [HarmonyPatch(typeof(uGUI_Binding), nameof(uGUI_Binding.RefreshValue))]
+        internal static bool RefreshValue_Prefix(uGUI_Binding __instance)
```
zorgesho, line 57: This patch should probably be for BZ only. Idk if this can cause any trouble in SN, but since it's needed just for BZ, I think this patch and also `ModBindingTag` should be for BZ only. Other than that, looks good to me.

MrPurple6411: e357faf

zorgesho (approved):

PR #205, opened 2021-02-05
BasicText - Documentation and BZ Support

PrimeSonic:
- Adds all missing XML documentation and adds a new `IBasicText` interface for future use
- Fixed all code style naming violations
- Added support for Below Zero using `TMPro`
- Added missing Unity DLLs to Dependencies


Cattlesquat (approved):
Certainly looks good to me! When I wrote this my recent code development had been in Java and so I was using those conventions -- still picking up the c# styling differences. Glad you had a BZ solution too. I hope folks will get some use out of this.

PR #208, opened 2021-02-21
Survival handler

Metious:
added a simple handler with ~~4~~ 2 methods overall that make the item give oxygen on consume, or heals on consume.

~~i had to make 2 methods for each feature because there are two actions(eat/use) in the game and they get called independently.~~

### DIFF
--- a/SMLHelper/Patchers/SurvivalPatcher.cs
+++ b/SMLHelper/Patchers/SurvivalPatcher.cs
@@ -50,7 +50,8 @@ namespace SMLHelper.V2.Patchers
                     result = true;
                 }
             }
-            FMODUWE.PlayOneShot(CraftData.GetUseEatSound(tt), Player.main.transform.position);
+            if (result)
+                FMODUWE.PlayOneShot(CraftData.GetUseEatSound(tt), Player.main.transform.position); // only play the sound if its useable
         }
     }
 }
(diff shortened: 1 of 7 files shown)


SMLHelper/Patchers/SurvivalPatcher.cs
```diff
@@ -0,0 +1,47 @@
...
+        internal static IDictionary<TechType, List<Action>> SurvivalDictionaryOnEat = new SelfCheckingDictionary<TechType, List<Action>>("SurvivalDictionaryOnEat", TechTypeExtensions.sTechTypeComparer);
+        internal static List<TechType> InventoryUseables = new List<TechType>();
+
+        internal static void Patch(Harmony harmony)
+        {
+            harmony.Patch(AccessTools.Method(typeof(Survival), nameof(Survival.Use)),
+                postfix: new HarmonyMethod(typeof(SurvivalPatcher), nameof(UsePostfix)));
+
+            harmony.Patch(AccessTools.Method(typeof(Survival), nameof(Survival.Eat)),
+                postfix: new HarmonyMethod(typeof(SurvivalPatcher), nameof(EatPostfix)));
```
PrimeSonic, line 22: Both of these methods have the same signature.  
`UsePostFix` and `EatPostFix` could be combine into a single method and re-used for both of these postfix patches.

PrimeSonic (approved):
I'm good with this

AlexejheroYTB (approved):

PR #211, opened 2021-04-28
TechTypeHandler now gets the mod's name by passing the Assembly instead of a string

Metious:

  - ITechTypeHandlerInternal interface's AddTechType methods now need an Assembly instead of a string for the ModName.
  - Added 2 new methods to the TechTypeHandler for manual ModAssembly passing.
  - Adjusted the Assets classes TechType creation to fit the new system.
  
this changes will hopefully fix the issues for mod Libraries that encounter into SMLHelper giving them the wrong mod name for the TechType made by their Library for other mods.

### DIFF
--- a/SMLHelper/Handlers/TechTypeHandler.cs
+++ b/SMLHelper/Handlers/TechTypeHandler.cs
@@ -28,20 +28,20 @@
             // Hide constructor
         }
 
-        TechType ITechTypeHandlerInternal.AddTechType(Assembly assembly, string internalName, string displayName, string tooltip)
+        TechType ITechTypeHandlerInternal.AddTechType(Assembly modAssembly, string internalName, string displayName, string tooltip)
         {
-            return Singleton.AddTechType(assembly, internalName, displayName, tooltip, true);
+            return Singleton.AddTechType(modAssembly, internalName, displayName, tooltip, true);
         }
 
-        TechType ITechTypeHandlerInternal.AddTechType(Assembly assembly, string internalName, string displayName, string tooltip, bool unlockAtStart)
+        TechType ITechTypeHandlerInternal.AddTechType(Assembly modAssembly, string internalName, string displayName, string tooltip, bool unlockAtStart)
         {
-            string modName = assembly.GetName().Name;
+            string modName = modAssembly.GetName().Name;
             
             // Register the TechType.
             var techType = TechTypePatcher.AddTechType(internalName);
             
             // Remember which Assembly added it
-            TechTypesAddedBy[techType] = assembly;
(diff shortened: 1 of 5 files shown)


MrPurple6411 (approved):

SMLHelper/Handlers/TechTypeHandler.cs
```diff
@@ -60,7 +61,30 @@ TechType ITechTypeHandlerInternal.AddTechType(string modName, string internalNam
...
+        /// <summary>
+        /// Adds a new <see cref="TechType"/> into the game.
+        /// </summary>
+        /// <param name="addedByAssembly">The assembly this TechType is getting added by.</param>
+        /// <param name="internalName">The internal name of the TechType. Should not contain special characters</param>
+        /// <param name="displayName">The display name of the TechType. Can be anything.</param>
+        /// <param name="tooltip">The tooltip, displayed when hovered in an inventory. Can be anything.</param>
+        /// <param name="unlockAtStart">Whether this TechType should be unlocked on game start, or not. By default, <see langword="true"/>.</param>
+        /// <returns>The new <see cref="TechType"/> that is created.</returns>
+        public static TechType AddTechType(Assembly addedByAssembly, string internalName, string displayName, string tooltip, bool unlockAtStart) =>
```
PrimeSonic, line 83: So how is this going to solve the issue?  
Because it's not "wrong". SMLHelper is recording which assembly made the request to it. And if that assembly was a library and not a final mod, that's just what happened.  
It's why all items created by CustomBatteries just report that mod as their origin.

Does this require other mods to pass in the assembly of the calling mod? If so, you've only "fixed" this for the libraries where the author knows to do that.

Metious: SMLHelper is indeed just doing what it has been told to do and there's nothing to blame it about, but before there was no way to "fix" this kind of issues for library mods and now there is.

And pretty much the only other libraries that messes with techtype adding are ECCLibrary and the upcoming ArchitectsLibrary - both are actively being developed on and aware of this issue.

Libraries do indeed need to pass the assembly manually if they encounter to this issue and normally this issue only happens if the techtype wasnt made by a Spawnable class or another mod's assets class that inherits from SMLHelper ones, its mostly just for extra techtypes made manually with TechTypeHandler.AddTechType().

If u have any better workaround on top of ur head, let me know :)

PrimeSonic: I find it surprising that we still have to manually invoke `AddTechType` at all.  
But that aside, as long as this is a _new_ method being added and the original (public) ones are still in place and with their original signatures, then we're good.

PrimeSonic (approved):

PR #215, opened 2021-05-09
SMLHelper 2.9.6 Both Games all 4 Branches

MrPurple6411:
### Changes made in this pull request

  - TechType Assembly name handling #211 
  - Coordinated Spawn system #212 
  - Fixed Caches not checking in Deactivated cache #213 
  - TechGroup/TechCategory Handlers #214 
  - Added Virtual bool HasSprite check to Spawnables for when not overriding GetItemSprite
  - Added a new Enum Patcher for BackgroundType. #216 
  - Added a new Handler for adding BackgroundTypes. #216 
  - fixed an issue with the FishPrefab.cs that causes SMLHelper fail to build for the Subnautica Experimental branch. #216 
  - Setup an easy way to add Background sprites and use them as a background for any item. #216 
  - BasicText Fixes #217 

[SMLHelper_BZ.STABLE.zip](https://github.com/SubnauticaModding/SMLHelper/files/6475650/SMLHelper_BZ.STABLE.zip)
[SMLHelper_BZ.EXP.zip](https://github.com/SubnauticaModding/SMLHelper/files/6475651/SMLHelper_BZ.EXP.zip)
[SMLHelper_SN.STABLE.zip](https://github.com/SubnauticaModding/SMLHelper/files/6475652/SMLHelper_SN.STABLE.zip)
[SMLHelper_SN.EXP.zip](https://github.com/SubnauticaModding/SMLHelper/files/6475653/SMLHelper_SN.EXP.zip)

PR #217, opened 2021-05-12
BasicText Fixes

Metious:

  - Fixed couple NREs in the BasicText.cs and added some null checks
  - Fixed an error where multiple Text components gets added to an object which causes SMLHelper to fail.


PrimeSonic (approved):

tobeyStraitjacket (approved):

tobeyStraitjacket:
Recommend merging this into dev ahead of merging #215

PR #219, opened 2021-06-07
Purge FishHandler

MrPurple6411:

  - Delete Outdated Fish system


tobeyStraitjacket (approved):
Calls for a major version bump (3.0).

Suggest investigating the possibility of merging ECCLib into SMLHelper as a separate PR to replace the deprecated features of the API this PR removes.

MrPurple6411:
> Calls for a major version bump (3.0).
> 
> Suggest investigating the possibility of merging ECCLib into SMLHelper as a separate PR to replace the deprecated features of the API this PR removes.

Looks like the next version is going to be 3.0.0 as I cannot justify just doing a small bump for adding the sound handler and 2.10.0 seems odd to me.

zorgesho:
> Looks like the next version is going to be 3.0.0 as I cannot justify just doing a small bump for adding the sound handler and 2.10.0 seems odd to me.

If FishHandler is used somewhere, then it shouldn't be removed. If it's not used anywhere, then there is no reason for major version bump, there is nothing wrong with v2.10.0.

tobeyStraitjacket:
> > Calls for a major version bump (3.0).
> > Suggest investigating the possibility of merging ECCLib into SMLHelper as a separate PR to replace the deprecated features of the API this PR removes.
> 
> Looks like the next version is going to be 3.0.0 as I cannot justify just doing a small bump for adding the sound handler and 2.10.0 seems odd to me.

Adding the sound handler is adding a new backwards compatible feature to the API.
[Semantic versioning](https://semver.org/#spec-item-7) states:

> Minor version Y (x.Y.z | x > 0) MUST be incremented if new, backwards compatible functionality is introduced to the public API.

2.10.0 is what comes after 2.9.x.x if you perform a minor version bump. It's a version, not a decimal.

Bumping from 2.9.0 to 3.0.0 for the sound handler just because you don't like 2.10.0 is bad practice.

tobeyStraitjacket:
> > Looks like the next version is going to be 3.0.0 as I cannot justify just doing a small bump for adding the sound handler and 2.10.0 seems odd to me.
> 
> If FishHandler is used somewhere, then it shouldn't be removed. If it's not used anywhere, then there is no reason for major version bump, there is nothing wrong with v2.10.0.

From talks it seems the fishhandler isn't used anywhere, removing it from the API is maintenance as it has been largely phased out in favour of ECCLib, nobody has used this API in a long time and it doesn't work. So, removing it is good maintenance IMO, but definitely calls for a major bump.

zorgesho:
> but definitely calls for a major bump

Well, technically you're right, but that FishHandler was used maybe in one old mod and it was more like a test mod than a real mod. I would say, we can bend the rules of semantic versioning a little and remove it without major bump or we can mark it as obsolete and remove it later, along with more significant changes.

tobeyStraitjacket:
> > but definitely calls for a major bump
> 
> remove it without major bump or we can mark it as obsolete and remove it later,

Yeah, upon further consideration I have been leaning towards this as well. I feel like it might be wisest to mark it all as obsolete for now so that no new code can be compiled that uses it, and then when a larger change comes to the API is the time to cut it completely. However, as it has never been used by any public-facing mods, I wouldn't be completely against removing it without a major bump.

PR #221, opened 2021-06-12
CoordinatedSpawns System Tweaks

Metious:

  - Altered the Coordinated Spawns list to use a record for optional rotations.
  - Added a new method that determines the rotation of the spawns for a single TechType.
  - Overhauled the Coordinated Spawns Save data to work on old saves and better handle newly added mods.
  - Set Save data to only save when game is saved.

### DIFF
--- a/SMLHelper/Assets/Spawnable.cs
+++ b/SMLHelper/Assets/Spawnable.cs
@@ -72,17 +72,17 @@
         public virtual Vector2int SizeInInventory { get; } = defaultSize;
         
         /// <summary>
-        /// Returns the list of <see cref="Vector3"/>s that specify the prefab's Coordinated Spawns.<br/>
+        /// A lightweight class used to specify the position of a Coordinated Spawn and optionally set its rotation.
+        /// </summary>
+        /// <param name="position"></param>
+        /// <param name="eulerAngles"></param>
+        public record SpawnLocation(Vector3 position, Vector3 eulerAngles = default);
+        
+        /// <summary>
+        /// Returns the list of <see cref="SpawnLocation"/>s that specify the prefab's Coordinated Spawns.<br/>
         /// /// By default this will be null.
         /// </summary>
-        public virtual List<Vector3> CoordinatedSpawns { get; } = null;
-
-        /// <summary>
-        /// Returns the Dictionary of (Vector3, Vector3) that specify the prefab's Coordinated Spawns as well as the rotations it will spawn as.<br/>
-        /// the Keys of the Dictionary will be counted as the spawns location and the Values of the Dictionary will be counted as the rotations of that spawn.<br/>
-        /// By default this will be null.
-        /// </summary>
-        public virtual Dictionary<Vector3, Vector3> RotatedCoordinatedSpawns { get; } = null;
+        public virtual List<SpawnLocation> CoordinatedSpawns { get; } = null;
 
         /// <summary>
         /// Returns the List of BiomeData that handles what Biomes this prefab will spawn, how probable it is to spawn there and how many per spawn.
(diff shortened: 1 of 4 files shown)


SMLHelper/Assets/Spawnable.cs
```diff
@@ -74,10 +73,17 @@ public abstract class Spawnable: ModPrefab
...
+        /// /// By default this will be null.
         /// </summary>
         public virtual List<Vector3> CoordinatedSpawns { get; } = null;
 
+        /// <summary>
+        /// Returns the Dictionary of (Vector3, Vector3) that specify the prefab's Coordinated Spawns as well as the rotations it will spawn as.<br/>
+        /// the Keys of the Dictionary will be counted as the spawns location and the Values of the Dictionary will be counted as the rotations of that spawn.<br/>
+        /// By default this will be null.
+        /// </summary>
+        public virtual Dictionary<Vector3, Vector3> RotatedCoordinatedSpawns { get; } = null;
```
zorgesho, line 85: There is no need for backward compatibility if it's not used anywhere yet.
Just use single list of records (C# 9.0) for both `CoordinatedSpawns` and `RotatedCoordinatedSpawns`, something like this:
```cs
public record PositionRotation(Vector3 position, Vector3 eulerAngles = default);

public virtual List<PositionRotation> CoordinatedSpawns { get; } = null;
```

tobeyStraitjacket: > Just use single list of records (C# 9.0) for both `CoordinatedSpawns` and `RotatedCoordinatedSpawns`, something like this:

AFAIK, we cannot use C#9 language features yet. No version of Unity exists that currently supports them. I would personally really love to adopt records for Unity mods.

zorgesho: `record` is just syntactic sugar, it doesn't need any support from .NET or Unity.

tobeyStraitjacket: > `record` is just syntactic sugar, it doesn't need any support from .NET or Unity.

Ah nice. In that case, I'm all for it!

tobeyStraitjacket: > `record` is just syntactic sugar, it doesn't need any support from .NET or Unity.

I did some investigation: `record` requires the `IsExternalInit` class to exist in `System.Runtime.CompilerServices` namespace, which it does not when targeting .NET 4.7.2 or earlier. It _is_ possible to define this yourself or to reference for example a NuGet package that defines it as a polyfill, but it is not _quite_ as simple as setting LangVersion = 9.0 in your .csproj. Therefore, I'm not sure suggesting `record` for this use is the best idea.

zorgesho: Huh, I didn't know that was required for records too, I did it for`init` property modifiers ([here](https://github.com/zorgesho/SubnauticaMods/blob/master/Common/Misc.cs#L5-L6))
Anyway, that line is also works with records.

tobeyStraitjacket: > Huh, I didn't know that was required for records too
By default, `record` properties are `public { get; init; }`. It's one of the best things about them!

> Anyway, that line is also works with records.
Yes, that line works. I personally would prefer to use this NuGet package: https://www.nuget.org/packages/IsExternalInit/

However, if we wanted to add records and C# 9.0 `LangVersion` to SMLHelper, that should really be its own PR, rather than forcing this PR to bundle it in for an unrelated change, shouldn't it?

zorgesho: > that should really be its own PR

If it's more than changing `LangVersion` then probably yes. That package seems like an overkill though.

MrPurple6411: So...... in the end ??????? No?

MrPurple6411: 0964fb8

zorgesho (approved):

PR #226, opened 2021-06-24
Coordinated spawns bug fixes

Metious:

  - Fixed a bug where non-global coordinated spawns objects were getting despawned after save/reload.
  - spawns now get cached when the object is completely spawned in the current save.
  - small doc fix.

### DIFF
--- a/SMLHelper/MonoBehaviours/EntitySpawner.cs
+++ b/SMLHelper/MonoBehaviours/EntitySpawner.cs
@@ -26,13 +26,29 @@ namespace SMLHelper.V2.MonoBehaviours
             var task = new TaskResult<GameObject>();
             yield return GetPrefabAsync(task);
 
-            if (task.Get() is null)
+            var prefab = task.Get();
+
+            if (prefab == null)
             {
                 Logger.Error($"no prefab found for {stringToLog}; process for Coordinated Spawn canceled.");
                 Destroy(gameObject);
             }
 
-            var obj = Utils.InstantiateDeactivated(task.Get(), spawnInfo.spawnPosition, spawnInfo.rotation);
+            var lwe = prefab.GetComponent<LargeWorldEntity>();
+
+            // non-global objects cannot be spawned in unloaded terrain so we need to wait
+            if (lwe.cellLevel is not (LargeWorldEntity.CellLevel.Batch or LargeWorldEntity.CellLevel.Global))
+            {
+                var lws = LargeWorldStreamer.main;
+                while (!lws.IsReady()) // first we wait for the world streamer to be initialized
+                {
+                    yield return new WaitForSecondsRealtime(2f);
+                }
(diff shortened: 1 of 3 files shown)


SMLHelper/Patchers/LargeWorldStreamerPatcher.cs
```diff
@@ -26,7 +26,7 @@ internal static void Patch(Harmony harmony)
         }
         
         internal static readonly List<SpawnInfo> spawnInfos = new List<SpawnInfo>();
-        private static List<SpawnInfo> savedSpawnInfos = new List<SpawnInfo>();
+        internal static List<SpawnInfo> savedSpawnInfos = new List<SpawnInfo>();
```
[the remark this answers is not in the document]
Metious: https://github.com/SubnauticaModding/SMLHelper/pull/226/commits/ab6a387597327fc25e8e6f30524e6c536848b05e

SMLHelper/MonoBehaviours/EntitySpawner.cs
```diff
@@ -24,24 +26,43 @@ IEnumerator SpawnAsync()
...
+            var lwe = obj.GetComponent<LargeWorldEntity>();
+
+            // non-global objects cannot be spawned in unloaded terrain so we need to wait
+            if (lwe == null || lwe.cellLevel is not (LargeWorldEntity.CellLevel.Batch or LargeWorldEntity.CellLevel.Global))
+            {
+                var lws = LargeWorldStreamer.main;
+                while (!lws.IsReady()) // first we make sure the world streamer is initialized
+                {
+                    yield return new WaitForSecondsRealtime(2f);
+                }
```
tobeyStraitjacket, line 49: Is there something we can wait for that isn't just an arbitrary 2 second timer in a while loop? Would `new WaitUntil(() => lws.IsReady()` not be more suitable, since it seems like you're just trying to wait until it is ready?

```suggestion
                yield return new WaitUntil(() => lws.IsReady()); // first we make sure the world streamer is initialized
```

tobeyStraitjacket (changes requested):

SMLHelper/MonoBehaviours/EntitySpawner.cs
Metious: Right, i wanted to use a `WaitUntil` for that one as well, but i thought checking for it on each frame wasn't necessary so i did that 2 seconds thing.

tobeyStraitjacket: How heavy is the `IsReady()` method call? If not very heavy, it's probably worthwhile. An arbitrary 2 second wait seems fragile to me.

Metious: yeah, it's not very heavy, it just checks for an `init` boolean field somewhere else, we can change it.

tobeyStraitjacket (approved):

PR #229, opened 2021-07-05
Coordinated spawns hotfixes

Metious:
## Changes made in this pull request

  - Fixed a bug where the savedSpawns Deserializer was failing.
  - Added two new `public` `JsonConverter` classes (`QuaternionConverter`/`Vector3Converter`) for simplified serializations.
  - Redesigned `SpawnInfo` into a `struct`, implementing `IEquatable<SpawnInfo>`, custom `Equals`, `GetHashCode` methods and operator overloads for `==` and `!=` respectively.
  - Fixes an issue where `EntitySpawner.SpawnAsync` was using `TechType.ToString()` rather than `TechType.AsString()`.

### 5555948 v2.10.0.3: `Equals(object obj)` & `GetHashCode()`
- This build features an implementation as initially intended by @Metious which overrides `Equals(object obj)` & `GetHashCode()` to provide a value-based equality comparison for all properties of `SpawnInfo`.
- As it overrides `Equals` and `GetHashCode` to provide value-based equality comparison, this is a breaking change for anyone who previously relied on `Equals` or `GetHashCode` (particularly a problem in the case of `HashSet`), but it solves the problem where `List.Contains` is returning `false` for deserialized `SpawnInfo` instances. However, it also means that we cannot later change the implementation of `Equals` or `GetHashCode` without another breaking change for consumers.

#### Subnautica
[Stable v2.10.0.3](https://github.com/SubnauticaModding/SMLHelper/files/6768155/SMLHelper_SN.STABLE_v2.10.0.3.zip)
[Experimental v2.10.0.3](https://github.com/SubnauticaModding/SMLHelper/files/6768157/SMLHelper_SN.EXP_v2.10.0.3.zip)

#### Below Zero
[Stable v2.10.0.3](https://github.com/SubnauticaModding/SMLHelper/files/6768159/SMLHelper_BZ.STABLE_v2.10.0.3.zip)
[Experimental v2.10.0.3](https://github.com/SubnauticaModding/SMLHelper/files/6768161/SMLHelper_BZ.EXP_v2.10.0.3.zip)

### 96cdaa3 v2.10.0.5: `IEquatable<SpawnInfo>` `Equals(SpawnInfo other)`
- This build does not override `Equals(object obj)` or `GetHashCode`, but instead implements `IEquatable<SpawnInfo>` and `Equals(SpawnInfo other)` to provide a value-based equality comparison for all properties of `SpawnInfo`.
- `List.Contains` should defer to`Equals(SpawnInfo)` for comparison since we implement `IEquatable<SpawnInfo>`, and the issue @Metious was trying to solve should be resolved.
- This should make for a less egregious backwards-incompatible change, as `Equals(object obj)` and `GetHashCode` still operate the same, it is only `Equals(SpawnInfo other)` that has been changed. Still, the issue remains that if we change this in future, it will have implications for consumers, so it must be approached with caution.

#### Subnautica
[Stable v2.10.0.5](https://github.com/SubnauticaModding/SMLHelper/files/6768741/SMLHelper_SN.STABLE_v2.10.0.5.zip)
[Experimental v2.10.0.5](https://github.com/SubnauticaModding/SMLHelper/files/6768745/SMLHelper_SN.EXP_v2.10.0.5.zip)

#### Below Zero
[Stable v2.10.0.5](https://github.com/SubnauticaModding/SMLHelper/files/6768747/SMLHelper_BZ.STABLE_v2.10.0.5.zip)
[Experimental v2.10.0.5](https://github.com/SubnauticaModding/SMLHelper/files/6768749/SMLHelper_BZ.EXP_v2.10.0.5.zip)

### b9802ce v2.10.0.6: `SpawnInfo` redesigned as a `struct`
- This build combines the logic of both 2.10.0.3 and 2.10.0.5, and on top of that, also redesigns `SpawnInfo` as a `struct` rather than a `class`.
- @Metious assures me that nobody outside of himself and the RotA team are making any use of the `SpawnInfo` API, so a backwards-incompatible change that fixes the design of the API is preferable, should this prove to be true.

### Subnautica
[Stable v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770222/SMLHelper_SN.STABLE_v2.10.0.6.zip)
[Experimental v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770224/SMLHelper_SN.EXP_v2.10.0.6.zip)

### Below Zero
[Stable v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770230/SMLHelper_BZ.STABLE_v2.10.0.6.zip)
[Experimental v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770231/SMLHelper_BZ.EXP_v2.10.0.6.zip)

## Builds by @Metious ca7018a
[SMLHelper_BZ.STABLE.zip](https://github.com/SubnauticaModding/SMLHelper/files/6765975/SMLHelper_BZ.STABLE.zip)
[SMLHelper_BZ.STABLE.zip](https://github.com/SubnauticaModding/SMLHelper/files/6765975/SMLHelper_BZ.STABLE.zip)

### DIFF
--- a/SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
+++ b/SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
@@ -128,15 +128,15 @@ namespace SMLHelper.V2.Handlers
     public class SpawnInfo : IEquatable<SpawnInfo>
     {
         [JsonProperty]
-        internal TechType techType { get; }
+        internal TechType techType { get; init; }
         [JsonProperty]
-        internal string classId { get; }
+        internal string classId { get; init; }
         [JsonProperty]
-        internal Vector3 spawnPosition { get; }
+        internal Vector3 spawnPosition { get; init; }
         [JsonProperty]
-        internal Quaternion rotation { get; }
+        internal Quaternion rotation { get; init; }
         [JsonProperty]
-        internal SpawnType spawnType { get; }
+        internal SpawnType spawnType { get; init; }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -144,12 +144,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="techType">TechType to spawn.</param>
         /// <param name="spawnPosition">Position to spawn into.</param>
         public SpawnInfo(TechType techType, Vector3 spawnPosition)
-        {
-            this.techType = techType;
-            this.spawnPosition = spawnPosition;
-            this.rotation = Quaternion.identity;
-            spawnType = SpawnType.TechType;
-        }
+            : this(default, techType, spawnPosition, Quaternion.identity) { }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -157,12 +152,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="classId">ClassID to spawn.</param>
         /// <param name="spawnPosition">Position to spawn into.</param>
         public SpawnInfo(string classId, Vector3 spawnPosition)
-        {
-            this.classId = classId;
-            this.spawnPosition = spawnPosition;
-            this.rotation = Quaternion.identity;
-            spawnType = SpawnType.ClassId;
-        }
+            : this(classId, default, spawnPosition, Quaternion.identity) { }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -171,12 +161,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="spawnPosition">Position to spawn into.</param>
         /// <param name="rotation">Rotation to spawn at.</param>
         public SpawnInfo(TechType techType, Vector3 spawnPosition, Quaternion rotation)
-        {
-            this.techType = techType;
-            this.spawnPosition = spawnPosition;
-            this.rotation = rotation;
-            spawnType = SpawnType.TechType;
-        }
+            : this(default, techType, spawnPosition, rotation) { }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -185,12 +170,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="spawnPosition">Position to spawn into.</param>
         /// <param name="rotation">Rotation to spawn at.</param>
         public SpawnInfo(string classId, Vector3 spawnPosition, Quaternion rotation)
-        {
-            this.classId = classId;
-            this.spawnPosition = spawnPosition;
-            this.rotation = rotation;
-            spawnType = SpawnType.ClassId;
-        }
+            : this(classId, default, spawnPosition, rotation) { }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -199,12 +179,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="spawnPosition">Position to spawn into.</param>
         /// <param name="rotation">Rotation to spawn at.</param>
         public SpawnInfo(TechType techType, Vector3 spawnPosition, Vector3 rotation)
-        {
-            this.techType = techType;
-            this.spawnPosition = spawnPosition;
-            this.rotation = Quaternion.Euler(rotation);
-            spawnType = SpawnType.TechType;
-        }
+            : this(default, techType, spawnPosition, Quaternion.Euler(rotation)) { }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -213,12 +188,7 @@ namespace SMLHelper.V2.Handlers
         /// <param name="spawnPosition">Position to spawn into.</param>
         /// <param name="rotation">Rotation to spawn at.</param>
         public SpawnInfo(string classId, Vector3 spawnPosition, Vector3 rotation)
-        {
-            this.classId = classId;
-            this.spawnPosition = spawnPosition;
-            this.rotation = Quaternion.Euler(rotation);
-            spawnType = SpawnType.ClassId;
-        }
+            : this(classId, default, spawnPosition, Quaternion.Euler(rotation)) { }
 
         [JsonConstructor]
         internal SpawnInfo(string classId, TechType techType, Vector3 spawnPosition, Quaternion rotation)
@@ -227,7 +197,11 @@ namespace SMLHelper.V2.Handlers
             this.techType = techType;
             this.spawnPosition = spawnPosition;
             this.rotation = rotation;
-            spawnType = this.techType == TechType.None ? SpawnType.ClassId : SpawnType.TechType;
+            spawnType = this.techType switch
+            {
+                TechType.None => SpawnType.ClassId,
+                _ => SpawnType.TechType
+            };
         }
 
         /// <summary>
--- a/SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
+++ b/SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
@@ -123,20 +123,20 @@ namespace SMLHelper.V2.Handlers
 
     #region SpawnInfo
     /// <summary>
-    /// a basic class that provides enough info for the <see cref="CoordinatedSpawnsHandler"/> System to function.
+    /// A basic struct that provides enough info for the <see cref="CoordinatedSpawnsHandler"/> System to function.
     /// </summary>
-    public class SpawnInfo : IEquatable<SpawnInfo>
+    public struct SpawnInfo : IEquatable<SpawnInfo>
     {
         [JsonProperty]
-        internal TechType techType { get; init; }
+        internal TechType TechType { get; }
         [JsonProperty]
-        internal string classId { get; init; }
+        internal string ClassId { get; }
         [JsonProperty]
-        internal Vector3 spawnPosition { get; init; }
+        internal Vector3 SpawnPosition { get; }
         [JsonProperty]
-        internal Quaternion rotation { get; init; }
+        internal Quaternion Rotation { get; }
         [JsonProperty]
-        internal SpawnType spawnType { get; init; }
+        internal SpawnType Type { get; }
 
         /// <summary>
         /// Initializes a new <see cref="SpawnInfo"/>.
@@ -193,11 +193,11 @@ namespace SMLHelper.V2.Handlers
         [JsonConstructor]
         internal SpawnInfo(string classId, TechType techType, Vector3 spawnPosition, Quaternion rotation)
         {
-            this.classId = classId;
-            this.techType = techType;
-            this.spawnPosition = spawnPosition;
-            this.rotation = rotation;
-            spawnType = this.techType switch
+            ClassId = classId;
+            TechType = techType;
+            SpawnPosition = spawnPosition;
+            Rotation = rotation;
+            Type = TechType switch
             {
                 default(TechType) => SpawnType.ClassId,
                 _ => SpawnType.TechType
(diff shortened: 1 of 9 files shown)


Metious:
Calling for a fast merge/publish as this bug was game breaking.

SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
```diff
@@ -220,6 +220,40 @@ public SpawnInfo(string classId, Vector3 spawnPosition, Vector3 rotation)
             spawnType = SpawnType.ClassId;
         }
 
+        [JsonConstructor]
+        internal SpawnInfo(string classId, TechType techType, Vector3 spawnPosition, Vector3 rotation)
+        {
+            this.classId = classId;
+            this.techType = techType;
+            this.spawnPosition = spawnPosition;
+            this.rotation = Quaternion.Euler(rotation);
```
tobeyStraitjacket, line 229:
```suggestion
            this.rotation = rotation;
```

SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
```diff
@@ -220,6 +220,40 @@ public SpawnInfo(string classId, Vector3 spawnPosition, Vector3 rotation)
             spawnType = SpawnType.ClassId;
         }
 
+        [JsonConstructor]
+        internal SpawnInfo(string classId, TechType techType, Vector3 spawnPosition, Vector3 rotation)
```
tobeyStraitjacket, line 224: The `rotation` is stored as a `Quaternion`, so this appears to be a mistake.
```suggestion
        internal SpawnInfo(string classId, TechType techType, Vector3 spawnPosition, Quaternion rotation)
```

SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
```diff
@@ -220,6 +220,40 @@ public SpawnInfo(string classId, Vector3 spawnPosition, Vector3 rotation)
...
+
+        /// <summary>
+        /// Serves as the default hash function.
+        /// </summary>
+        /// <returns>A hash code for the current object.</returns>
+        public override int GetHashCode()
+        {
+            // this was required for C# to shut up
+            return base.GetHashCode();
+        }
```
tobeyStraitjacket, line 255: Your `Equals` and `GetHashCode` need some work. I'm concerned that `spawnInfo.rotation == this.rotation` may now in some cases through false negatives, but this might be an acceptable caveat.
However, the `GetHashCode` definitely needs an implementation that reflects the new `Equals` method.
I'll see what I can do on this.

SMLHelper/Json/Converters/QuaternionConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        /// <summary>
+        /// A method that tells Newtonsoft how to Serialize the current object.
+        /// </summary>
+        /// <param name="writer"></param>
+        /// <param name="value"></param>
+        /// <param name="serializer"></param>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var quaternion = (Quaternion)value;
```
tobeyStraitjacket, line 33: Consider whether it might be appropriate to use `as` here instead of a traditional cast.
```suggestion
            var quaternion = value as Quaternion;
```

SMLHelper/Json/Converters/QuaternionConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+        /// A method that tells Newtonsoft how to Deserialize and read the current object.
+        /// </summary>
+        /// <param name="reader"></param>
+        /// <param name="objectType"></param>
+        /// <param name="existingValue"></param>
+        /// <param name="serializer"></param>
+        /// <returns></returns>
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var q = (QuaternionJson)serializer.Deserialize(reader, typeof(QuaternionJson));
```
tobeyStraitjacket, line 48: Prefer the generic form of `JsonSerializer.Deserialize<T>`:
```suggestion
            var q = serializer.Deserialize<QuaternionJson>(reader);
```

SMLHelper/Json/Converters/Vector3Converter.cs
```diff
@@ -0,0 +1,67 @@
...
+
+        /// <summary>
+        /// A method that tells Newtonsoft how to Serialize the current object.
+        /// </summary>
+        /// <param name="writer"></param>
+        /// <param name="value"></param>
+        /// <param name="serializer"></param>
+        public override void WriteJson(JsonWriter writer, object value, JsonSerializer serializer)
+        {
+            var vector3 = (Vector3)value;
```
tobeyStraitjacket, line 33: Consider whether it might be appropriate to use `as` here instead of a traditional cast.
```suggestion
            var vector3 = value as Vector3;
```

SMLHelper/Json/Converters/Vector3Converter.cs
```diff
@@ -0,0 +1,67 @@
...
+        /// A method that tells Newtonsoft how to Deserialize and read the current object.
+        /// </summary>
+        /// <param name="reader"></param>
+        /// <param name="objectType"></param>
+        /// <param name="existingValue"></param>
+        /// <param name="serializer"></param>
+        /// <returns></returns>
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            var v = (Vector3Json)serializer.Deserialize(reader, typeof(Vector3Json));
```
tobeyStraitjacket, line 48: Prefer the generic form of `JsonSerializer.Deserialize<T>`:
```suggestion
            var v = serializer.Deserialize<Vector3Json>(reader);
```

SMLHelper/Json/Converters/QuaternionConverter.cs
```diff
@@ -0,0 +1,69 @@
...
+        internal float w;
+
+        public QuaternionJson(float x, float y, float z, float w)
+        {
+            this.x = x;
+            this.y = y;
+            this.z = z;
+            this.w = w;
+        }
+    }
```
tobeyStraitjacket, line 68: I believe a C# 9 `record` might be a better fit, here.
```suggestion
    record QuaternionJson(float x, float y, float z, float w);
```

SMLHelper/Json/Converters/Vector3Converter.cs
```diff
@@ -0,0 +1,67 @@
...
+        public float y;
+        public float z;
+
+        public Vector3Json(float x, float y, float z)
+        {
+            this.x = x;
+            this.y = y;
+            this.z = z;
+        }
+    }
```
tobeyStraitjacket, line 66: I believe a C# 9 record might be a better fit, here.
```suggestion
    record Vector3Json(float x, float y, float z);
```

SMLHelper/MonoBehaviours/EntitySpawner.cs
```diff
@@ -40,7 +40,7 @@ IEnumerator SpawnAsync()
             var lwe = obj.GetComponent<LargeWorldEntity>();
 
             // non-global objects cannot be spawned in unloaded terrain so we need to wait
-            if (lwe == null || lwe.cellLevel is not (LargeWorldEntity.CellLevel.Batch or LargeWorldEntity.CellLevel.Global))
+            if (lwe == null || lwe?.cellLevel is not (LargeWorldEntity.CellLevel.Batch or LargeWorldEntity.CellLevel.Global))
```
tobeyStraitjacket, line 43: This seems unnecessary - if the LHS evaluates to true (lwe == null), then the RHS will not be evaluated, so there will be no `NullReferenceException`.
```suggestion
            if (lwe == null || lwe.cellLevel is not (LargeWorldEntity.CellLevel.Batch or LargeWorldEntity.CellLevel.Global))
```

tobeyStraitjacket (changes requested):
Have suggested changes in comments, I also intend to work on `GetHashCode()` implementation before this goes through as I believe there will be issues with `SpawnInfo` if we introduce an `Equals` method that is not inline with the `GetHashCode` method - particularly if someone ever wants to make a `HashSet` of `SpawnInfo` instances.
Simply deferring to `base.GetHashCode()` when you have modified the logic of `Equals` will get the compiler to "shut up", but it will also cause problems down the road.

SMLHelper/Handlers/CoordinatedSpawnsHandler.cs
tobeyStraitjacket: As of commit baa7a47 I have implemented what I believe to be an appropriate `Equals` and `GetHashCode` combination for this class.

## Builds baa7a47
### Subnautica
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768011/SMLHelper_SN.STABLE_v2.10.0.2.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768012/SMLHelper_SN.EXP_v2.10.0.2.zip)

### Below Zero
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768014/SMLHelper_BZ.STABLE_v2.10.0.2.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768017/SMLHelper_BZ.EXP_v2.10.0.2.zip)

tobeyStraitjacket: ## Builds 5555948
### Subnautica
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768155/SMLHelper_SN.STABLE_v2.10.0.3.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768157/SMLHelper_SN.EXP_v2.10.0.3.zip)

### Below Zero
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768159/SMLHelper_BZ.STABLE_v2.10.0.3.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768161/SMLHelper_BZ.EXP_v2.10.0.3.zip)

tobeyStraitjacket: It is worth noting that once these `Equals(object b)` and `GetHashCode()` are added to the `public SpawnInfo` class, *any changes to these methods will be a breaking change*, as any consumers of the SMLHelper API using `SpawnInfo` in for example a `HashSet` will suddenly have unexpected results. The implications of this need to be considered very carefully before this can be released.

tobeyStraitjacket: As an alternative to overriding `GetHashCode()` and `Equals(object obj)`, I would propose that we instead implement `IEquatable<SpawnInfo>` instead, overriding `Equals(SpawnInfo other)`. This does not require us to implement `GetHashCode()`, and only changes the behaviour of `Equals(SpawnInfo other)` to be based on value equality, whereas `Equals(object obj)` will remain based on reference equality, as is the standard for a C# class. This will mean that `HashSet`s operate as you expect.

For the sake of value-based equality, other approaches I considered but rejected are:
- Convert `SpawnInfo` to a `struct`. This would be a breaking change, as it will change the fundamentals how `Equals` operates, similar to the original proposed change by @Metious.
- Convert `SpawnInfo` to a `record`. Unfortunately, it is not possible with a `record` to override `Equals`, and the default implementation will not give the same results as desired by @Metious, so this will not work, and also will be a breaking change.

Implementing `IEquatable<SpawnInfo>` is the least egregious version of these implementations, but only if it provides everything that is required to fix the bug where List.Contains() is returning false after serialization & deserialization. As such, I am attaching builds to this comment with this implementation for testing by @Metious and @LeeTwentyThree.

## Builds 7267bc4
### Subnautica
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768478/SMLHelper_SN.STABLE_v2.10.0.4.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768486/SMLHelper_SN.EXP_v2.10.0.4.zip)

### Below Zero
[Stable](https://github.com/SubnauticaModding/SMLHelper/files/6768489/SMLHelper_BZ.STABLE_v2.10.0.4.zip)
[Experimental](https://github.com/SubnauticaModding/SMLHelper/files/6768491/SMLHelper_BZ.EXP_v2.10.0.4.zip)

tobeyStraitjacket: Upon further discussion with @Metious, it seems that the `SpawnInfo` class is not in widespread use, so a redesign of the class is unlikely to have a big impact.

As such, converting it to a `struct` so that value-based equality is the expected default makes sense and is acceptable.

Attached is a build that mixes the best of v2.10.0.3 and v.2.10.0.5 builds, implementing the same `Equals(object obj)` and `GetHashCode` of v2.10.0.3, but also implementing `IEquatable<SpawnInfo>` and `Equals(SpawnInfo other)`, and on top of that, also implementing overloads for the `==` and `!=` operators, to keep everything consistent.

Since the `SpawnInfo` class has been changed to a `struct`, I also took the opportunity to fix the naming violations of its `internal` properties. Attached are builds for v2.10.0.6 as of commit b9802ce.

## Builds b9802ce
### Subnautica
[Stable v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770222/SMLHelper_SN.STABLE_v2.10.0.6.zip)
[Experimental v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770224/SMLHelper_SN.EXP_v2.10.0.6.zip)

### Below Zero
[Stable v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770230/SMLHelper_BZ.STABLE_v2.10.0.6.zip)
[Experimental v2.10.0.6](https://github.com/SubnauticaModding/SMLHelper/files/6770231/SMLHelper_BZ.EXP_v2.10.0.6.zip)

Metious:
Tested SMLHelper 2.10.6 and i can confirm that the potential bugs were fixed and everything works just fine.

tobeyStraitjacket (approved):
v2.10.0.6 appears to be bug free after testing from @Metious. Issues that were game-breaking are resolved, and no breaking changes for their 2.9.7 build of mods utilising the API.

PR #243, opened 2021-08-23
FMOD 2.01.09

Metious:
FMOD was updated in Subnautica experimental branch and FMOD stuff were moved out of the `Assembly-CSharp-firstpass.dll`.
Added the `FMODUnity.dll` to the reference list due to that reason.


PrimeSonic (approved):

kassandraberquez-bit:
r

PR #249, opened 2021-11-16
Add initial EquipmentType Enum Patcher

NeisesMike:

  - Adds functionality for registering and retrieving novel "EquipmentType" enumeration values
  - Follows the example of similar enumeration patchers
  - Works well when tested with VehicleFramework

### DIFF
--- a/SMLHelper/Handlers/EquipmentHandler.cs
+++ b/SMLHelper/Handlers/EquipmentHandler.cs
@@ -24,14 +24,12 @@
         /// Adds a new <see cref="Equipment" /> into the game.
         /// </summary>
         /// <param name="equipmentName">The name of the Equipment. Should not contain special characters.</param>
-        /// <param name="displayName">The display name of the Equipment. Can be anything.</param>
         /// <returns>
         /// The new <see cref="Equipment" /> that is created.
         /// </returns>
-        public EquipmentType RegisterNewEquipmentType(string equipmentName, string displayName)
+        public EquipmentType AddEquipmentType(string equipmentName)
         {
             EquipmentType equipment = EquipmentTypePatcher.AddEquipmentType(equipmentName);
-            LanguageHandler.SetLanguageLine(equipmentName, displayName);
             return equipment;
         }
 
@@ -45,14 +43,8 @@
         public bool ModdedEquipmentExists(string equipmentString)
         {
             EnumTypeCache cache = EquipmentTypePatcher.cacheManager.RequestCacheForTypeName(equipmentString, false);
-            if (cache != null) // Item Found
-            {
-                return true;
-            }
-            else // Mod not present or not yet loaded
-            {
-                return false;
-            }
+            // if we don't have it cached, the mod is not present or not yet loaded
+            return cache != null;
         }
 
         /// <summary>
--- a/SMLHelper/Handlers/EquipmentHandler.cs
+++ b/SMLHelper/Handlers/EquipmentHandler.cs
@@ -1,0 +1,82 @@
+﻿namespace SMLHelper.V2.Handlers
+{
+    using SMLHelper.V2.Handlers;
+    using SMLHelper.V2.Interfaces;
+    using SMLHelper.V2.Patchers.EnumPatching;
+    using SMLHelper.V2.Utility;
+
+    /// <summary>
+    /// A handler class for everything related to creating new Equipments.
+    /// </summary>
+    public class EquipmentHandler : IEquipmentHandler
+    {
+        /// <summary>
+        /// Main entry point for all calls to this handler.
+        /// </summary>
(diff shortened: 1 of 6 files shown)


SMLHelper/Handlers/EquipmentHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+        /// </summary>
+        /// <param name="equipmentName">The name of the Equipment. Should not contain special characters.</param>
+        /// <param name="displayName">The display name of the Equipment. Can be anything.</param>
+        /// <returns>
+        /// The new <see cref="Equipment" /> that is created.
+        /// </returns>
+        public EquipmentType RegisterNewEquipmentType(string equipmentName, string displayName)
+        {
+            EquipmentType equipment = EquipmentTypePatcher.AddEquipmentType(equipmentName);
+            LanguageHandler.SetLanguageLine(equipmentName, displayName);
```
Metious, line 34: Does the game use the display name anywhere? if not, then we dont need a `LanguageHandler.SetLanguageLine()` line.
while you're at it, rename the `RegisterNewEquipmentType` method to `AddEquipmentType` for consistency reasons.

Metious (changes requested):

Metious:
Enum patcher looks good to me.
Also, is there anyway we could add a few helper methods for EquipmentTypes? for instance, to add slots to the `Equipment.slotMapping` dictionary, i believe there's room to expand this a bit more.

SMLHelper/Handlers/EquipmentHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+        {
+            EnumTypeCache cache = EquipmentTypePatcher.cacheManager.RequestCacheForTypeName(equipmentString, false);
+            if (cache != null) // Item Found
+            {
+                return true;
+            }
+            else // Mod not present or not yet loaded
+            {
+                return false;
+            }
```
PrimeSonic, line 55: These types of blocks can be simplified
```suggestion
            // if we don't have it cached, the mod is not present or not yet loaded
            return cache != null;
```
This is really minor though so I wouldn't really worry about it.

PrimeSonic (commented):
Not bad.
Keep the method names consistent, as @Metious pointed out, and this can go right in.

NeisesMike:
I made all the requested changes.

Concerning adding more to the handler, I think that could be useful. Currently I'm relying on some magic names:

https://github.com/NeisesMike/SubnauticaAtramaVehicle/blob/5abbb0c88f3bb4479d5430b79cc75438e9d82180/VehicleFramework/VehicleFramework/Patches/EquipmentPatcher.cs#L30

All these patches basically just feed this dictionary to the Equipment class to ensure the slots are recognized correctly. I could replicate these patches for every vehicle, but if there's a way to accomplish this here, I'd jump on it. The only exception is the patch on IsCompatible, which should eventually ensure that Atrama modules can't be put into an Odyssey.

With this EquipmentType patching, I can replicate what I have in a much better way, but moving to distinct modules for each vehicle remains a challenge.

SMLHelper/Handlers/EquipmentHandler.cs
NeisesMike: No, I don't think it uses the display name. That was a little bit of greedy copy-pasting. I also amended the method name.

SMLHelper/Handlers/EquipmentHandler.cs
NeisesMike: I ensured consistency in the names, and I've also added this little optimization. I was a bit hasty, else I would have noticed it myself!

SMLHelper/Interfaces/IEquipmentHandler.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    /// A handler related to EquipmentTypes
+    /// </summary>
+    public interface IEquipmentHandler
+    {
+        /// <summary>
+        /// Registers an equipment type for use when creating a equipment
+        /// </summary>
+        /// <param name="equipmentName">The name of the new equipment type</param>
+        /// <param name="displayName">The display name of the new equipment type</param>
```
Metious, line 12: The `displayName` parameter xml doc is no longer necessary since it's gone.
Also i just noticed the `ModdedEquipmentExists` method is missing from the `IEquipmentHandler` interface. add it there, and also rename the `ModdedEquipmentExists` to `ModdedEquipmentTypeExists` for consistency and clarity reasons. then hopefully we're good to merge.

NeisesMike: I've removed the `displayName` xml doc line, added that method to the interface, added appropriate xml documentation for that method, and changed the method name to ensure consistency.

Metious (approved):
Looks pretty good to me. Tested on 4 branches and everything seems to be working just fine.

Issue #263, opened 2022-05-10
KeyCode from ConfigFile input doesn't recognize controller


celvro:
I'll probably fix this, just tracking the issue.

When I try to use KeyCode in [QuickSlots+](https://www.nexusmods.com/subnautica/mods/984) it allows my 360 controller to activate the input, but it doesn't recognize any inputs afterwords except from keyboard.

For reference: https://github.com/celvro/QuickSlotsPlus/blob/53a81b7cac80ee491e0f398cedcdb336ecf5847b/StandardConfig.cs#L25-L65

In a related controller issue, you can't select the Toggle Headers from SMLHelper in the mod options. 
https://github.com/SubnauticaModding/QModManager/issues/275

PR #264, opened 2022-05-13
Bugfixes for controllers.

celvro:
Fix https://github.com/SubnauticaModding/SMLHelper/issues/263

  - When KeyBinds are created use either controller OR keyboard instead of always keyboard.
     - Ideally these could be changed after the options are rendered based on your last input, but I couldn't figure out how to do that
  - Changed HeadingToggle to extend Selectable instead of Monobehaviour.
     This allows you to "select" headings but they don't get a selected background color and you can't toggle them, however you CAN get into the options with a controller now. They can still be toggled by mouse click.

FYI I found out sliders can't be activated by controller. 😂
not as important as the keybinds though since you can still change them with mouse.

### DIFF
--- a/SMLHelper/Options/Attributes/OptionsMenuBuilder.cs
+++ b/SMLHelper/Options/Attributes/OptionsMenuBuilder.cs
@@ -202,7 +202,8 @@
         private void BuildModKeybindOption(string id, string label, MemberInfoMetadata<T> memberInfoMetadata)
         {
             KeyCode value = memberInfoMetadata.GetValue<KeyCode>(ConfigFileMetadata.Config);
-            AddKeybindOption(id, label, GameInput.Device.Keyboard, value);
+            GameInput.Device primaryDevice = GameInput.IsPrimaryDeviceGamepad() ? GameInput.Device.Controller : GameInput.Device.Keyboard;
+            AddKeybindOption(id, label, primaryDevice, value);
         }
 
         /// <summary>
--- a/SMLHelper/Patchers/OptionsPanelPatcher.cs
+++ b/SMLHelper/Patchers/OptionsPanelPatcher.cs
@@ -184,7 +184,8 @@
                 public static void store(string name, HeadingState state) => statesConfig[name] = state;
             }
 
-            // we add arrow button from Choice ui element to the options headings for collapsing/expanding 
+            // we add arrow button from Choice ui element to the options headings for collapsing/expanding
+            // it cannot be activated by controller since there are only click handlers
             private static void InitHeadingPrefab(uGUI_TabbedControlsPanel panel)
             {
                 if (headingPrefab)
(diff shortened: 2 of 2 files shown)


PrimeSonic (commented):
This looks good to me but I don't remember enough about the keybind options to stamp it

Metious (commented):
This is a good start. Is there anything else regarding the Mods options menu that doesn't work very well with controllers? We can work on expanding this.

celvro:
So far this is what I know about controller:

## Works
* ToggleAttribute - works
* ButtonAttribute - works
  * Button gets un-highlighted (minor issue)
* ChoiceAttribute - works
  * Can change by pressing "A/X" or right analog stick (not left)

## Functional but has issues
* KeyBind - fixed (kind of)
  * mentioned earlier, KeyBind determines whether to accept Keyboard or Controller input when it's rendered. So that happens when "Mods" is selected in the Options menu.
* Enter options/Select ToggleHeaders - fixed (kind of)
  * Header can be selected but the background doesn't change color and it can't be activated by controller.
   
## Bugged 
* SliderAttribute - doesn't work, tested both analog stick and directional pad.

celvro:
I found something else strange, consider this code to log the binding name and keycode name:
```
string bindingName;
// Hotkeys 1-5
if (slotId < 5)
{
    bindingName = GameInput.GetBindingName(GameInput.Button.Slot1 + slotId, GameInput.BindingSet.Primary);
}
else
{
    KeyCode keyCode = (KeyCode)Mod.Config.GetType().GetField("HotKey" + (slotId + 1)).GetValue(Mod.Config);
    bindingName = keyCode.ToString();
}
Logger.Log(Logger.Level.Debug, $"Binding name for slot {slotId}: {bindingName}");
```

Now I went and set 'Y' for both keycode and game binding for slot 1.
Output:
```
[QuickSlotsPlus:DEBUG] Input name for slot 0: ControllerButtonY
[QuickSlotsPlus:DEBUG] Input name for slot 5: JoystickButton3
```

toebeann:
@celvro 
>     * KeyBind - fixed (kind of)
>       
>       * mentioned earlier, KeyBind determines whether to accept Keyboard or Controller input when it's rendered. That happens when the Options menu is opened.
It might be possible to set up something that watches which input type is currently being used and switch accordingly, but this is somewhat low priority as there is a workaround if you back out of the menu and open it again.

>     * Enter options/Select ToggleHeaders - fixed (kind of)
>       
>       * Header can be selected but the background doesn't change color and it can't be activated by controller.
Perhaps we can set it up to so that simply pressing a button will activate it?

> ## Bugged
> 
>     * SliderAttribute
>       
>       * float with no `Step` value works with right stick
>       * int and float with `Step = 1` don't work
Is this in both SN and BZ? IIRC in BZ step is implemented natively, but in SN1 we have a custom implementation. Likely it is the custom implementation of step that is bugged for controller.

>     * Tooltips don't show for selected menu items
Perhaps we can set it up so pressing a button (maybe pressing the right stick in?) would toggle the display of the tooltip. I'd say this is the lowest priority issue here, however.

PR #274, opened 2022-09-04
Converted Projects To .NET SDK

Metious:

  - Converted both ExampleMod and SMLHelper projects to .NET SDK template

The .NET framework msbuild template is now legacy and deprecated, and very hard to work with outside of Windows.
Using the .NET SDK template, while having all of .NET 6 features (outside of the compiler, of course), we can still target .NET Framework 4.7.2 for the build so it is completely backward compatible.

I've tested this on Linux and it works perfectly. I need somebody else to just build the project on Windows to confirm i didn't miss anything.

**Requires .NET 6 to be installed.**


Nagorogan:
Tested it on windows, compiled fine to my knowledge. (not ruling out the possibility that I somehow forgot something, but I'm like 95% sure it's all good)

PR #276, opened 2022-09-11
added MakeEatable method and pickupable patch

EldritchCarMaker:
pickupable patch only gets patched if at least one edible is added, and it takes over the original eatable patch if so. No need to have two patches there when one would do, right?

mostly so that CC2 can allow for items to be made edible solely through smlhelper.

### DIFF
--- a/SMLHelper/Handlers/EatableHandler.cs
+++ b/SMLHelper/Handlers/EatableHandler.cs
@@ -28,6 +28,15 @@
                 decomposes = decomposes,
             });
         }
+        void IEatableHandler.MakeEatable(TechType item, float food, float water, bool decomposes)
+        {
+            PickupablePatcher.AddedEatables.Add(item, new EditedEatableValues()
+            {
+                food = food,
+                water = water,
+                decomposes = decomposes,
+            });
+        }
         /// <summary>
         /// Use this to change the values of a specific TechType.
         /// </summary>
@@ -39,7 +48,17 @@
         {
             Main.ModifyEatable(item, food, water, decomposes);
         }
-
+        /// <summary>
+        /// Use this to change the values of a specific TechType.
+        /// </summary>
+        /// <param name="item">The TechType of the item you want to change.</param>
+        /// <param name="food">The food value you want to change it to.</param>
+        /// <param name="water">The water value you want to change it to.</param>
+        /// <param name="decomposes">Whether or not the item decomposes over time</param>
+        public static void MakeEatable(TechType item, float food, float water, bool decomposes = true)
+        {
+            Main.MakeEatable(item, food, water, decomposes);
+        }
 #elif BELOWZERO 
         void IEatableHandler.ModifyEatable(TechType item, float food, float water, bool decomposes, float health, float coldValue, int maxCharges)
         {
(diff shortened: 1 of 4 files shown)


SMLHelper/Interfaces/IEatableHandler.cs
```diff
@@ -14,19 +14,39 @@ public interface IEatableHandler
...
         /// <param name="decomposes">Whether or not the item decomposes over time, losing food and water values in the process.</param>
         void ModifyEatable(TechType item, float food, float water, bool decomposes = true);
+        /// <summary>
+        /// Allows you to change the water/food values of a specific item.
+        /// </summary>
+        /// <param name="item">The TechType of the item you wish to edit.</param>
+        /// <param name="food">The food value you want the item to have.</param>
+        /// <param name="water">The water value you want the item to have.</param>
+        /// <param name="decomposes">Whether or not the item decomposes over time, losing food and water values in the process.</param>
+        void MakeEatable(TechType item, float food, float water, bool decomposes = true);
```
Metious, line 24: Make it more descriptive and accurate (edit -> add, etc..)

SMLHelper/Interfaces/IEatableHandler.cs
```diff
@@ -14,19 +14,39 @@ public interface IEatableHandler
...
+        /// </summary>
+        /// <param name="item">The TechType of the item you want to change.</param>
+        /// <param name="food">The food value you want to change it to.</param>
+        /// <param name="water">The water value you want to change it to.</param>
+        /// <param name="decomposes">Whether or not the item decomposes over time.</param>
+        /// <param name="health">How much you want to set the health gained from eating this item to.</param>
+        /// <param name="coldValue">How much eating this item changes the current cold meter value.<br/>
+        /// Negative values heats up the player while positive values makes the player colder.</param>
+        /// <param name="maxCharges">How many times the item can be used before being consumed.</param>
+        void MakeEatable(TechType item, float food, float water, bool decomposes = true, float health = 0f, float coldValue = 0f, int maxCharges = 0);
```
Metious, line 49: Make the doc more descriptive and accurate (edit -> add, etc..)

SMLHelper/Initializer.cs
```diff
@@ -79,7 +79,11 @@ public static void PostPatch()
             TooltipPatcher.Patch(harmony);
             SurvivalPatcher.Patch(harmony);
             CustomSoundPatcher.Patch(harmony);
-            EatablePatcher.Patch(harmony);
+
+            if (PickupablePatcher.AddedEatables.Count > 0)
+                PickupablePatcher.Patch(harmony);
+            else
+                EatablePatcher.Patch(harmony);
```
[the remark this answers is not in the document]
PrimeSonic: That else shouldn't be there.
It's only the pickup able patcher that needs to be skipped.

SMLHelper/Handlers/EatableHandler.cs
```diff
@@ -8,6 +8,7 @@
     /// </summary>
     public class EatableHandler : IEatableHandler
     {
+        private static bool pickupablePatched = false;
```
Metious, line 11: Looks like this is never assigned. Move it to the patcher and set it to `true` after patching.
The handler should only call `PickupablePatcher.Patch` and have no business in determining if it's patched yet or not.

Metious (changes requested):

PR #277, opened 2022-09-14
Automatic Icon Loading Fixes

### LINKED ISSUE #273: Automatic icon loading for Spawnable doesn't work
The directions to simply place a .png in the mod directory don't work, because it looks in the wrong place, because AssetsFolder doesn't default to the mod directory.

`Assembly.GetExecutingAssembly()` is the wrong call here; at that instant, code from SMLHelper is running, so it will always return the SMLHelper directory. 

https://github.com/SubnauticaModding/SMLHelper/blob/8dc2aae055f52cf0b9e2fdaf8a61dd768cc6bc17/SMLHelper/Assets/Spawnable.cs#L184

Leading to:

```
[SMLHelper/Error] Sprite for 'SurveyToolPrefab'
Did not find an image file at 'C:\Program Files (x86)\Steam\steamapps\common\Subnautica\QMods\Modding Helper\SurveyTool.png'
```

Presumably this wants to just reference the internal `Mod` field set by ModPrefab, yes?

https://github.com/SubnauticaModding/SMLHelper/blob/8dc2aae055f52cf0b9e2fdaf8a61dd768cc6bc17/SMLHelper/Assets/ModPrefab.cs#L81

Metious:
Fix #273 

  - Fixed automatic mod folder location targeting SMLHelper's folder rather than the mod in question's folder.

### DIFF
--- a/SMLHelper/Assets/Spawnable.cs
+++ b/SMLHelper/Assets/Spawnable.cs
@@ -41,7 +41,7 @@
         /// By default, this will point to the same folder where your mod DLL is.
         /// </summary>
         /// <example>"MyModAssembly/Assets"</example>
-        public virtual string AssetsFolder => modFolderLocation;
+        public virtual string AssetsFolder => ModFolderLocation;
 
         /// <summary>
         /// Override with the file name for this item's icon.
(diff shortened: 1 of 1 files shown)


PrimeSonic (approved):
Looks good.
Minor nitpick: since this is now a (get only) Property and not a Field anymore, the name should be capitalized.

That would cascade into more changes so I'm fine with this as a small fix.

PR #278, opened 2022-09-23
Add the ability to register assemblies as "skipped" in reflective stack checking

ReikaKalseki:
To allow for core mods to not be listed as the owner of items, for example.

Also includes SML and the two base-game DLLs by default.

### DIFF
--- a/SMLHelper/Utility/ReflectionHelper.cs
+++ b/SMLHelper/Utility/ReflectionHelper.cs
@@ -169,11 +169,23 @@
         }
 
         private static readonly Assembly smlHelperAssembly = Assembly.GetExecutingAssembly();
+        
+        private static readonly HashSet<Assembly> assembliesToSkip = new HashSet<Assembly>(){
+            smlHelperAssembly,
+            Assembly.GetAssembly(typeof(LargeWorldEntity)), //Assembly-CSharp
+            Assembly.GetAssembly(typeof(FMODAsset)) //firstpass
+        };
+        
+        /** Adds an assembly to a set to be skipped when reflectively determining "calling mod". Usually used for 'core' mods. */
+        public static void addSkippedAssembly(Assembly a) {
+            assembliesToSkip.Add(a);
+        }
 
         internal static string CallingAssemblyNameByStackTrace()
         {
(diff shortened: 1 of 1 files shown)


SMLHelper/Utility/ReflectionHelper.cs
```diff
@@ -169,11 +169,23 @@ public static class ReflectionHelper
...
         private static readonly Assembly smlHelperAssembly = Assembly.GetExecutingAssembly();
+        
+        private static readonly HashSet<Assembly> assembliesToSkip = new HashSet<Assembly>(){
+            smlHelperAssembly,
+            Assembly.GetAssembly(typeof(LargeWorldEntity)), //Assembly-CSharp
+            Assembly.GetAssembly(typeof(FMODAsset)) //firstpass
+        };
+        
+        /** Adds an assembly to a set to be skipped when reflectively determining "calling mod". Usually used for 'core' mods. */
+        public static void addSkippedAssembly(Assembly a) {
```
PrimeSonic, line 180:
```suggestion
        /// <summary>
        /// Adds an assembly to a set to be skipped when reflectively determining "calling mod". Usually used for 'core' mods,
        /// </summary>
        /// <param name="assembly">The assembly to register<param>
        public static void addSkippedAssembly(Assembly assembly) {
```

PrimeSonic (changes requested):
Looks good to me.
Added a comment to clean up the documentation.
Will add the others to also review.

Metious:
Idea behind this makes sense, we tried to achieve the same thing before, and it worked for the use-cases at that time.
As of SMLHelper 2.9.6, the `TechTypeHandler` now also takes an assembly as an argument to set the "owner" of that TechType: #211.

These changes of yours look great, however, it's very hard to notice that it is there, unless the modder in question knows how SMLHelper works.

The `Mod` field in the `Spawnable` class was changed from a string to an assembly, and was also changed to use that new overload. However, it's `internal` at the meantime meaning mods cant assign it. I'm proposing to just make it `protected` so you can easily change the "owner" of the item, and also more obvious for modders to notice it.

PR #280, opened 2022-10-15
Fix compound tech patching

EldritchCarMaker:

  -fixed KnownTechPatcher from setting all CompoundTech dependencies to the most recently added ones
  -added some comments, by request of Lee23 cause some of the syntax can be confusing
  -reduced the code down to one foreach loop, rather than the previous three loops

  -includes portions of old code commented out, I am hoping someone can test this further than I can and make sure it still works as intended, as this code does not seem to be necessary but it very well may be and I just haven't tested enough

### DIFF
--- a/SMLHelper/Patchers/KnownTechPatcher.cs
+++ b/SMLHelper/Patchers/KnownTechPatcher.cs
@@ -86,30 +86,19 @@
                     analysisTech.Add(tech);
             }
 
-            // make sure the KnownTech class has the correct modded values
-
             List<KnownTech.CompoundTech> compoundTech = KnownTech.compoundTech;
-
-
-            // get all of the techtypes that are in the CompoundTech.Values collection but not in the KnownTech.compoundTech list
-            IEnumerable<KnownTech.CompoundTech> compoundTechToAdd = CompoundTech.Values.Where(a => !compoundTech.Any(a2 => a.techType == a2.techType));
-
-
-            // go through every custom compound tech in the CompoundTech dictionary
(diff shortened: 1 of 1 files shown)


SMLHelper/Patchers/KnownTechPatcher.cs
```diff
@@ -86,25 +86,31 @@ internal static void InitializePostfix()
...
-                    compoundTech.Add(tech);
+
+                if (!KnownTech.Contains(customTech.techType))//if it's not already there, add it to known tech
+                    compoundTech.Add(customTech);
+
+
+                //find the tech in the compoundTech list
+                var foundTech = compoundTech.Find(tech => tech.techType == customTech.techType);
+
+                if (foundTech != null)//set its dependencies properly
```
Metious, line 112:
```suggestion
            
            List<KnownTech.CompoundTech> compoundTech = KnownTech.compoundTech;

            foreach (KnownTech.CompoundTech customTech in CompoundTech.Values)
            {
                if (customTech == null) // Safety check
                    continue;
                
                // Only add the new compound tech if it isn't unlocked yet 
                if (!KnownTech.Contains(customTech.techType))
                    compoundTech.Add(customTech);
                
                // If a compound tech already exists, set the dependencies correctly.
                var foundTech = compoundTech.Find(tech => tech.techType == customTech.techType);

                if (foundTech != null)
```

Metious (commented):
Looks good overall. Left a suggestion with some comments and old code cleanup.

PR #290, opened 2022-12-27
2.15.0.1 Hotfix for KnownTech.Analyze null ref exception with custom scannables 

MrPurple6411:
Validate modded analysisTechs and compoundTechs before adding to games lists
Initialize AnalysisTech.storyGoals list as it is no longer auto created and will cause a null ref exception after a modded tech with a scannable is unlocked

[SMLHelper_SN.STABLE.zip](https://github.com/SubnauticaModding/SMLHelper/files/10307105/SMLHelper_SN.STABLE.zip)

### DIFF
--- a/Example mod/Mod.cs
+++ b/Example mod/Mod.cs
@@ -21,7 +21,7 @@
             MODNAME = "SMLHelper",
             AUTHOR = "The SMLHelper Dev Team",
             GUID = "SMLHelperExampleMod",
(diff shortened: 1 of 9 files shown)


MrPurple6411:
Forcing through hotfix for release.

Have tested with TechPistol mod and verified this fixes it as that is where I was getting the reports of this issue.

PR #293, opened 2023-01-06
Try play sound

MrPurple6411:

  - Depreciated PlaySound
  - Added 3 versions of TryPlaySound
  - Fixed potential Index out of bounds issue in FModMultiSounds.PlaySound
  - Converted patches to use the new TryPlaySound


SMLHelper/FMod/FModMultiSounds.cs
```diff
@@ -79,13 +81,14 @@ public FModMultiSounds(string[] soundPaths, MODE mode, string busPath, bool rand
...
 
+        [Obsolete("Deprecated. Use TryPlaySound instead.")]
         Channel IFModSound.PlaySound()
         {
             if (_sounds is {Length: > 0})
             {
                 if (randomizeSounds)
                 {
-                    return AudioUtils.PlaySound(_sounds[Random.Range(0, _sounds.Length)], _bus);
+                    return AudioUtils.PlaySound(_sounds[Random.Range(0, _sounds.Length - 1)], _bus);
```
Metious, line 91: `Random.Range` for integers is min-inclusive & max-exclusive as opposed to range for floats, so it should be 0 to length, or the last sound in the collection wont be ever played.
```suggestion
                    return AudioUtils.PlaySound(_sounds[Random.Range(0, _sounds.Length)], _bus);
```

SMLHelper/FMod/FModMultiSounds.cs
```diff
@@ -94,5 +97,23 @@ Channel IFModSound.PlaySound()
...
         }
+
+        bool IFModSound.TryPlaySound(out Channel channel)
+        {
+            channel = default;
+            if (_sounds is {Length: > 0})
+            {
+                if (randomizeSounds)
+                {
+                    int index = Random.Range(0, _sounds.Length-1);
```
Metious, line 108: Same here
```suggestion
                    int index = Random.Range(0, _sounds.Length);
```

Metious (changes requested):
Left some comments.

Looks good overall to me, but I wonder what's the reason behind the `TryPlaySound` change?
Also we can remove all of `PlaySound` & `TryPlaySound` overloads without a bus since those are fragile and completely ignore the sound channels of the game.

MrPurple6411:
> Looks good overall to me, but I wonder what's the reason behind the `TryPlaySound` change?
This way mods can actually programmatically tell if the sound has failed to play and put in logging for it.
Also all those methods had the returns for not ok so what happens if one failed and it just kept going to the next? 

> Also we can remove all of `PlaySound` & `TryPlaySound` overloads without a bus since those are fragile and completely ignore the sound channels of the game.

I reworked them to actually use the proper Master Bus of the game if a bus was not specified.
That is really only meant for sound that are supposed to ignore the channels of the game.

I guess we could remove them and anyone who wants to use the master bus will have to go and figure out how to get the list of them like I did.

Are you sure you want them removed though? I just added them to make testing sounds and stuff easier before deciding on a bus.

That being said I am 100% good to remove ALL of the normal PlaySound all together as they have no fail checks atm and idk if that will break things when they fail.

Metious:
We have the [BusPaths](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Utility/AudioUtils_Subnautica.cs#L9) class that pretty much covers all types of sounds. Sounds shouldn't play in the master bus because it doesn't have any effects applied to it at all, therefore very difficult to manage.

Also since all other buses are under the master bus, messing it up will pretty much destroy all sounds.

> I reworked them to actually use the proper Master Bus of the game if a bus was not specified.
That is really only meant for sound that are supposed to ignore the channels of the game.

If there's this possibility, then we should probably also add it in the `BusPaths` class, even though I very much hate it.
Curious to learn more about those use-cases, though

> That being said I am 100% good to remove ALL of the normal PlaySound all together as they have no fail checks atm and idk if that will break things when they fail.

I believe we should remove all obsolete-marked overloads since all of them are discouraged.
That includes the `PlaySound` methods and `SoundChannel` enumeration. Should've been part of #291 though

MrPurple6411:
> We have the [BusPaths](https://github.com/SubnauticaModding/SMLHelper/blob/master/SMLHelper/Utility/AudioUtils_Subnautica.cs#L9) class that pretty much covers all types of sounds. Sounds shouldn't play in the master bus because it doesn't have any effects applied to it at all, therefore very difficult to manage.

Ohh that would have been good to find when I was trying to help CC with sound stuff earlier! lol

Metious (approved):

PR #297, opened 2023-01-29
Fix BZ build

JKohlman:

  - Fixes to allow BZ to build (untested if it works in game)


JKohlman:
Going to close as Purple has dealt with most of this it seems, will take a look again if need be after their merge.

PR #302, opened 2023-02-26
Change Options from List to IReadOnlyCollection

JKohlman:
Moving to an `IReadOnlyCollection` so that this value is more obvious that it is non-editable.

Not tested but builds and should be fairly straightforward as it's only really used in 2 places:

https://github.com/SubnauticaModding/SMLHelper/blob/8178e8b451a331a7c9b4d90f061c14d52682e082/SMLHelper/Options/Attributes/OptionsMenuBuilder.cs#L115

https://github.com/SubnauticaModding/SMLHelper/blob/8178e8b451a331a7c9b4d90f061c14d52682e082/SMLHelper/Options/ModOptions.cs#L81-L85

### DIFF
--- a/SMLHelper/Options/Attributes/OptionsMenuBuilder.cs
+++ b/SMLHelper/Options/Attributes/OptionsMenuBuilder.cs
@@ -100,7 +100,7 @@
         /// <summary>
         /// Adds options to the menu based on the <see cref="ConfigFileMetadata"/>.
         /// </summary>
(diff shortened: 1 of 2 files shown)


Metious:
I believe `IReadOnlyCollection` makes more sense if the goal is to just emphasize on the fact that there aren't any editing involved.  
With this, the `Count` property will stay as well.

JKohlman:
Swapped as suggested! I'd definitely recommend a squash merge because of the revert but otherwise should be fine.

PR #303, opened 2023-03-09
Material modifiers fix

LeeTwentyThree:

  - Improved design of MaterialModifiers to be viable for general usage.
    - Fixed a bug that stopped them from running on most materials.
    - Made BlockShaderConversion virtual (it should have been from the start).
  - Updated documentation of MaterialUtils and related classes.
  - Added basic ColorModifier material modifier for example purposes & basic use cases.
  - Pass in `materialIndex` to allow swapping of materials through renderers by index.


Metious (approved):
Eventually ended up removing the `ApplyChangesToMaterial` method as it was doing the same thing as the `EditMaterial`. Looks good to me overall

PR #307, opened 2023-04-17
Prefab info doc changes

LeeTwentyThree:

  - Added a mention to the WithTechType method in PrefabInfo docs as a convenient reminder.
  - Fixed minor issue with docs pointing to TechType property rather than type.

### DIFF
--- a/SMLHelper/Assets/PrefabInfo.cs
+++ b/SMLHelper/Assets/PrefabInfo.cs
@@ -13,7 +13,7 @@ using Handlers;
 /// </summary>
 /// <param name="ClassID">The class identifier used for the <see cref="PrefabIdentifier"/> component whenever applicable.</param>
 /// <param name="PrefabFileName">Name of the prefab file.</param>
-/// <param name="TechType">The <see cref="TechType"/> of the corresponding item.</param>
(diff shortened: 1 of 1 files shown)


SMLHelper/Assets/PrefabInfo.cs
```diff
@@ -8,11 +8,12 @@ namespace SMLHelper.Assets;
 using Handlers;
 
 /// <summary>
-/// Data class that represents information about a prefab.
+/// <para>Data class that represents information about a prefab.</para>
+/// <para>Also see <see cref="WithTechType"/>, which may be more convenient than directly calling the constructor.</para>
```
Metious, line 12: Instead of a paragraph at the top, I think it would be better if we use a `seealso` tag. This will help make it better looking in our [API docs](https://subnauticamodding.github.io/SMLHelper/api/SMLHelper.Assets.html) as well.

I included an existing example below.

https://github.com/SubnauticaModding/SMLHelper/blob/38556d8972ca0b8d99f90abe9955668956c1f699/SMLHelper/Handlers/ConsoleCommandsHandler.cs#L54-L56

Metious (changes requested):

LeeTwentyThree:
I attempted this, but it seems to defeat the point since it isn't visible through IDEs.

PR #348, opened 2023-05-18
Create AssetBundleTemplate.cs

EldritchCarMaker:
Added new PrefabTemplate, used to load asset bundles and prefabs from them


32Kallies:
`AssetBundleTemplate(Assembly modAssembly, string assetBundleFileName, string prefabName, PrefabInfo info)` seems dangerous for memory and could be slow. Otherwise looks good.

32Kallies (changes requested):
Option 1: Consider removing the overload that loads an Asset Bundle, or at least give a warning. This would only be safe in a mod that is very small or has very few assets.

Option 2: Cache loaded bundles in an internal dictionary somewhere (filename: bundle).

EldritchCarMaker:
There's still a few problems with it, but I think its good enough for use

The constructor overload that loads the bundle won't work well with mods that have multiple bundles. This is because it caches the bundles based on Assembly. Unfortunately we can't go by file name, because then multiple mods could not have bundles named, for instance, `assets`, or any other common and non-descript name. I think it's more important that other mods don't mess with each other with commonly named bundles than it is for the rare time when a mod uses more than one bundle

It doesn't return the bundle once it's loaded, so the user would have to load it themselves if they wanted to use the bundle for their own purposes (like loading a sprite, or other non-GameObject asset). This could be solved with an `out` argument, but I think that would be annoying to deal with for mods with a larger number of assets, as there would be an extra argument in there that they wouldn't need, but would still have to supply.

Unfortunately I can't think of a good way to handle these

32Kallies:
Asset Bundles already require unique names, so that can be safely used as a unique identififer.

"... the runtime asset bundle cache system uses Name & Hash pair to uniquely identify each asset bundle on disk. So if you rename a bundle, it is considered a new bundle when requested even if the hash is one you used previously." ([Unity forum post](https://forum.unity.com/threads/asset-bundles-names-and-hashes.536111/))

Look into the `Path.GetFileName(string)` method for finding the name for identification purposes.

32Kallies:
> Unfortunately we can't go by file name, because then multiple mods could not have bundles named, for instance, `assets`, or any other common and non-descript name. I think it's more important that other mods don't mess with each other with commonly named bundles than it is for the rare time when a mod uses more than one bundle

Correct me if I'm wrong but I believe this is already the case. In the past I have had to rename my asset bundles because of conflicting mods.

32Kallies:
How about this change/overload using [Assembly.GetCallingAssembly()](https://learn.microsoft.com/en-us/dotnet/api/system.reflection.assembly.getcallingassembly?view=net-7.0)?

```csharp
// private static Dictionary<Assembly, AssetBundle> _loadedBundles = new Dictionary<Assembly, AssetBundle>();
// use asset bundle file name for identifier
private static Dictionary<string, AssetBundle> _loadedBundles = new Dictionary<string, AssetBundle>();

/// ....
public AssetBundleTemplate(string assetBundleFileName, string prefabName, PrefabInfo info) : base(info)
    {
        AssetBundle bundle;

        if(!_loadedBundles.TryGetValue(assetBundleFileName, out bundle))
        {
            bundle = Utility.AssetBundleLoadingUtils.LoadFromAssetsFolder(Assembly.GetCallingAssembly(), assetBundleFileName);
            _loadedBundles.Add(assetBundleFileName, bundle); // iirc you forgot to add it to the cache
        }
        
        _prefab = bundle.LoadAsset<GameObject>(prefabName);
    }
```

32Kallies:
I would like to see this in Nautilus soon but currently I think as it is implemented it could be improved, or do you think it's fine as is?

EldritchCarMaker:
> Look into the `Path.GetFileName(string)` method for finding the name for identification purposes.

This just returns the file name of the bundle, which is already included in the method arguments anyway, so this won't help much

EldritchCarMaker:
> > Unfortunately we can't go by file name, because then multiple mods could not have bundles named, for instance, `assets`, or any other common and non-descript name. I think it's more important that other mods don't mess with each other with commonly named bundles than it is for the rare time when a mod uses more than one bundle
> 
> Correct me if I'm wrong but I believe this is already the case. In the past I have had to rename my asset bundles because of conflicting mods.

The bundle names themselves have to be different, yes, but the file names can be the same. I just tested this in game using two separate bundles, where I renamed one to the same as the other. Both bundles had the same file name, and could be loaded together just fine. Once they were loaded, you could use `bundle.name` to get the actual unique name of the bundle, but that doesn't help much when trying to load/cache them as you'd only be able to access that `bundle.name` field once it's been loaded

EldritchCarMaker:
There's three routes we can use to cache the bundles

We can go by assembly, which will work perfectly for the vast majority of mods as very few have multiple asset bundles for the same assembly, but won't work 100% of the time because sometimes, very rarely, a mod may use two separated asset bundles.

We can also go by file name, which will allow mods to have multiple bundles but also means that any two mods that have the same file names as each other would be unable to use the template, because the caches would interfere (note: you *can* still use two bundles with the same file name, as long as those two bundles do not share the same internal name. It's unlikely for people to rename their bundles, but it is still a possibility.)

Or we can go by both, and use a tuple or some other struct to store both the assembly and the file name. This would, I believe, solve both issues. But I'm also lazy and don't think it's worth doing

I think the best option is to stick with the assembly caching, as the vast majority of mods don't use multiple bundles so they wouldn't be affected in any major way, and there is only one mod that I know of that uses two bundles and it wouldn't be using this template anyway.

32Kallies:
Not sure why assembly caching would be a good idea, when we already have bundle names to go by which should be a lot safer. It is *not* unusual to use multiple asset bundles. For example I have a mod project that uses an ancient asset bundle alongside a new one because we lost the source unity project and can't merge all the assets into one bundle. Any large mod could use it.

On the other hand, renaming an asset bundle seems quite unusual to me.

EldritchCarMaker:
recheck

EldritchCarMaker:

Nautilus/Assets/PrefabTemplates/AssetBundleTemplate.cs
```diff
@@ -0,0 +1,61 @@
...
+    /// <summary>
+    /// Instantiates a new AssetBundleTemplate. Automatically loads the bundle by calling <see cref = "Utility.AssetBundleLoadingUtils.LoadFromAssetsFolder(Assembly, string)"/>
+    /// <para>Also caches the loaded bundle for future use</para>
+    /// <para>If you are loading and using your bundle on your own, it's highly recommended to use the AssetBundle constructor overload instead</para>
+    /// <para>Bundles are cached per Assembly, and won't work with mods that use multiple seperate bundles</para>
+    /// </summary>
+    /// <param name="assetBundleFileName">The file name of the asset bundle. These often do not have file extensions</param>
+    /// <param name="prefabName">The name of the prefab gameobject to load from the bundle</param>
+    /// <param name="info">The prefab info to base this template off of.</param>
+    public AssetBundleTemplate(string assetBundleFileName, string prefabName, PrefabInfo info) : base(info)
```
32Kallies: Pass in the Assembly as a parameter. Wouldn't `Assembly.GetExecutingAssembly()` return Nautilus?

32Kallies (changes requested):

Nautilus/Assets/PrefabTemplates/AssetBundleTemplate.cs
EldritchCarMaker: Actually you said to remove the assembly argument. But also yes, I accidentally used ExecutingAssembly instead of CallingAssembly

PR #369, opened 2023-05-24
Allow controllers to use options menu

### LINKED ISSUE #263: KeyCode from ConfigFile input doesn't recognize controller

### LINKED ISSUE #289: Cannot interact with Mods menu using controller
For the stock Options menu you can navigate using a controller with the Dpad and A/B for select and back. Trying to enter the Mods menu with the controller highlights the Mods button but otherwise has no effect. You do need to hit back to get to any of the Stock options menus.

This is for version 2.15 on Subnautica Living Large update.

celvro:
1. GameInput is hardcoded to keyboard, changed to use primary device. This updates when the menu is opened. Maybe there's a better spot to attach the device changing so it works while the menu is open
2. Fixed when controller gets stuck on the HeadingToggle since it only has a click handler. Thanks @JKohlman for fixing the selected background and button toggle!
3. Added controller icons for the KeyBind input! this uses uGUI.buttonCharacters internally so theoretically it works with anything in that array

fixes #263, #289

### DIFF
--- a/Nautilus/Options/Attributes/OptionsMenuBuilder.cs
+++ b/Nautilus/Options/Attributes/OptionsMenuBuilder.cs
@@ -245,7 +245,7 @@ internal class OptionsMenuBuilder<T> : ModOptions where T : ConfigFile, new()
     private void BuildModKeybindOption(string id, string label, MemberInfoMetadata<T> memberInfoMetadata)
     {
         KeyCode value = memberInfoMetadata.GetValue<KeyCode>(ConfigFileMetadata.Config);
-        if(!AddItem(ModKeybindOption.Create(id, label, GameInput.Device.Keyboard, value)))
+        if(!AddItem(ModKeybindOption.Create(id, label, GameInput.GetPrimaryDevice(), value)))
             InternalLogger.Warn($"Failed to add ModKeybindOption with id {id} to {Name}");
     }
 
--- a/Nautilus/Options/ModKeybindOption.cs
+++ b/Nautilus/Options/ModKeybindOption.cs
@@ -27,7 +27,7 @@ public class ModKeybindOption : ModOption<KeyCode, KeybindChangedEventArgs>
     /// <summary>
     /// The currently select input source device for the <see cref="ModKeybindOption"/>.
     /// </summary>
-    public GameInput.Device Device { get; }
+    public GameInput.Device Device { get; set; }
 
     /// <summary>
     /// The tooltip to show when hovering over the option.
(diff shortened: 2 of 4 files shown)


32Kallies (approved):
What I have seen is great and what I can test works well, although I don't have a controller myself to try. Regarding the code, it seems like a harmless change with very little room for errors.

JKohlman:
Seems to work, does leave a couple issues/feature requests open - mainly actually showing selection on tabs and the Clear implementation

JKohlman (commented):
Seems to work, does leave a couple issues/feature requests open - mainly actually showing selection on tabs and the Clear implementation.

@celvro are those things you'd want to look into for it or just at a later time?

JKohlman:
Also not a huge issue but to note: scrolling down it seems like we do invisibly select the headers but scrolling up it seems like we just immediately skip over the header

JKohlman (changes requested):
@celvro Please take a look at https://github.com/celvro/Nautilus/pull/1 for potentially more improvements and approve, modify or reject that before we can merge this.

In that PR there is still a bug where the cursor skips the option above each heading when scrolling upwards

celvro:
Ok looks good! I also fixed ps4 icons, my controller was sending the wrong inputs but the icons looked correct for the buttons it was sending 🤷 

Here's a list of known issues, I'm fine merging as is and fixing incrementally:

1. Options don't update immediately when primary device changes, you have to close and open the menu first.
2. If you close the final options header, it lets you keep scrolling down and starts scrolling through tabs. You can't activate them until you hit back.
3. The clear button doesn't work.
4. When scrolling up from a header it skips one option.
5. If SMLHelper is installed, the options panel doesn't get patched. This is unrelated to the controller fixes and seems to be intentional, Nautilus checks it in OptionsPanelPatcher before patching.

JKohlman:
I'm helping a family member move today but can you just confirm whether 2 and 4 are completely Nautilus bugs or whether they happen in vanilla game/options menus too?

celvro:
Neither of them happen on vanilla but I don't think it has collapsible headers

On 4 I tested with a second mod installed and it _didn't_ skip over the last setting on that one. It still happened on the Nautilus option. Weird but imo it's not a big deal

JKohlman:
Approved with the caveat that we note down the issues and take another crack at figuring them out but they're pretty weird/abstract

JKohlman (approved):

PR #372, opened 2023-05-29
Fix AnalysisTech unlock message & sound

vlyon:

  - Fixed spelling in `DefaultUnlockMessage`
  - Added `DefaultPickupMessage`, which is used when `RequiredForUnlock == TechType.None`
  - Fixed missing `unlockSound`

### DIFF
--- a/Nautilus/Patchers/KnownTechPatcher.cs
+++ b/Nautilus/Patchers/KnownTechPatcher.cs
@@ -61,7 +61,7 @@ internal class KnownTechPatcher
             
         foreach (KnownTech.AnalysisTech tech in analysisTech)
         {
-            if (UnlockSound == null && tech.unlockSound != null && tech.techType == TechType.Lead)
+            if (UnlockSound == null && tech.unlockSound != null && tech.techType == TechType.CreepvinePiece)
             {
                 UnlockSound = tech.unlockSound;
             }
--- a/Nautilus/Assets/Gadgets/ScanningGadget.cs
+++ b/Nautilus/Assets/Gadgets/ScanningGadget.cs
@@ -13,7 +13,8 @@ namespace Nautilus.Assets.Gadgets;
 /// </summary>
 public class ScanningGadget : Gadget
 {
-    private const string DefaultUnlockMessage = "NotficationBlueprintUnlocked";
+    private const string DefaultPickupMessage = "NotificationBlueprintPickup";
+    private const string DefaultUnlockMessage = "NotificationBlueprintUnlocked";
 
     private bool _isBuildable;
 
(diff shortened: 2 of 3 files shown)


Nautilus/Patchers/KnownTechPatcher.cs
```diff
@@ -61,7 +61,7 @@ internal static void InitializePostfix()
             
         foreach (KnownTech.AnalysisTech tech in analysisTech)
         {
-            if (UnlockSound == null && tech.unlockSound != null && tech.techType == TechType.Lead)
+            if (UnlockSound == null && tech.unlockSound != null && tech.techType == TechType.CreepvinePiece)
```
32Kallies, line 64: I'm surprised that went unnoticed, definitely a needed change. However I noticed that when I play the game, unlock sounds/notifications for KnownTech are hidden ENTIRELY. It occurs in vanilla, so might be a game bug or user error.

32Kallies:
Sorry for hijacking this PR, I never realized how messed up this gadget file is. I got carried away with writing assorted changes instead of leaving written reviews.

Nautilus/Patchers/KnownTechPatcher.cs
vlyon: True, but the patch below fixes the bug in subnautica:

```c#
// This patch fixes a bug in Subnautica that silences notifications like unlocked blueprints
[HarmonyPatch(typeof(Pickupable), nameof(Pickupable.Pickup))]
class Pickupable_Pickup_Patch
{
    static void Postfix(Pickupable __instance, bool events)
    {
        if (events)
            Analyze(__instance.GetTechType());
    }
}
```

I wasn't sure if I should suggest adding it to Nautilus. They may fix the bug in future.

32Kallies: We are currently waiting to see if that change was intentional or not.

vlyon: I'm sure it wasn't intentional. All the "creature discovered" & "new blueprint" sounds are still there, and calling `KnwonTech.Analyze` with `verbose` set to `true` makes it work, but only if you haven't already called it with `verbose` set to `false`. Unfortunately the code now _always_ calls it with `false` first, before calling it with `true`, which means it's _always_ silent.

32Kallies: Exactly. Since we aren't sure about whether we should provide a fix via Nautilus or wait for UWE to fix it officially, I think this PR is good to go (as long as you are fine with it) and we can make a new issue report for that.

vlyon: Sounds Good.

32Kallies (approved):
Seems good for now. We will have to check again after we hear back from UWE.

PR #397, opened 2023-06-03
Mod Databank entries.

jonahnm:
### Changes made in this pull request

  - Handler to add databank entries for mods.
  
May need further testing.

PR #400, opened 2023-06-08
Change a singular check so that #397 works.

jonahnm:

- `if (uGUI.isMainLevel)` changed to `if(PDAEncyclopedia.initialized)`

### Breaking changes

- None (as far as I know)


32Kallies:
Edited the cursed formatting of the original description

32Kallies (approved):
This seems like a much safer check than the original one which was rather arbitrary.

32Kallies:
I noticed there are other methods that still use `uGUI.isMainLevel` as a check. I think this is "good enough" but we should probably look into that further down the line.

PR #403, opened 2023-06-10
Improve PDA item sorting capabilities

vlyon:

  - Ability to append after or insert before a target item when adding to a PDA category
  - Ability to insert before all entries when adding to a PDA category

### Breaking changes

- `CraftDataHandler.AddToGroup(TechGroup, TechCategory, TechType, TechType)` overload was modified.


Nautilus/Assets/Gadgets/GadgetExtensions.cs
```diff
@@ -107,13 +107,52 @@ public static ScanningGadget SetPdaGroupCategory(this ICustomPrefab customPrefab
         if (!customPrefab.TryGetGadget(out ScanningGadget scanningGadget))
         {
             scanningGadget = new ScanningGadget(customPrefab, TechType.None);
+            customPrefab.AddGadget(scanningGadget);
         }
 
-        scanningGadget.WithPdaGroupCategory(group, category);
+        return scanningGadget.WithPdaGroupCategory(group, category);
+    }
```
32Kallies: Since we haven't really decided on PR #404, I would suggest changing this to the previous implementation of a gadget extension.

Nautilus/Assets/Gadgets/GadgetExtensions.cs
```diff
@@ -107,13 +107,52 @@ public static ScanningGadget SetPdaGroupCategory(this ICustomPrefab customPrefab
...
+    /// become buildable. To avoid this, or make this item a buildable manually, use the <see cref="ScanningGadget.SetBuildable"/> method.</remarks>
+    public static ScanningGadget SetPdaGroupCategoryBefore(this ICustomPrefab customPrefab, TechGroup group, TechCategory category, TechType target)
+    {
+        if (!customPrefab.TryGetGadget(out ScanningGadget scanningGadget))
+        {
+            scanningGadget = new ScanningGadget(customPrefab, TechType.None);
+            customPrefab.AddGadget(scanningGadget);
+        }
+
+        return scanningGadget.WithPdaGroupCategoryBefore(group, category, target);
```
32Kallies: While the methods look great, again I would request you to instead use the old method of doing things that used TryAddGadget.

Nautilus/Assets/Gadgets/ScanningGadget.cs
```diff
@@ -40,6 +40,16 @@ public class ScanningGadget : Gadget
     /// </summary>
     public TechCategory CategoryForPda { get; set; }
 
+    /// <summary>
+    /// Whether to add after or insert before the target, for sorting purposes.
+    /// </summary>
+    public bool PdaAppendAfter { get; set; }
+
+    /// <summary>
+    /// It will be added/inserted next to this item or at the end/beginning if not found.
```
32Kallies, line 65: Just a nitpick with docs: what is "it" here? How about "The icon in the blueprints tab of the PDA will be added next to this item or at the end/beginning if not found." ?

Edit: Or if we decide to go with the whole enum thing, it will have to be written after that is settled.

Nautilus/Assets/Gadgets/ScanningGadget.cs
```diff
@@ -40,6 +40,16 @@ public class ScanningGadget : Gadget
     /// </summary>
     public TechCategory CategoryForPda { get; set; }
 
+    /// <summary>
+    /// Whether to add after or insert before the target, for sorting purposes.
+    /// </summary>
+    public bool PdaAppendAfter { get; set; }
```
32Kallies: Maybe change docs to something along the lines of this:
`Whether the blueprint is added before (false) or after (true) the <see cref="PdaSortTarget"/> in the PDA, when that property is assigned.`
In my mind, PdaAppendAfter = false could imply no sorting done at all. However, that's not the case. I almost think an enum could be fitting here as well... but maybe a bit of bloat for the file? I think it's worth it for the clarity.
```csharp
public enum InsertionPosition {
    AfterTarget,
    BeforeTarget,
    Beginning,
    End
}
```
Thoughts? I think an enum would definitely clear up some ambiguity and possible mistakes we could make down the line.

Nautilus/Handlers/CraftDataHandler.cs
```diff
@@ -27,10 +27,11 @@ public static void AddToGroup(TechGroup group, TechCategory category, TechType t
     /// <param name="group">The TechGroup you want to add your TechType to.</param>
     /// <param name="category">The TechCategory (in the TechGroup) you want to add your TechType to.</param>
     /// <param name="techType">The TechType you want to add.</param>
-    /// <param name="after">Added TechType will be added after this TechType, for sorting purposes.</param>
-    public static void AddToGroup(TechGroup group, TechCategory category, TechType techType, TechType after)
+    /// <param name="after">Whether to add after or insert before the target, for sorting purposes.</param>
+    /// <param name="target">It will be added/inserted next to this item or at the end/beginning if not found.</param>
+    public static void AddToGroup(TechGroup group, TechCategory category, TechType techType, bool after = true, TechType target = TechType.None)
```
32Kallies: Please mark any breaking changes in your PR description.

Nautilus/Patchers/CraftDataPatcher.cs
32Kallies, line 1: This whole class is pretty chunky. I'm hoping you have tested it? Would you have an example mod that uses it by any chance?

Edit: Actually, the logic looks pretty clean, I trust it.

Nautilus/Patchers/CraftDataPatcher.cs
```diff
@@ -35,45 +35,30 @@ internal static void AddToGroup(TechGroup group, TechCategory category, TechType
...
-        }
-
-        if(!techCategory.Contains(techType))
+        if (CraftData.groups.TryGetValue(group, out var techGroup)
+            && techGroup.TryGetValue(category, out var techCategory)
+            && techCategory.Remove(techType))
         {
-            return;
+            InternalLogger.Log($"Successfully Removed \"{techType:G}\" from groups under \"{group:G}->{category:G}\"", LogLevel.Debug);
         }
```
LeeTwentyThree, line 61: Not a request or anything, just wanted to say I'm surprised it looked like that before. Nice cleanup.

Nautilus/Patchers/CraftDataPatcher.cs
```diff
@@ -35,45 +35,30 @@ internal static void AddToGroup(TechGroup group, TechCategory category, TechType
...
+            techCategory.Insert(after ? techCategory.Count : 0, techType);
+            InternalLogger.Log($"{(after ? "Add" : "Insert")}ed \"{techType:G}\" {(after ? "" : "in")}to groups under \"{group:G}->{category:G}\"", LogLevel.Debug);
         }
         else
         {
-            techCategory.Insert(index + 1, techType);
-
-            InternalLogger.Log($"Added \"{techType.AsString():G}\" to groups under \"{group:G}->{category:G}\" after \"{after.AsString():G}\"", LogLevel.Debug);
+            techCategory.Insert(index + (after ? 1 : 0), techType);
+            InternalLogger.Log($"{(after ? "Add" : "Insert")}ed \"{techType:G}\" {(after ? "" : "in")}to groups under \"{group:G}->{category:G}\" {(after ? "after" : "before")} \"{target:G}\"", LogLevel.Debug);
```
32Kallies, line 50: If we go with an enum for the type of insertion, which I'm still considering for the sake of code clarity, I think this part could be cleaned up a bit. As it stands now it's good as it could be though.

32Kallies (changes requested):
Looks good, just a few suggestions/nitpicks

Nautilus/Assets/Gadgets/GadgetExtensions.cs
vlyon: Oops. I've rebased to remove that.

Nautilus/Assets/Gadgets/ScanningGadget.cs
vlyon: It sounds good, but would have to be only 2 entries: `AppendAfter` & `InsertBefore` as the code needs to insert it somewhere when the target is not found, and since the target is `TechType.None` by default, it doesn't make sense to have separate Enums for `Beginning` & `End`.

Nautilus/Handlers/CraftDataHandler.cs
vlyon: This breaking change can actually be avoided by swapping the last two params. Shall I do that?

Nautilus/Assets/Gadgets/ScanningGadget.cs
LeeTwentyThree: Alright, fair. So then it would be AppendAfter by default. Sounds good to me.

LeeTwentyThree (changes requested):

Nautilus/Handlers/CraftDataHandler.cs
32Kallies: I think it would still break, since the optional parameters must stay the same as well.

32Kallies: I'm not sure many people used this override anyway, a breaking change would really be fine. It's just helpful when the breaking changes are listed in the PR description.

32Kallies:
Ready for merge as soon as all of the above requests are resolved.

Nautilus/Assets/Gadgets/ScanningGadget.cs
32Kallies: I accidentally deleted my comment it seems.

Yes, that sounds best.

Nautilus/Handlers/CraftDataHandler.cs
vlyon: By swapping the last two params (both of which are optional) it averts the breaking change while keeping the new functionality.

Nautilus/Assets/Gadgets/ScanningGadget.cs
vlyon: I've added an Enum and some tweaks to documentation.

32Kallies:
Did some testing and everything seems to work as intended. Great improvement :+1:

32Kallies (approved):
I could see issues in the future with mod compatibility, but it's fine.

PR #404, opened 2023-06-11
Better fix for #390 - Use AddGadget, but only when needed

vlyon:
This fixes the duplicate gadget exception by only using AddGadget when needed. (More efficient)


32Kallies:
Personally I preferred it the old way since it had less duplication of the "functional" code and property setting, but it's really personal preference. We can ask other contributors what they prefer.

vlyon:
This way actually has less duplication and less redundant code. It makes use of the return value of `TryGetGadget` instead of ignoring it and then trying to add the gadget again. This way it also avoids setting properties when the newly created gadgets have just been initialised with those properties.

32Kallies:
I already asked ECM and I think she didn't mind either way. I'll ask one other "maintainer" and see what they think.

JKohlman:
Ask and you shall receive an even harder decision, A THIRD OPTION!

```cs
public static CraftingGadget SetRecipe(this ICustomPrefab customPrefab, RecipeData recipeData)
{
    if (!customPrefab.TryGetGadget(out CraftingGadget craftingGadget))
        return customPrefab.AddGadget(new CraftingGadget(customPrefab, recipeData));
    craftingGadget.RecipeData = recipeData;
    return craftingGadget;
}
```

32Kallies:
> Ask and you shall receive an even harder decision, A THIRD OPTION!
> 
> ```cs
> public static CraftingGadget SetRecipe(this ICustomPrefab customPrefab, RecipeData recipeData)
> {
>     if (!customPrefab.TryGetGadget(out CraftingGadget craftingGadget))
>         return customPrefab.AddGadget(new CraftingGadget(customPrefab, recipeData));
>     craftingGadget.RecipeData = recipeData;
>     return craftingGadget;
> }
> ```

I prefer this

32Kallies (changes requested):
I suggest that the code should be reformatted to fit [this suggestion](https://github.com/SubnauticaModding/Nautilus/pull/404#issuecomment-1595807122).

32Kallies (approved):

JKohlman (approved):

PR #429, opened 2023-07-02
[BZ ONLY] Added a better support for Hoverbike modules

VELD-Dev:

  - Added a monobehaviour that is automatically added to the Hoverbike prefab on Hoverbike awake.
  - Added a few patches to fully handle the mono.
  - Added securities so the mono is always added to the Hoverbike mono.
  - I fixed the SeaTruck transpiler btw, it was not required so I just replaced with a postfix. It was crashing the VehicleUpgradesPatcher on BZ but it does not anymore now.
  - Tested module types and approved:
    - Instant modules
    - Chargeable modules
    - Selectable modules
    - SelectableChargeable modules
  - Toggleable modules might still be tried, but theoretically they work perfectly.
  - This time, I checked compatibility with vanilla modules, it's working perfectly.

### Breaking changes

  - None.


Example mod/VehicleUpgradeExample_BelowZero.cs
```diff
@@ -0,0 +1,143 @@
...
+            {
+                new Ingredient(TechType.Polyaniline, 1),
+                new Ingredient(TechType.Quartz, 2),
+                new Ingredient(TechType.Battery, 2),
+                new Ingredient(TechType.AluminumOxide, 1)
+            }
+        };
+        Action<Hoverbike, int> onAdded = (Hoverbike instance, int slotID) =>
+        {
+            if (electricalDefensePrefab is null)
```
32Kallies: Don't use `is` for GameObjects

Example mod/VehicleUpgradeExample_BelowZero.cs
```diff
@@ -0,0 +1,143 @@
...
+        };
+
+        Action<Hoverbike, int> onRemoved = (Hoverbike instance, int slotID) =>
+        {
+            Subtitles.Add("Hoverbike perimeter defense upgrade module disengaged. Caution advised.");
+        };
+
+        Action<Hoverbike, int, float, float> onUsed = (Hoverbike instance, int slotID, float charge, float chargeScalar) =>
+        {
+            if (electricalDefensePrefab is not null)
```
32Kallies: Don't use `is` for GameObjects

32Kallies (changes requested):

Example mod/VehicleUpgradeExample_BelowZero.cs
VELD-Dev: Why ?

Example mod/VehicleUpgradeExample_BelowZero.cs
VELD-Dev: Aight..

Example mod/VehicleUpgradeExample_BelowZero.cs
32Kallies: Sometimes if a GameObject is destroyed, `gameObject is null` on something that used to reference it will still return true.

VELD-Dev: Hmm i see

PR #449, opened 2023-08-20
SML Load Order Issues

Metious:

  - All patches now get executed _after_ SML patches. This is necessary because in the case of clashing (which exists and especially noticeable for craft trees (fabricators)), Nautilus still has the last word and never gets hidden by SML.

This is a big change and does make having both SML and Nautilus installed at the same time a worse experience for SML mods, but Nautilus is now always in favor.

I'm aware that I didn't include some patches in this PR, and those were only postfixes and new features. Both of those aren't possible to clash.

**Please test this PR thoroughly**

A price to pay, but hopefully a worthwhile one.

### Breaking changes

  - Nautilus is now the big boss instead of previously SML.

### DIFF
--- a/Nautilus/Patchers/ConsoleCommandsPatcher.cs
+++ b/Nautilus/Patchers/ConsoleCommandsPatcher.cs
@@ -11,6 +11,7 @@ using UnityEngine;
 
 namespace Nautilus.Patchers;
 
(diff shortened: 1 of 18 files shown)


32Kallies:
No major issues so far, but testing is needed.

PR #455, opened 2023-09-01
MarmosetUBER is initialized too late

Metious:

  - Fixed an issue where the MarmosetUBER is initialized quite a bit late in the stack, resulting in possible race conditions.

I've tested this both in SN1 and BZ and it's working as expected.

Special thanks to UWE, especially Artyom for their insights and assistance in making this fix a possibility.


32Kallies:
Why is `WaitForCompletion` even legal? Either way, nice, glad we could get this sorted out.

32Kallies:
Works as intended

PR #457, opened 2023-09-02
New WithTechType overload and explicit cast from Enum to EnumBuilder<Enum>

Metious:

  - Added a `PrefabInfo.WithTechType` overload that doesn't take a display name and a description. Especially useful for tech types that have the language entries set elsewhere (E.G: localizations)
  - Added an explicit casting operator from Enum to EnumBuilder<Enum>. Can be used to cast a vanilla enum to EnumBuilder, then use the extension methods we provide for each one of them.


32Kallies:
Great changes

PR #463, opened 2023-09-08
Creature Egg hatching fixes & large ACU support

Metious:

  - Fixed eggs throwing on hatch
  - Fixed eggs cant be dropped in large ACUs
  - Added a way to configure the required large ACU floors.


MrPurple6411 (approved):
Have not tested myself however the code itself looks like it should do its job.

PR #469, opened 2023-09-25
Multiple Fixes for the ModPrefabCache

### LINKED ISSUE #394: Issue with duplicate cache items
**Describe the bug**
With some prefabs, several duplicate GameObjects get added to the prefab cache. Seems to happen before a game is even loaded, and maybe with prefabs that have non-matching TechTypes and Class IDs.

**Expected behavior**
One instance max per cache object.

**Screenshots**
image

MrPurple6411:

  - Added a ModPrefabRequest Cache to call upon so that it is not making multiple requests anymore.
  - Null Check ProcessPrefabAsync
  - Add prefab name to error log when missing PrefabIdentifier

### Breaking changes

  - Should be none.

### DIFF
--- a/Nautilus/Assets/ModPrefabCache.cs
+++ b/Nautilus/Assets/ModPrefabCache.cs
@@ -71,6 +71,7 @@ public static class ModPrefabCache
         _cacheInstance = new GameObject("Nautilus.PrefabCache").AddComponent<ModPrefabCacheInstance>();
     }
 }
+
 internal class ModPrefabCacheInstance : MonoBehaviour
 {
     public Dictionary<string, GameObject> Entries { get; } = new Dictionary<string, GameObject>();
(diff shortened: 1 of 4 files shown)


32Kallies:
This seems fine for now, should be ready to merge, but it would probably be best to make a system that uses existing prefab requests for “work in progress” prefabs rather than running multiple requests at the same time.

Nautilus/Assets/ModPrefabCache.cs
```diff
@@ -89,27 +90,24 @@ private void Awake()
...
-        if (!Entries.ContainsKey(prefabIdentifier.classId))
+        // Proper prefabs can never exist in the scene, so parenting them is dangerous and pointless. 
+        if (prefab.IsPrefab())
+        {
+            InternalLogger.Debug($"Game Object: {prefab} is a proper prefab. Skipping parenting for cache.");
+        }
+        else if(!Entries.ContainsKey(prefabIdentifier.classId))
         {
+            prefab.transform.parent = _prefabRoot;
+            prefab.SetActive(true);
```
Metious: While this appears to be much better than before and won't always just add an entry to the cache without much checks, I think clashing and duplicates is still possible, but it's now just "less" likely.

I suggest testing the example mod. that one has a duplicate on the copper clone iirc.

Nautilus/Handlers/PrefabHandler.cs
```diff
@@ -69,7 +76,18 @@ private static IEnumerator ProcessPrefabAsync(GameObject obj, PrefabInfo info, P
         if (Prefabs.TryGetPostProcessorForInfo(info, out var postProcessor))
             yield return postProcessor?.Invoke(obj);
 
-        ModPrefabCache.AddPrefab(obj);
+
+        if (ModPrefabCache.TryGetPrefabFromCache(info.ClassID, out var obj2))
+        {
+            gameObject.Set(obj2);
+            GameObject.Destroy(obj);
```
Metious: This looks dangerous. Since non-instantiated asset bundle prefabs can be custom prefabs, wouldn't destroying a prefab would make the current one also null?

Metious (commented):
Added a few comments

Nautilus/Handlers/PrefabHandler.cs
MrPurple6411: > This seems fine for now, should be ready to merge, but it would probably be best to make a system that uses existing prefab requests for “work in progress” prefabs rather than running multiple requests at the same time.

> This looks dangerous. Since non-instantiated asset bundle prefabs can be custom prefabs, wouldn't destroying a prefab would make the current one also null?

Ohh ooohhhhhhhh   brain explosion..... 

Forgot that addressables Cache their request calls so as not to duplicate them.

I need to add a modPrefabRequest cache and it will fix both of these issues.

Thanx guys.

Metious (approved):
lgtm

PR #470, opened 2023-09-25
Remove custom tech type be unlocked by default

### LINKED ISSUE #467: Remove custom tech type from being unlocked by default
**Describe the feature**
Instead of making the `unlockedAtStart` parameter in the [PrefabInfo.WithTechType](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Assets/PrefabInfo.cs#L54) and [EnumExtensions.WithPdaInfo](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Handlers/Enums/Extensions/EnumExtensions_TechType.cs#L19) set to `true` by default, it should be `false`.

**Purpose**
Rarely a custom prefab is supposed to be unlocked at the start, so that "feature" quickly becomes very annoying.

MrPurple6411:

  - Instead of making the unlockedAtStart parameter in the [PrefabInfo.WithTechType](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Assets/PrefabInfo.cs#L54) and [EnumExtensions.WithPdaInfo](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Handlers/Enums/Extensions/EnumExtensions_TechType.cs#L19) set to true by default, it should be false.

### Breaking changes

  - Unless people specify that things are unlocked they will not be.

### DIFF
--- a/Nautilus/Assets/PrefabInfo.cs
+++ b/Nautilus/Assets/PrefabInfo.cs
@@ -1,4 +1,4 @@
-﻿using System.Reflection;
+using System.Reflection;
 using Nautilus.Handlers;
 using Nautilus.Utility;
 using UnityEngine;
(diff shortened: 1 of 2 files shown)


32Kallies:
Where is the change to PrefabInfo.WithTechType?

Otherwise I heavily agree with the change, although it will obviously be breaking.

32Kallies:
There is also another overload of PrefabInfo.WithTechType.

32Kallies (approved):
Looks good

PR #471, opened 2023-09-25
Update ModCraftTreeLinkingNode.cs

### LINKED ISSUE #319: CraftTreeHandler.AddTabNode issue with tab names
https://github.com/SubnauticaModding/Nautilus-Proto/issues/26

Issue mirrored @RamuneNeptune

MrPurple6411:
The only place I can find for tab nodes to be added without their language lines being set is when you only pass the ID and don't give it any display text at all.

Meaning that if your using that method you need to set the language lines yourself.

  - Add info that ModCraftTreeLinkingNode.AddTabNode(string nameID) does not set language lines 
  - Add check if AddModdedCraftingNode(string moddedTechTypeName) fails to find the moddedTechType

### Breaking changes

  - People will need to make sure they check if a mod is present before just willy nilly adding craft nodes.

### DIFF
--- a/Nautilus/Crafting/ModCraftTreeLinkingNode.cs
+++ b/Nautilus/Crafting/ModCraftTreeLinkingNode.cs
@@ -1,9 +1,7 @@
 ﻿using System.Collections.Generic;
 using Nautilus.Handlers;
-using Nautilus.Utility;
 using UnityEngine;
 using UnityEngine.Assertions;
-using static HandReticle;
 
 namespace Nautilus.Crafting;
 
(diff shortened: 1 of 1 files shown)


Nautilus/Crafting/ModCraftTreeLinkingNode.cs
```diff
@@ -61,6 +61,9 @@ public ModCraftTreeTab AddTabNode(string nameID, string displayText, Sprite spri
     /// <summary>
     /// Creates a new tab node for the crafting tree and links it to the calling node.
     /// </summary>
+    /// <remarks>
+    /// Please not that this method will NOT set the language lines for the node and you must do it yourself.
```
Metious: Minor spelling mistake
```suggestion
    /// Please note that this method will NOT set the language lines for the node and you must do it yourself.
```

Metious (changes requested):

Metious (approved):

PR #473, opened 2023-09-27
ScanningGadget & KnownTech handling

### LINKED ISSUE #402: AnalysisTech & Unlocking
I wasn't sure if this should be a bug report or feature request, as it falls somewhere in the middle.

Most items are either unlocked by default or require one or more _discoveries_ to unlock them.
Discoveries are made by either picking up or scanning an item or triggering a story goal.

Most items unlock nothing! This means that when you pickup or scan that item, it seems to unlock, by playing the sound and showing a message even though it actually hasn't unlocked anything (except itself). This could use the `NotificationBlueprintUnlocked` or `NotificationCreatureDiscovered` message.
Example: Lubricant unlocks only itself with a `NotificationBlueprintUnlocked` message.

Some items will unlock other blueprints. Like when a fish is picked up, it unlocks the cooked variants. This could use the `NotificationBlueprintPickup` , `NotificationCreatureDiscovered` or `NotificationBlueprintUnlocked` message.
Example: CreepvineSeedCluster unlocks Lubricant with a `NotificationBlueprintPickup` message.

However some items have _both_ entries. This is currently difficult to implement in mods. Currently this is because it only holds one `AnalysisTech` entry per item. We should be able to add both types to an item.
As the examples above show, Lubricant is unlocked by a CreepvineSeedCluster and unlocks itself.

To make matters worse the logic to determine if an item should be unlocked by default is not working correctly.

### Suggestion
1. Could we have `CustomPrefab.SetUnlockData` to set the `AnalysisTech` entry for this item. Along with an optional `.AlsoUnlocks` to add a list of other items to unlock when this item is unlocked.
2. Could we have `CustomPrefab.SetUnlockedBy` to set the `AnalysisTech` or `CompoundTech` entry for some other item(s) that unlock this item. Along with an optional `.WithScannerEntry` as usual.
3. Then we could implement better logic to find all items that need unlocking and remove them from `UnlockedAtStart`.

### LINKED ISSUE #466: ScanningGadget is written incorrectly (AnalysisTech & ScannerEntry)
**Describe the bug**
Currently, the `RequiredForUnlock` and `AnalysisTech` are used and expected on the same item (tech type), and overwrite the `RequiredForUnlock`'s analysis tech to include the current item's tech type. This works as long as the `RequiredForUnlock` already has analysis tech setup (E.G: vanilla items), but for custom items, this algorithm is extremely buggy.

**To Reproduce**
1. Create two custom prefabs
2. Set the unlock requirement of the second prefab to the first one using `SetUnlock` (this step doesn't matter as long as it's the ScanningGadget)
3. Call the `WithAnalysisTech` on the second prefab to add an analysis tech entry to make it popup once unlocked
4. The analysis tech is added to the first prefab and displays the first prefab's name as the blueprint that got unlocked

**Expected behavior**
AnalysisTech _have_ to be on the second prefab, which is where we called the `WithAnalysisTech` method. The only way we edit the requirement techtype's entry is only to add our prefab's tech type to the said requirement so it's also unlocked when the requirement is unlocked. That's about it.

**How to achieve**
1. Don't treat `RequiredForUnlock` and `AnalysisTech` as the same thing
  - When the `WithAnalysisTech` is called, always set the key to the current prefab's techtype and never to the requirement's techtype.
  - `RequiredForUnlock` _always only_ adds the current prefab's techtype to the requirement's analysis tech entry's `unlockTechTypes` list. If the requirement doesn't already have an analysis tech, we throw an error.
2. Add another scanner-entry-based overload that takes the blueprint tech type as a parameter and assign the scanner entry's key to the current tech type. This way, for fragments you just add the scanner entry to that fragment and assign the unlocked blueprint rather than overwriting the fragment's scanner entry in the blueprint's scanning gadget.

Metious:

resolves #402 
fixes #466

  - Improved the `KnownTechPatcher` implementation; special thanks to @vlyon for the suggestion in #402 
  - Fixed AddAnalysisTech assignment being flipped, results in overwriting values never being implemented for existing entries
  - Added `AddRequirementForUnlock` method that just adds blueprints to a requirement tech type
  - Added a better `WithScannerEntry` overload for the `ScanningGadget` that doesn't override a techtype's scanner entry that it has no business in
  - `ScanningGadget` now always adds analysis tech to the current tech type, when the `WithAnalysisTech` method is called
  - Fixed log spams for buildables
  - Added hard-locking to make items not unlockable by default or by the `unlockall` command
  - Added a cross-game-friendly overload for `AnalysisTech`.

### Breaking changes

  - You can no longer set requirements for tech types that don't have an Analysis tech

### Deprecated methods
  - [ScanningGadget.WithScannerEntry](https://github.com/Metious/Nautilus/blob/95d47bb8423450bfd5f3b7b5a372abc849685af9/Nautilus/Assets/Gadgets/ScanningGadget.cs#L253)
  
### TODO:
- [x] Decide on how to implement the `KnownTechPatcher.RemovalTechs` efficiently

### DIFF
--- a/Nautilus/Assets/Gadgets/CraftingGadget.cs
+++ b/Nautilus/Assets/Gadgets/CraftingGadget.cs
@@ -89,11 +89,8 @@ public class CraftingGadget : Gadget
         
         CraftDataHandler.SetRecipeData(prefab.Info.TechType, RecipeData);
         
-        if (FabricatorType == CraftTree.Type.None)
-        {
-            InternalLogger.Log($"Prefab '{prefab.Info.ClassID}' was not automatically registered into a crafting tree.");
-        }
-        else
+        
+        if (FabricatorType is not CraftTree.Type.None)
         {
             if (StepsToFabricatorTab == null || StepsToFabricatorTab.Length == 0)
             {
@@ -103,6 +100,11 @@ public class CraftingGadget : Gadget
             {
                 CraftTreeHandler.AddCraftingNode(FabricatorType, prefab.Info.TechType, StepsToFabricatorTab);
             }
+            
+        }
+        else if (!prefab.TryGetGadget(out ScanningGadget scanningGadget) || !scanningGadget.IsBuildable)
+        {
+            InternalLogger.Log($"Prefab '{prefab.Info.ClassID}' was not automatically registered into a crafting tree.");
         }
 
         if (CraftingTime >= 0f)
--- a/Nautilus/Patchers/KnownTechPatcher.cs
+++ b/Nautilus/Patchers/KnownTechPatcher.cs
@@ -1,11 +1,10 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Linq;
-using HarmonyLib;
-using Nautilus.Handlers;
-using Nautilus.Utility;
+﻿namespace Nautilus.Patchers;
 
-namespace Nautilus.Patchers;
+using System;
+using System.Collections.Generic;
+using HarmonyLib;
+using Handlers;
+using Utility;
 
 internal class KnownTechPatcher
 {
@@ -31,16 +30,20 @@ internal class KnownTechPatcher
     internal static void Reinitialize()
     {
         // At this point the KnownTech stuff haven't been initialized yet, so no need to re-patch.
-        if (!Player.main)
+        if (!KnownTech.initialized)
         {
             return;
         }
 
-        var pdaData = Player.main.pdaData;
-        InitializePrefix(pdaData);
-        KnownTech.AddRange(pdaData.defaultTech, false);
+        var knownTech = KnownTech.knownTech;
+        var knownCompound = KnownTech.knownCompoundTech;
+        var pdaData = UnityEngine.Object.Instantiate(Player.main.pdaData);
+        KnownTech.Initialize(pdaData);
         KnownTech.compoundTech = KnownTech.ValidateCompoundTech(pdaData.compoundTech);
         KnownTech.analysisTech = KnownTech.ValidateAnalysisTech(pdaData.analysisTech);
+        KnownTech.knownTech.AddRange(knownTech);
+        knownCompound.ForEach(x => knownCompound.Add(x.Key, x.Value));
+        KnownTech.AddRange(pdaData.defaultTech, false);
     }
 
     private static void InitializePrefix(PDAData data)
--- a/Nautilus/Patchers/KnownTechPatcher.cs
+++ b/Nautilus/Patchers/KnownTechPatcher.cs
@@ -75,7 +75,7 @@ internal class KnownTechPatcher
             var index = data.analysisTech.FindIndex(tech => tech.techType == blueprintRequirements.Key);
             if (index == -1)
             {
-                InternalLogger.Error($"TechType '{blueprintRequirements.Key.AsString()}' does not have an analysis tech. Cancelling requirement addition.");
+                InternalLogger.Error($"TechType '{blueprintRequirements.Key.AsString()}' does not have an analysis tech. Cancelling requirement addition for TechTypes '{blueprintRequirements.Value.Join()}'.");
                 continue;
             }
             
--- a/Nautilus/Patchers/KnownTechPatcher.cs
+++ b/Nautilus/Patchers/KnownTechPatcher.cs
@@ -13,11 +13,11 @@ internal class KnownTechPatcher
 
     internal static HashSet<TechType> UnlockedAtStart = new();
     internal static HashSet<TechType> LockedWithNoUnlocks = new();
+    internal static HashSet<TechType> RemovalTechs = new();
     internal static IDictionary<TechType, KnownTech.AnalysisTech> AnalysisTech = new SelfCheckingDictionary<TechType, KnownTech.AnalysisTech>("AnalysisTech", AsStringFunction);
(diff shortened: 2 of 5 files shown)


Metious:
The `KnownTechPatcher.RemovalTechs` list is only used once and only for compound techs. I'd like to know why are we removing compound techs from other unlocks and if it is necessary anymore

Nautilus/Patchers/KnownTechPatcher.cs
MrPurple6411, line 1: A few notes on what I am seeing here.

1. This will UTTERLY break Unknown Name Mod if Reinitialize is called even once because you're effectively redoing everything over WITHOUT actually calling the method the game calls to do the Initialize and instead straight calling ONLY your patch.

2. This will end up with MANY Duplicates in the defaulttech as every time it is reinitialized you are addrange without reseting it to empty.

this is just a very quick look into these changes which I will continue to check and test using UnknownName mod to test it with.

So far, I would like the reinitialize to call the proper game method so that any mods that patch this will be able to make their changes too.

I will also try and see if I can modify my mod to better work with this new change but at the VERY least, I think this is only partially done.

Relevent Unknown Name Patch https://github.com/SubnauticaModding/Purples-Mods/blob/development/UnkownName/Patches/KnownTech_Initialize.cs

Maybe use the KnownTech.Deinitialize and then call KnownTech.Initialize(PDAData data) instead of just doing it over in the Reinitialize when theirs already does it and more so as not to have it duplicating data and allow for other mods to patch in.

image


I don't think we even need the Player.main check in the if as the Initialized check should be sufficient.

so it could be as simple as: 
```cs

    internal static void Reinitialize()
    {
        // At this point the KnownTech stuff haven't been initialized yet, so no need to re-patch.
        if (!KnownTech.initialized)
            return;

        KnownTech.Deinitialize();
        KnownTech.Initialize(UnityEngine.Object.Instantiate<PDAData>(Player.main.pdaData));
    }
```

Also yes, the Instantiate is necessary as that is actually how the game itself does it so as not to corrupt the default one on the player prefab.

I mean we could leave the null check just for extra safety but the KnownTech by default games code CANNOT be initialized without the Player as its Player.Awake that triggers these methods in the first place.

MrPurple6411 (changes requested):

Nautilus/Patchers/KnownTechPatcher.cs
Metious: 1. The reason why I'm not calling the game's Initialize directly is because that method resets everything, including the existing unlocked list and whatnot. So if we do call `KnownTech.Initialize`, the world will be left with only modded entries or none. I think we could probably work around this, but I thought leaving it as-is would be harmless. It's worth mentioning that since the UnknownName mod is coupled on the `KnownTech.Initialize`, it's already (before this PR) not working properly as we still aren't calling the method directly in Nautilus for the aforementioned reasons.
2. Valid. I forgot that the data's defaultTech is a List and not a HashSet, so even though they'll eventually _not_ be duplicates, the initial collection will grow and potentially cause some ridiculous lags.

> Also yes, the Instantiate is necessary as that is actually how the game itself does it so as not to corrupt the default one on the player prefab.

The game doesn't instantiate the PDAData at all and interact with it directly. It's harmless for us to also use it directly as long as we're not stacking changes over and over again.

Overall very good points and I will consider them.

MrPurple6411: > The game doesn't instantiate the PDAData at all and interact with it directly. It's harmless for us to also use it directly as long as we're not stacking changes over and over again.

image

MrPurple6411: All the modifications done in these initialize methods are done to a copy :-)

Metious: Well, didn't know that. Interesting

MrPurple6411 (approved):
It looks right and doesn't seem to break any of my mods from the small testing I did.

And while my gut still feels like somethings wrong here it seems to work, and I can't figure out what my gut is trying to tell me so.......

Nautilus/Assets/Gadgets/CraftingGadget.cs
```diff
@@ -103,6 +100,11 @@ protected internal override void Build()
             {
                 CraftTreeHandler.AddCraftingNode(FabricatorType, prefab.Info.TechType, StepsToFabricatorTab);
             }
+            
+        }
+        else if (!prefab.TryGetGadget(out ScanningGadget scanningGadget) || !scanningGadget.IsBuildable)
+        {
+            InternalLogger.Log($"Prefab '{prefab.Info.ClassID}' was not automatically registered into a crafting tree.");
```
MrPurple6411, line 107: Think we could make this Debug log instead?

MrPurple6411 (approved):
either way I think this is good enough. 

Just wondering about setting that 1 log to Debug.   if not then good, if yes then good, either way get this merged already lol

Nautilus/Assets/Gadgets/CraftingGadget.cs
Metious: I think marking that as a debug log will make it easy to miss because a craftable item without a crafting tree that isn't a buildable is alarming and it's safer to assume that the developer missed something. In fact, I'd even argue it should be a warning instead.

MrPurple6411 (approved):
LGTM

PR #476, opened 2023-09-27
StoryGoal actions receive the name of the unlocked goal

### LINKED ISSUE #475: StoryGoal Custom Events do not receive name of triggering goal
**Describe the bug**
When registering a callback for a StoryGoal, the documentation notes that the callback is meant to receive the name of the goal when it is invoked. This enables reusing the same method for multiple custom events. However, the method signature requires an `Action` rather than an `Action<string>` and the event is similarly invoked without the goal name.

**To Reproduce**
1. See [RegisterCustomEvent()](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Handlers/StoryGoalHandler.cs#L140) and [NotifyGoalComplete()](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/MonoBehaviours/CustomStoryGoalManager.cs#L71)
2. Try using an `Action<string>` anyway
3. Be sad as the compiler yells at you

**Expected behavior**
Get the goal name on unlock.

tinyhoot:

  - Custom events for StoryGoals behave as the documentation claims and now receive the name of the calling goal as an argument. Fix for #475 

### Breaking changes

  - None

### DIFF
--- a/Nautilus/Handlers/StoryGoalHandler.cs
+++ b/Nautilus/Handlers/StoryGoalHandler.cs
@@ -137,10 +137,19 @@ public static class StoryGoalHandler
     /// Registers a given <see cref="Action"/> to be performed when its associated goal is completed.
     /// </summary>
     /// <param name="key">The key of the goal that triggers the <paramref name="customEventCallback"/>.</param>
-    /// <param name="customEventCallback">The method that is called when the associated goal is completed. The name of the goal will be passed as a parameter.</param>
+    /// <param name="customEventCallback">The method that is called when the associated goal is completed.</param>
     public static void RegisterCustomEvent(string key, Action customEventCallback)
     {
         CustomStoryGoalManager.StoryGoalCustomEvents.GetOrAddNew(key).Add(customEventCallback);
+    }
+
+    /// <summary>
+    /// Registers a given <see cref="Action"/> to be performed when a goal is completed.
+    /// </summary>
+    /// <param name="customEventCallback">The method that is called when any goal is completed. The name of the goal will be passed as a parameter.</param>
+    public static void RegisterCustomEvent(Action<string> customEventCallback)
+    {
+        CustomStoryGoalManager.StoryGoalCustomKeyHandlerEvents.Add(customEventCallback);
     }
 
     /// <summary>
@@ -154,5 +163,14 @@ public static class StoryGoalHandler
         {
             callbacks.Remove(customEventCallback);
         }
+    }
+    
+    /// <summary>
+    /// Unregisters a custom event.
+    /// </summary>
+    /// <param name="customEventCallback">The method to unregister.</param>
+    public static void UnregisterCustomEvent(Action<string> customEventCallback)
+    {
+        CustomStoryGoalManager.StoryGoalCustomKeyHandlerEvents.Remove(customEventCallback);
     }
 }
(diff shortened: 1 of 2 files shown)


Metious (changes requested):
As discussed [in the issue](https://github.com/SubnauticaModding/Nautilus/issues/475#issuecomment-1738040840), revert the breaking changes to the current methods, and just add two other overloads that only takes an `Action<string>`.

tinyhoot:
_Only_ an `Action<string>`? Without a key, any such method would receive the custom events of all other mods too.

Nautilus/MonoBehaviours/CustomStoryGoalManager.cs
```diff
@@ -9,7 +9,7 @@ internal class CustomStoryGoalManager : MonoBehaviour, IStoryGoalListener
 {
     public static CustomStoryGoalManager Instance { get; private set; }
     
-    internal static readonly Dictionary<string, List<Action>> StoryGoalCustomEvents = new();
+    internal static readonly Dictionary<string, List<Action<string>>> StoryGoalCustomEvents = new();
```
MrPurple6411:
```suggestion
    internal static readonly Dictionary<string, List<Action>> StoryGoalCustomEvents = new();
    internal static readonly List<Action<string>> StoryGoalCustomKeyHandlerEvents = new();
```

Nautilus/MonoBehaviours/CustomStoryGoalManager.cs
```diff
@@ -68,7 +68,7 @@ void IStoryGoalListener.NotifyGoalComplete(string key)
         foreach (var customEvent in StoryGoalCustomEvents)
```
MrPurple6411:
```suggestion
        StoryGoalCustomKeyHandlerEvents.ForEach(x => x?.Invoke(key));

        foreach (var customEvent in StoryGoalCustomEvents)
```

Nautilus/MonoBehaviours/CustomStoryGoalManager.cs
```diff
@@ -68,7 +68,7 @@ void IStoryGoalListener.NotifyGoalComplete(string key)
...
         {
             if (key == customEvent.Key)
-                customEvent.Value.ForEach(x => x?.Invoke());
+                customEvent.Value.ForEach(x => x?.Invoke(key));
```
MrPurple6411:
```suggestion
                customEvent.Value.ForEach(x => x?.Invoke());
```

Nautilus/Handlers/StoryGoalHandler.cs
MrPurple6411, line 1: Revert these changes and add 2 more instead that add/remove to/from the new List suggested in the other file.

MrPurple6411 (changes requested):
Think these are the changes @Metious is meaning. 

Hopefully he will clarify if I am off base.

Metious:
> _Only_ an `Action<string>`? Without a key, any such method would receive the custom events of all other mods too.

Isn't that the point? Perhaps I'm simply not familiar with your use-case, but I think it's useless to assign a key while also taking the key as a parameter in the delegate.

If the method only takes an `Action<string>`, you should be able to handle multiple goals in the same method. This is also the closest implementation to the vanilla custom goals implementation, so it's fine for those delegates to be callbacked on every goal.

It's a feature, not a bug :)

tinyhoot:
> It's a feature, not a bug :)

That's a good point! I've incorporated both of your suggested changes into this. The documentation for the method which only takes an `Action` also no longer claims that a parameter will be passed.

Please let me know if there's anything else I can adjust.

MrPurple6411:
hrmmm @Metious Do we want the keyhandlers to run before or after the regular ones.

Metious (approved):
Lgtm.

MrPurple6411 (approved):

PR #483, opened 2023-10-02
Basic Patcher to force Nautilus to be the first Plugin loaded.

### LINKED ISSUE #477: ModID extra info not showing correctly.
**Describe the bug**
When using Nautilus to show modid on the tooltips it is showing the number (number) of the techtype instead of name (number)

**To Reproduce**
Turn on mod name and itemid in options menu.

**Expected behavior**
Techtype string and number are shown

**Screenshots**
image

MrPurple6411:

  - Created a Bepinex Patcher that forces Nautilus to load first to fix the issue where Enums are failing when people forget to set Nautilus as a dependency.

### DIFF
--- a/Nautilus.Priority/Initializer.cs
+++ b/Nautilus.Priority/Initializer.cs
@@ -1,4 +1,4 @@
-namespace NautilusFirst;
+namespace Nautilus.Priority;
 
 using BepInEx;
 using HarmonyLib;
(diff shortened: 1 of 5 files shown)


Metious (changes requested):
Looks good to me. Just rename the project to `Nautilus.Priority` or `Nautilus.PriorityPatcher` to adhere with the prefix we own (Nautilus.*).

Also be sure to include the assembly to the zip file.

MrPurple6411:
> Also be sure to include the assembly to the zip file.

Hrmm ho would we like to do that?

MrPurple6411:
So was talking in ⁠with tobey and he made me realize that this Nautilus first is actually kind of a bad idea because it promotes laziness as anyone having this issue is having it because they forgot to put the dependency attribute.

And anyone who is having said issue due using SMLHelper is also self-inflicting it.

PR #486, opened 2023-10-04
Workbench Organization.

MrPurple6411:

- Move all Crafting Nodes out of the Workbench Root
- Added a Fallback Node when adding craftnodes and they fail due to mixing tabs and crafting nodes. 
- Created a TraverseTree method as we seem to be doing the same thing in multiple places. 
- Added logging to ModCraftTreeRoot to show what nodes are in the way when trying to add a node fails due to cross node type.

### DIFF
--- a/Nautilus/Patchers/CraftTreePatcher.cs
+++ b/Nautilus/Patchers/CraftTreePatcher.cs
@@ -71,20 +71,26 @@ internal class CraftTreePatcher
 
     private static void CreateVanillaTabNode(CraftTree.Type treeType, string DisplayName, TechType spriteTechType, TreeNode root)
     {
-        CraftTreeHandler.AddTabNode(treeType, VanillaRoot, DisplayName, SpriteManager.Get(spriteTechType));
-
         var removedNodes = new List<CraftNode>();
         foreach (var node in root.nodes.Cast<CraftNode>())
         {
+            if (node.action == TreeAction.Expand)
+                continue;
+
             CraftTreeHandler.RemoveNode(treeType, new[] { node.id });
             removedNodes.Add(node);
         }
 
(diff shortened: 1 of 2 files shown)


Nautilus/Patchers/CraftTreePatcher.cs
```diff
@@ -16,18 +18,81 @@ internal class CraftTreePatcher
...
     {
+        CreateFallbackNodes();
         harmony.PatchAll(typeof(CraftTreePatcher));
-
         InternalLogger.Log($"CraftTreePatcher is done.", LogLevel.Debug);
     }
 
+    private static void CreateFallbackNodes()
+    {
+        Enum.GetValues(typeof(CraftTree.Type)).Cast<CraftTree.Type>().ToList().ForEach(x =>
```
Metious: Iteration here seems useless because you're still only making changes to specific craft tree types and ignoring the rest. Maybe just make the change directly without iterating?
```cs
// Workbench
CreateVanillaTabNode(CraftTree.Type.Workbench, "Modification Station", TechType.Workbench, CraftTree.WorkbenchScheme().root);
CraftTreeHandler.AddTabNode(CraftTree.Type.Workbench, FallbackTabNode, "Unsorted Mod Items", SpriteManager.Get(TechType.Workbench));

// Fabricator
CraftTreeHandler.AddTabNode(CraftTree.Type.Fabricator, FallbackTabNode, "Unsorted Mod Items", SpriteManager.Get(TechType.Fabricator));

// ...

Metious (changes requested):
Added a comment on something that looked a bit strange to me.
I've tested this PR in-game and it seems to be working fine.

Metious (approved):

PR #489, opened 2023-10-12
Extend the CloneTemplate

MrPurple6411:
  - Add Overload to pass a AssetReferenceGameObject to clone by GUID or Asset Path.

### Breaking changes
  - None

### DIFF
--- a/Nautilus/Assets/PrefabTemplates/CloneTemplate.cs
+++ b/Nautilus/Assets/PrefabTemplates/CloneTemplate.cs
@@ -60,7 +60,10 @@ public class CloneTemplate : PrefabTemplate
         _techTypeToClone = techTypeToClone;
         _classIdToClone = classIdToClone;
         _prefabToClone = prefabToClone;
-        _spawnType = techTypeToClone != TechType.None ? SpawnType.TechType : string.IsNullOrWhiteSpace(classIdToClone) ? prefabToClone != null ? SpawnType.Prefab : throw new System.Exception("CloneTemplate Missing valid identifier.") : SpawnType.ClassId;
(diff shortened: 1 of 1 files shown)


Nautilus/Assets/PrefabTemplates/CloneTemplate.cs
```diff
@@ -35,24 +37,30 @@ public class CloneTemplate : PrefabTemplate
...
     {
         _techTypeToClone = techTypeToClone;
         _classIdToClone = classIdToClone;
-        _spawnType = techTypeToClone switch
-        {
-            default(TechType) => SpawnType.ClassId,
-            _ => SpawnType.TechType
-        };
+        _prefabToClone = prefabToClone;
+        _spawnType = techTypeToClone != TechType.None ? SpawnType.TechType : string.IsNullOrWhiteSpace(classIdToClone) ? prefabToClone != null ? SpawnType.Prefab : throw new System.Exception("CloneTemplate Missing valid identifier.") : SpawnType.ClassId;
```
Metious: This is very hard to read, I suggest simplifying it and making it read like a stack instead:
```suggestion
        _spawnType = techTypeToClone is not TechType.None ? SpawnType.TechType
            : !string.IsNullOrWhiteSpace(classIdToClone) ? SpawnType.ClassId
            : prefabToClone is not null ? SpawnType.Prefab
            : throw new System.Exception("CloneTemplate Missing valid identifier.");
```

Metious (changes requested):
I added a suggestion. Other than that looks good to me.

Also this PR has no breaking changes as the changed stuff were all internal. Even if the `SpawnType` enum was public, adding a new one wouldn't break anything but custom enums for this type, which are unlikely.

Metious (approved):

PR #493, opened 2023-10-24
Mention marking Nautilus as a dependency in the docs

Metious:

  - Now we mention marking Nautilus as a dependency wherever necessary.


MrPurple6411 (approved):
While these all LGTM I think we could also use a whole page about BepInDependency OR add in a pointer to Bepinex's own documentation on it.

PR #503, opened 2023-11-16
UploadHelper Overhaul/Fixing BepInExPluginVersion

MrPurple6411:

  - Setup BepinexPluginVersion to use the new SuffixNumber
image

  - Overhauled Upload Helper
    - Added a lot of Ask calls to bypass things that may not need to be done.
    - Automated projects rebuild with NuGet restore after version number changes.
    - Ask permission more often instead of forcing opening folders and websites.
    - Append BepinexPluginVersion to end of zip file.
    - Upload helper will remove VersionSuffix from Version.targets and set the SuffixNumber to 0 if no pre-release suffix is used.
    
image

image

    
image

### DIFF
--- a/Example mod/Example mod.csproj
+++ b/Example mod/Example mod.csproj
@@ -1,47 +1,48 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
-  <Import Project="..\common.props" />
-  <PropertyGroup>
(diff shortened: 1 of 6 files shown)


32Kallies (approved):
Works as intended. Very convenient too.

PR #505, opened 2023-11-24
Allow nullables and params arrays in console commands

Govorunb:
#### Console commands
  - Allow using nullable value types in parameters
    - A command like `ToggleSomething(bool? enable = null)` can now be registered and can e.g. print the current status of the toggle if called with no arguments
  - Allow using a `params T[]` parameter to consume the rest of the input
    - This allows commands to parse an arbitrary number of (still primitively-typed) parameters
    - Something simple like a command that repeats your input back at you is now extremely easy to define
    - It only makes sense to have one `params` array and only as the last parameter - which is enforced by the compiler

### Breaking changes
  - No breaking changes to existing commands from my testing

### DIFF
--- a/Nautilus/Commands/ConsoleCommand.cs
+++ b/Nautilus/Commands/ConsoleCommand.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Reflection;
 using HarmonyLib;
+using Nautilus.Extensions;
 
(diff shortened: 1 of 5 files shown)


Govorunb:

Alexejhero (approved):
Dg0AnDCW0AA741R

32Kallies (approved):
Very useful, I also appreciate the friendly type names 👍

PR #523, opened 2024-01-02
Support protobuf serializer for custom types

tornac1234:

  - Add the possibility to serialize our own types through the protobuf serializer
  - Replace ProtobufSerializerPrecompiled.Serialize and ProtobufSerializerPrecompiled.Deserialize methods which used a big if-tree to have a dictionary lookup instead
  
As I don't own BZ, someone else will need to adapt this code for it.

### DIFF
--- a/Nautilus/Patchers/ProtobufSerializerPrecompiledPatcher.cs
+++ b/Nautilus/Patchers/ProtobufSerializerPrecompiledPatcher.cs
@@ -1,9 +1,7 @@
 using System;
-using System.Collections.Generic;
-using System.Linq;
 using System.Reflection;
-using System.Reflection.Emit;
 using HarmonyLib;
+using Nautilus.Handlers;
 using Nautilus.Utility;
 using ProtoBuf;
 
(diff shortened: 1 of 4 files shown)


tornac1234:

tornac1234:
recheck

Nautilus/Patchers/ProtobufSerializerPrecompiledPatcher.cs
```diff
@@ -0,0 +1,173 @@
...
+    /// </param>
+    /// <param name="serializeMethod">
+    /// A reference to the serialize method.
+    /// NB: Take inspiration from <see cref="ProtobufSerializerPrecompiled.Serialize11492366"/> to understand how make one.
+    /// </param>
+    /// <param name="deserializeMethod">
+    /// A reference to the serialize method.
+    /// NB: Take inspiration from <see cref="ProtobufSerializerPrecompiled.Deserialize11492366"/> to understand how make one.
+    /// </param>
+    public static void RegisterSerializableType<T>(int typeId, Action<T, int, ProtoWriter> serializeMethod, Func<T, ProtoReader, T> deserializeMethod)
```
Metious: I believe you're missing a handler class. This method is made public, but the patcher class is internal, which makes the method unusable by mods.

The way we have such things setup in Nautilus is by having a public handler under the `Handlers` namespace that subscribes to the patch(es).

Example: [LanguagePatcher](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Patchers/LanguagePatcher.cs#L97) & [LanguageHander](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Handlers/LanguageHandler.cs#L22).

Metious (changes requested):
Overall it looks pretty good. Though I believe we should try to make the serializing callbacks auto-generated since they shouldn't contain any custom code.

This is a very nice implementation. Nicely done

32Kallies:
It's nearing 3 years. @Metious, do you have thoughts on this PR? Or perhaps plans for a more modern version? I believe you were the last person to look into it.

PR #531, opened 2024-01-14
Coordinated Spawns (Purple Edition)

MrPurple6411:
  - Moved spawn creation to OnBatchFullyLoaded patch
  - Preload unspawned prefabs during initialization postfix to ensure speedy spawning later.

### DIFF
--- a/Nautilus/MonoBehaviours/EntitySpawner.cs
+++ b/Nautilus/MonoBehaviours/EntitySpawner.cs
@@ -1,24 +1,24 @@
+namespace Nautilus.MonoBehaviours;
+
 using System.Collections;
 using System.Collections.Generic;
-using System.Linq;
-using Nautilus.Extensions;
 using Nautilus.Handlers;
 using Nautilus.Patchers;
 using Nautilus.Utility;
 using UnityEngine;
 using UWE;
 
-namespace Nautilus.MonoBehaviours;
-
 internal class EntitySpawner : MonoBehaviour
 {
     internal Int3 batchId;
     internal IReadOnlyCollection<SpawnInfo> spawnInfos;
     internal bool global;
+    private List<SpawnInfo> delayedSpawns = new List<SpawnInfo>();
 
     private IEnumerator Start()
     {
         yield return SpawnAsync();
+        yield return new WaitUntil(() => delayedSpawns.Count == 0);
         Destroy(gameObject);
     }
 
@@ -32,10 +32,7 @@ internal class EntitySpawner : MonoBehaviour
             // then we wait until the terrain is fully loaded (must be checked on each frame for faster spawns)
             yield return new WaitUntil(() => lws.IsBatchReadyToCompile(batchId));
         }
-        
-        var batchCenter = lws.GetBatchCenter(batchId);
-        var bounds = new Bounds(batchCenter, Vector3.zero);
-        
+
         LargeWorld lw = LargeWorld.main;
         
         yield return new WaitUntil(() => lw != null && lw.streamer.globalRoot != null); // need to make sure global root is ready too for global spawns.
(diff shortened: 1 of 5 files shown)


Nautilus/MonoBehaviours/EntitySpawner.cs
```diff
@@ -35,40 +35,18 @@ IEnumerator SpawnAsync()
...
-            Int3 batch = lws.GetContainingBatch(spawnInfo.SpawnPosition);
-            yield return new WaitUntil(() => lws.IsBatchReadyToCompile(batch)); // then we wait until the terrain is fully loaded (must be checked on each frame for faster spawns)
-        }
-
-        LargeWorld lw = LargeWorld.main;
-
-        yield return new WaitUntil(() => lw != null && lw.streamer.globalRoot != null); // need to make sure global root is ready too for global spawns.
-            
-        lw.streamer.cellManager.RegisterEntity(obj);
+        if (lwe != null)
```
Metious: Don't make having a `LargeWorldEntity` optional because the game has no idea how to save it and we cant assume that they're transient.

If an object is missing an LWE component, it's most likely that the user forgot to add it, so it's better to let them know about than to have an arbitrary fallback.

Metious (changes requested):
Testing this pull request, I'm now getting an NRE first that goes like the following:
```
[Error  : Unity Log] NullReferenceException
Stack trace:
CellManager.RegisterCellEntity (LargeWorldEntity ent) (at <9ec6c1b62c324758b9c2d79021505be6>:0)
CellManager.RegisterEntity (LargeWorldEntity lwe) (at <9ec6c1b62c324758b9c2d79021505be6>:0)
Nautilus.MonoBehaviours.EntitySpawner+<>c__DisplayClass2_0.<SpawnAsync>b__0 () (at <c13f3a3dfa5c494f9fd922a160750b0b>:0)
UnityEngine.WaitUntil.get_keepWaiting () (at <bd0d47c27bd84106afaaecc2c74cdb94>:0)
UnityEngine.CustomYieldInstruction.MoveNext () (at <bd0d47c27bd84106afaaecc2c74cdb94>:0)
UnityEngine.SetupCoroutine.InvokeMoveNext (System.Collections.IEnumerator enumerator, System.IntPtr returnValueAddress) (at <bd0d47c27bd84106afaaecc2c74cdb94>:0)
```
Then I end up with my object infinitely spawning.
image

This PR is not done yet. I'm converting it to a draft to prevent accidental merges.

Metious (approved):
Works fine on my tests

Nautilus/MonoBehaviours/EntitySpawner.cs
```diff
@@ -1,74 +1,106 @@
...
+                continue;
+                // 😎 Nice.
+            }
+
+            if (lwe.cellLevel < LargeWorldEntity.CellLevel.Batch && !lws.IsRangeActiveAndBuilt(new Bounds(spawnInfo.SpawnPosition, Vector3.one)))
+            {
+                // Cells aren't ready yet. We have to wait until they are.
+                StartCoroutine(WaitForCellLoaded(lws, prefab, spawnInfo, stringToLog));
+                continue;
+            }
```
MrPurple6411: @Metious can I get you to check this part out because the game itself Ensures a LWE on any registered object and just uses the default cell level which would be near I think

Metious: It always has been the case, but I dislike it because we're assuming cell levels, which are quite important on how the object loads/saves. I think since we always had them error before, it's better to keep it that way so users still get notified for possible mistakes and also backward compatibility.

Also `LargeWorldStreamer.IsRangeActiveAndBuilt` depends on the clipmap being ready, which is extremely slow and kinda broken especially in the void. In my experience it always returned false no matter how much I waited in the spawning zone, but it only returns true when I teleport to the exact spawn position.
I was using that before switching to just checking for the cells being ready; in my tests in SN cells are almost immediately initialized after the batch is ready for compilation, so if BZ doesn't have the method, they probably don't need them anymore due to that, so maybe keep the cell check for SN only?

MrPurple6411: Before you actually had a zero sized bounds for some reason and were using it based on the batch center.

I converted this to actually use the same code the game itself uses to check the location
By checking a vector3.one×5 at the spawn location instead of the batch Location. 

I will do some testing for void spawns with my way to see.

Metious: I changed it to batch center only after making the entity spawner per-batch. During my tests I was using the spawn info's position and checked it against 0 size, 3 size and 5 size. All of them returned false when they should be spawning. 0 sized bounds is the most precise because it only checks for the exact spawn position's cells and nothing else.

I will test your code again and report back, but the game never waits for the clipmap to be ready to spawn stuff.

Metious: Just tested your code and can confirm that the issue I mentioned exists.

In the void, before my object that is supposed to spawn at `373 -400 1920` was being spawned when I was at `372 -395 1801`, but with your code I had to swim to the exact spawning position.

The same is true for spawns that are in the crater, but you don't have to go inside the spawning position; just a little bit closer to the spawn position but the spawn will still surprise you have to get awfully close.

Metious (approved):

32Kallies (approved):

PR #533, opened 2024-01-17
Add IFModPositionalSound.

MrPurple6411:

  - Add IFModPositionalSound interface for modders to get positional information for where the sounds are supposed to be played.

### DIFF
--- a/Nautilus/FMod/Interfaces/IFModPositionalSound.cs
+++ b/Nautilus/FMod/Interfaces/IFModPositionalSound.cs
@@ -1,0 +1,18 @@
+namespace Nautilus.FMod.Interfaces;
+
+using FMOD;
+using UnityEngine;
+
+/// <summary>
+/// This interface is used to integrate with <see cref="Handlers.CustomSoundHandler"/>.
+/// </summary>
+public interface IFModPositionalSound : IFModSound
+{
+    /// <summary>
+    /// Defines how to play sound in this object.
+    /// </summary>
+    /// <param name="position">The position to play the sound at.</param>
+    /// <param name="channel">The channel on which the sound was created.</param>
+    /// <returns>If the sound was reported as played.</returns>
+    bool TryPlaySound(Vector3 position, out Channel channel);
+}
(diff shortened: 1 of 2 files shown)


Nautilus/FMod/Interfaces/IFModPositionalSound.cs
```diff
@@ -0,0 +1,18 @@
...
+/// </summary>
+public interface IFModPositionalSound : IFModSound
+{
+    /// <summary>
+    /// Defines how to play sound in this object.
+    /// </summary>
+    /// <param name="position">The position to play the sound at.</param>
+    /// <param name="channel">The channel on which the sound was created.</param>
+    /// <returns>If the sound was reported as played.</returns>
+    bool TryPlaySound(Vector3 position, out Channel channel);
```
Metious, line 17: Since this is an interface and derives from the `IFModSound` interface, users will always end up having to implement two `TryPlaySound` overloads, but only one of them is invoked, which is a bit weird.

Making the interface not derive from anything would be better, but I'd like to know the use-case for this because it could be made into a class instead like the [FModMultiSounds](https://github.com/SubnauticaModding/Nautilus/blob/36f928e81f47f2830e3b45074a6353f1008ee357/Nautilus/FMod/FModMultiSounds.cs) class, which is even more ideal.

Metious (changes requested):

Nautilus/FMod/Interfaces/IFModPositionalSound.cs
MrPurple6411: Use case is moot atm as in my testing I am unable to get the stupid 3d attribute setting to properly make a fmod sound only be played when near an object. 


My intent is to be able to have looping audio attached to a buildable but using the fmod sounds system instead of the unity audio systems. 

My issue is though that when I set the custom looping emitter on an object and add the fmodasset and register this it does indeed play the sound BUT if the sound is set to any of the 3d modes it just plays once even through its on a looping emitter and no matter if its set to 3d mode or looping mod it will play  without being directionally positioned....

PR #545, opened 2024-04-25
Now you can stop custom prefabs from getting cached

Metious:

  - Previously we had methods to remove a prefab from the cache, but users didn't have a place to call it, so it was useless. This PR allows users to remove prefabs at real-time (regardless of being in patch time, or in a world).
  - Added a new extension method for `CustomPrefab` to remove a custom prefab from the cache.
  - Fixed a bug where if a custom prefab's game object returns null, the game freezes
  
  ### ~~NOT TESTED~~


Metious:
Tested this PR on stable branch of SN1 (v71288) and works.

32Kallies (approved):
Did not notice any issues while testing

PR #558, opened 2024-09-14
Thunderkit Utilities

### LINKED ISSUE #557: Utilities for Thunderkit Mods
**Utilities for mods made with Thunderkit**
It would be nice if scripts such as these were added to Nautilus for use in mods made in Thunderkit:
https://github.com/Indigocoder1/SubLibrary/blob/main/SubLibrary/Utilities/ApplySNFont.cs
https://github.com/Indigocoder1/SubLibrary/blob/main/SubLibrary/Utilities/ApplySNLayer.cs
https://github.com/Indigocoder1/SubLibrary/blob/main/SubLibrary/Materials/ApplySNShaders.cs
https://github.com/Indigocoder1/SubLibrary/blob/main/SubLibrary/Materials/MaterialSetter.cs

**Purpose**
This would centralize utilities such as these into Nautilus, which is already needed for almost every single mod that uses Thunderkit.

**Usage**
Modders could just apply the given scripts to their needed objects, and upon loading into the scene the changes would be applied.

Indigocoder1:

  - Added various scripts for use with Thunderkit:
  - Added a script for applying the SN font to GameObjects in TK
  - Added a script for applying a layer from SN to a GameObject in TK
  - Added a script for applying the SN shaders to a GameObject in TK
  - Added a script for applying various materials to a GameObject in TK

Note:
Could someone look at these 2 preprocessors? I don't particularly like using them like that and I'm wondering if there's a better way to do that.
[Different Layers in SN & BZ](https://github.com/Indigocoder1/Nautilus/blob/cbd18cbdee8bbee67b29e30040a3ef208a5c2d21/Nautilus/Utility/ThunderkitUtilities/ApplySNLayer.cs#L48l)
[Different materials in SN & BZ](https://github.com/Indigocoder1/Nautilus/blob/ad6de78d5e3afaf968044976351c3ccb362e7cd7/Nautilus/Utility/ThunderkitUtilities/ApplySNMaterial.cs#L184)

My old Asset Bundle tutorial stuff is in here but that was merged a while ago so idk why that shows up still.

### Breaking changes

  - None


Nautilus/Utility/ThunderkitUtilities/ApplySNMaterial.cs
```diff
@@ -113,40 +115,54 @@ private Material GetMaterial(MaterialType type)
...
+
+        string path = "";
+#if BZ_STABLE
+        path = "model/seatruck_anim/Seatruck_cabin_exterior_glass_geo";
+#elif SN_STABLE
+        path = "Model/Submersible_SeaMoth/Submersible_seaMoth_geo/Submersible_SeaMoth_glass_geo";
+#endif
+
+        var glassMaterial = task.GetResult()
             .transform.Find("Model/Submersible_SeaMoth/Submersible_seaMoth_geo/Submersible_SeaMoth_glass_geo")
```
EldritchCarMaker: Did you forget to use the path variable here? This is only going to look for the seamoth in both games.

EldritchCarMaker:
I also question the necessity of the layer assigning at all, to be honest. Do layers assigned to prefabs in unity not work when imported into the game? I feel like that would be the better option to use, rather than relying on a component to set the layer at runtime

EldritchCarMaker:
We've been wanting to set up a "clone" of the subnautica marmosetuber shader for use in unity to allow shader property customization within the editor for a while now. If we were to get that set up and potentially commonplace, wouldn't the shader setting component overwrite any values already on the shader? Is that something worth worrying about? If so, would that be better fixed in the AddSNShaders utility method itself, rather than this wrapper component for it?

Indigocoder1:
> I also question the necessity of the layer assigning at all, to be honest. Do layers assigned to prefabs in unity not work when imported into the game? I feel like that would be the better option to use, rather than relying on a component to set the layer at runtime

They would have to have the same index in Unity as in the game. Technically you could set that up but it seems more convenient to just have them all in a list/enum to apply. Tbh the only use I've found for it so far is for custom damage decals (which need the Usable layer), but I just added them all anyway

Indigocoder1:
> We've been wanting to set up a "clone" of the subnautica marmosetuber shader for use in unity to allow shader property customization within the editor for a while now. If we were to get that set up and potentially commonplace, wouldn't the shader setting component overwrite any values already on the shader? Is that something worth worrying about? If so, would that be better fixed in the AddSNShaders utility method itself, rather than this wrapper component for it?

Probably? Tbh I'm not too familar with the ApplySNShaders method itself but I would say it should be fixed in there, not the wrapper (Assuming this is a problem)

Nautilus/Utility/ThunderkitUtilities/ApplicationModes.cs
```diff
@@ -0,0 +1,16 @@
+namespace Nautilus.Utility.ThunderkitUtilities;
+
+internal enum GeneralApplicationMode
```
Metious: I think this enum name may come across as confusing given "Application" is already a term within Unity, especially in the editor. I think `SetMode` would work better

Nautilus/Utility/ThunderkitUtilities/ApplySNLayer.cs
```diff
@@ -0,0 +1,56 @@
...
+{
+    [Tooltip("The name of the layer you want to apply")]
+    public LayerName layerName;
+
+    [Tooltip("How to apply the layer")]
+    public GeneralApplicationMode applicationMode;
+
+    private void Start()
+    {
+        FieldInfo field = typeof(LayerID).GetField(layerName.ToString(), BindingFlags.Static | BindingFlags.Public);
```
Metious: This is not only using reflection, but also relies on `enum.ToString()`. Since a prefab can have multiple active components at the same time, I don't think the code lines the reflection saves are worth the consequences. A switch statement would be better

Nautilus/Utility/ThunderkitUtilities/ApplySNMaterial.cs
```diff
@@ -0,0 +1,192 @@
...
+            MaterialType.ExteriorGlass => _exteriorGlassMaterial,
+            MaterialType.ShinyGlass => _shinyGlassMaterial,
+            MaterialType.InteriorWindowGlass => _interiorWindowGlassMaterial,
+            _ => null
+        };
+
+        return mat;
+    }
+
+    private IEnumerator LoadMaterialsAsync()
```
Metious: This method should be moved to `MaterialUtils` and called in `MaterialUtils.PatchInternal`

Nautilus/Utility/ThunderkitUtilities/ApplicationModes.cs
```diff
@@ -0,0 +1,16 @@
...
+{
+    SingleObject,
+    AllChildObjects,
+    AllChildObjectsIncludeInactive,
+}
+
+internal enum MaterialApplicationMode
```
Metious: Same here, `MaterialSetMode`

Metious (changes requested):
Left a few comments. The code as a whole LGTM and useful. Nicely done

Metious:
> We've been wanting to set up a "clone" of the subnautica marmosetuber shader for use in unity to allow shader property customization within the editor for a while now. If we were to get that set up and potentially commonplace, wouldn't the shader setting component overwrite any values already on the shader? Is that something worth worrying about? If so, would that be better fixed in the AddSNShaders utility method itself, rather than this wrapper component for it?

The wrapper doesn't need to know how to handle them, it would just "redirect" a game object to what knows how to handle them.

EldritchCarMaker (approved):
I forgot to add my review earlier :(

Metious has already pointed out my one comment by now, but otherwise looks good to me

Nautilus/Utility/ThunderkitUtilities/ApplySNLayer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+namespace Nautilus.Utility.ThunderkitUtilities;
+
+internal class ApplySNLayer : MonoBehaviour
+{
+    [Tooltip("The name of the layer you want to apply")]
+    public LayerName layerName;
+
+    [Tooltip("How to apply the layer")]
+    public GeneralSetMode layerSetMode;
```
EldritchCarMaker: Should this be GeneralSetMode or MaterialSetMode? Feels odd to not allow the ui only option, when the materials have it

Nautilus/Utility/ThunderkitUtilities/ApplySNLayer.cs
```diff
@@ -0,0 +1,75 @@
...
+                GetComponentsInChildren<GameObject>().ForEach(g => g.layer = layer);
+                break;
+            case GeneralSetMode.AllChildObjectsIncludeInactive:
+                GetComponentsInChildren<GameObject>(true).ForEach(g => g.layer = layer);
+                break;
+        }
+    }
+
+    /// <summary>
+    /// These are taken from <see cref="LayerID"/>, and are retrieved using reflection
```
EldritchCarMaker: This comment is outdated now

Nautilus/Utility/ThunderkitUtilities/ApplySNLayer.cs
Indigocoder1: I can't really see a use case where the UI only would be needed but I suppose there's no harm in adding it

Metious (approved):

PR #564, opened 2024-11-10
FMOD fade-outs

Metious:

  - The game now respects fade-out points for custom sounds.

You can add a fade-out point using one of the two methods:
1. Directly on the `Sound` object when you're creating one. This makes the Sound always fade-out when fading is allowed
2. On the Channel object after the sound is played. This approach is manual and only applies the fade-out once.

### **REQUIRES THOROUGH TESTING**

### DIFF
--- a/Nautilus/Patchers/CustomSoundPatcher.cs
+++ b/Nautilus/Patchers/CustomSoundPatcher.cs
@@ -958,6 +958,11 @@ internal class CustomSoundPatcher
             channel.stop();
             return false;
         }
+
+        channel.getDelay(out ulong _, out ulong _, out bool stopChannels);
+        
+        if (stopChannels)
+            return false;
             
         RuntimeManager.CoreSystem.getSoftwareFormat(out var samplesRate, out _, out _);
         channel.AddFadeOut(fadeOut.Seconds, out var dspClock);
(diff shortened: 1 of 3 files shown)


Nautilus/Patchers/CustomSoundPatcher.cs
```diff
@@ -923,4 +950,18 @@ internal static void SetChannel3DAttributes(Channel channel, Vector3 position)
...
     }
+    
+    private static bool TryFadeOutBeforeStop(Channel channel)
+    {
+        if (channel.getCurrentSound(out var sound) != RESULT.OK || !FadeOuts.TryGetValue(sound.handle, out var fadeOut))
+        {
+            channel.stop();
+            return false;
+        }
+
```
32Kallies, line 966:
```suggestion

channel.getDelay(out ulong _, out ulong _, out bool stopChannels);
        
if (stopChannels)
    return false;
    
```

This prevents unnecessary and conflicting fade points from cluttering the sound channel.

32Kallies (changes requested):
Needs a fix.

32Kallies (approved):

Nautilus/Patchers/CustomSoundPatcher.cs
```diff
@@ -761,14 +778,22 @@ public static bool FMOD_CustomEmitter_Play_Prefix(FMOD_CustomEmitter __instance)
         
         [HarmonyPatch(typeof(FMOD_CustomEmitter), nameof(FMOD_CustomEmitter.Stop))]
         [HarmonyPrefix]
-        public static bool FMOD_CustomEmitter_Stop_Prefix(FMOD_CustomEmitter __instance)
+        public static bool FMOD_CustomEmitter_Stop_Prefix(FMOD_CustomEmitter __instance, STOP_MODE stopMode)
```
JKohlman, line 781: Causes #573. This method in BZ doesn't take a STOP_MODE parameter.

PR #588, opened 2025-02-17
Fixed buttons appearing at bottom of mod menu.

### LINKED ISSUE #539: Config buttons always go to the bottom of the config
**Describe the bug**
In code the buttons are each below the "MK1/2/3 Blue (B)" sliders, but they all end up at the bottom. 

You have to manually order *every* attribute to fix this which has been phenomenally annoying as I've been making mods with a fuck ton of options and need to re-order them every time I add a new option

**Expected behavior**
The buttons should be positioned under the "MK1/2/3 Blue (B)" sliders, not bunched at the bottom of the config

**Screenshots**
Note: old screenshot from 15 months ago but the issue still exists, my config is much prettier than this ugly thing now
image

FearDragon:

  - Process field, property, and method info in one foreach statement when processing metadata.
  - Resolves #539 
  -

### Breaking changes

  - N/A
  -
  -

### DIFF
--- a/Nautilus/Options/Attributes/ConfigFileMetadata.cs
+++ b/Nautilus/Options/Attributes/ConfigFileMetadata.cs
@@ -65,10 +65,12 @@ internal class ConfigFileMetadata<T> where T : ConfigFile, new()
             switch (member)
             {
                 case FieldInfo field:
-                    processFieldOrProperty(field, MemberType.Field, field.FieldType);
+                    if (!field.IsStatic)
+                        processFieldOrProperty(field, MemberType.Field, field.FieldType);
                     break;
                 case PropertyInfo property:
-                    processFieldOrProperty(property, MemberType.Property, property.PropertyType);
+                    if (!property.GetGetMethod(true).IsStatic)
+                        processFieldOrProperty(property, MemberType.Property, property.PropertyType);
                     break;
                 case MethodInfo method:
                     processMethod(method);
--- a/Nautilus/Options/Attributes/ConfigFileMetadata.cs
+++ b/Nautilus/Options/Attributes/ConfigFileMetadata.cs
@@ -56,26 +56,24 @@ internal class ConfigFileMetadata<T> where T : ConfigFile, new()
 
     private void processMetadata()
     {
-        BindingFlags bindingFlags = BindingFlags.Instance | BindingFlags.Public;
-        foreach (PropertyInfo property in typeof(T).GetProperties(bindingFlags)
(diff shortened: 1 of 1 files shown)


FearDragon:

32Kallies:
I wish this worked, because it looks like it would. Unfortunately it seems to place buttons at the top of the menu.

According to [Microsoft documentation](https://learn.microsoft.com/en-us/dotnet/api/system.type.getmembers):
> In .NET 6 and earlier versions, the [GetMembers](https://learn.microsoft.com/en-us/dotnet/api/system.type.getmembers?view=net-9.0) method does not return members in a particular order, such as alphabetical or declaration order. Your code must not depend on the order in which members are returned, because that order varies. However, starting with .NET 7, the ordering is deterministic based upon the metadata ordering in the assembly.

Unfortunately, Unity does not use .NET 6 or later. We might have to mark #539 as "won't fix," since it appears to be impossible.

Nautilus/Options/Attributes/ConfigFileMetadata.cs
```diff
@@ -56,26 +56,24 @@ public void ProcessMetadata()
...
-
-        foreach (MethodInfo method in typeof(T).GetMethods(bindingFlags | BindingFlags.Static)
-                     .Where(memberIsDeclaredInConfigFileSubclass) // Only care about members declared in a subclass of ConfigFile
-                     .Where(memberIsNotIgnored)) // Filter out explicitly ignored members
-        {
-            processMethod(method);
+            switch (member)
+            {
+                case FieldInfo field:
+                    processFieldOrProperty(field, MemberType.Field, field.FieldType);
```
32Kallies: The previous code checked that fields were not static.

```suggestion
                    if (!field.IsStatic)
                        processFieldOrProperty(field, MemberType.Field, field.FieldType);
```

Nautilus/Options/Attributes/ConfigFileMetadata.cs
```diff
@@ -56,26 +56,24 @@ public void ProcessMetadata()
...
+                    break;
+                case PropertyInfo property:
+                    processFieldOrProperty(property, MemberType.Property, property.PropertyType);
```
32Kallies: The previous code also checked that properties were not static.
```suggestion
                    if (!property.IsStatic)
                        processFieldOrProperty(property, MemberType.Property, property.PropertyType);
```

32Kallies (changes requested):
Unfortunately, order still doesn't seem to work properly. Might be impossible.

32Kallies:
Seems like we can't fix this because the order of class members isn't deterministic. I'll close the issue as well.

PR #601, opened 2025-04-07
fix: Prevent ArgumentException on KnownTech reinit in running game.

tinyhoot:
`KnownTechPatcher` uses a naked dictionary Add() for known compound tech based on data that is already in that same dictionary. This leads to an ArgumentException if a mod attempts to change KnownTech in any way late into a running game, particularly if the cyclops has already been unlocked.

- Check whether known compound tech is already in the dictionary before adding to it.

### Breaking changes

- None, this is a small fix.

### DIFF
--- a/Nautilus/Patchers/KnownTechPatcher.cs
+++ b/Nautilus/Patchers/KnownTechPatcher.cs
@@ -2,6 +2,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 using HarmonyLib;
 using Handlers;
 using Utility;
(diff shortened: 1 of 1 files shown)


Metious (approved):

JKohlman:
I don't have access to the game code atm but couldn't you just use `Dictionary.TryAdd` and disregard the result without needing a whole where?

tinyhoot:
`Dictionary.TryAdd()` is not part of .NET Framework [(see docs)](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.dictionary-2.tryadd). For modern C# you would be correct.

JKohlman:
Absolutely fair, I originally went to the docs, switched to net framework 472 and Microsoft in all their wisdom doesn't say "this doesn't exist" but rather just gives a purple bar (not even red) "You have been redirected to the newest product version available for this page" and swaps back to .NET 9

PR #606, opened 2025-06-01
Add post-build deploy to game directory.

tinyhoot:

  - Add a simple post-build event to automatically deploy the assembly to the game directory for testing purposes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -260,3 +260,6 @@ paket-files/
 __pycache__/
 *.pyc
 Assembly-CSharp*
+
+# Personalised information about game installation directory for post-build deploy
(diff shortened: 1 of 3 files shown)


32Kallies (approved):
Great change, and long overdue.

(Is it coincidence that we implement this in the exact same way? Or is this just the best way to approach it?)

tinyhoot:
The example file is taken directly from the Seal sub after @EldritchCarMaker threw it at me, but the build targets are the ones I've refined for my own mods over the years. I don't remember where I got the idea from but it's probably the most straightforward solution, I've definitely seen it floating around in various unity games' modding scenes.

PR #623, opened 2025-08-04
feat: Support version checking via FindMyUpdates

### LINKED ISSUE #622: Version checking support via FindMyUpdates
**Describe the feature**
Allowing users to check for Nautilus updates via [Find My Updates - FMU](https://www.nexusmods.com/subnautica/mods/2431)

**Purpose**
Any additional mod that supports version checking is helpful to users

**Implementation**
1. Use Nautilus' own system to support version checking, finding the most suitable place to call:
```cs
ModMessageSystem.SendGlobal("FindMyUpdates", "URL for raw .json file");.
```
2. Add a `Version.json` file to the solution folder
```json
{
  "ModName": "Nautilus",
  "LatestVersion": "1.0.0.41",
  "LatestURL": "https://www.nexusmods.com/subnautica/mods/1262"
}
```
3. (Optional) Write a powershell script to run in Post Build Events that will automatically set the "LatestVersion" field of "Version.json" to the version found here: https://github.com/SubnauticaModding/Nautilus/blob/master/Version.targets#L5-L6

RamuneNeptune:
  - Resolves #622

### DIFF
--- a/Version.json
+++ b/Version.json
@@ -1,0 +1,5 @@
+{
+  "ModName": "Nautilus",
+  "LatestVersion": "1.0.0.42",
+  "LatestURL": "https://www.nexusmods.com/subnautica/mods/1262"
+}
--- a/Nautilus.sln
+++ b/Nautilus.sln
@@ -15,6 +15,7 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution 
 		common.props = common.props
 		Directory.Build.props = Directory.Build.props
 		PostBuild.targets = PostBuild.targets
+		Version.json = Version.json
 		Version.targets = Version.targets
 	EndProjectSection
 EndProject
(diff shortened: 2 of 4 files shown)


Version.json
```diff
@@ -0,0 +1,5 @@
+{
```
32Kallies, line 1: I do want to add support for this at some point, but we already have so much to keep up with when updating, we don't want to have another file that needs to be edited with each version increase.

32Kallies (changes requested):

Version.json
Law-Abiding-Developer: No bot or command can fix that? Interesting

RamuneNeptune: File can be updated automatically for releases through various methods, Ramune cbf to implement any of them. We will wait for a Hero.

RamuneNeptune:
[Find My Updates](https://www.nexusmods.com/subnautica/mods/2431) 1.1.0 or later now supports version checking for Nautilus, without any changes necessary within Nautilus itself

PR #630, opened 2025-08-18
Unity input system support

Metious:

  - Upgraded to C# 12 (REQUIRES .NET 8!)
  - Added support for Unity.InputSystem
  - Made the KeybindOption Below zero only.
  - Added a new options tab called "Mod Input" for custom buttons that are bindable.
  - Added a new enum processing event that's called whenever a custom enum is created.
  
### Changes open for discussion
  - [ ] Provide a way to add inputs created in a Unity project
    - While it's extremely easy to get their bindings and whatnot, the requirement of enum objects for each input action is what complicates it. (source generator?)
  - [ ] The possibility of keeping the `KeybindAttribute` for Subnautica
  - [x] Support raw display name addition

### Breaking changes

  - KeybindOption and `Keybind` attribute no longer exist in SN1.
  
### Build (SN1) Up-to-date with 5d95aac
[Nautilus_SN.STABLE_1.0.0.45.zip](https://github.com/user-attachments/files/22911251/Nautilus_SN.STABLE_1.0.0.45.zip)

## New Input System (2025)
The new input system requires two things:
1. GameInput.Button
2. InputAction corresponding the button enum.

While Unity recommends relying on input action objects, Subnautica took a different approach and instead made wrappers that render the input actions internal-only by making every action type retrievable via `GameInput.GetButtonDown`, `GameInput.GetButtonHeld`, `GameInput.GetButtonUp`, `GameInput.GetFloat`, and `GameInput.GetVector2` methods.

This means that we'll have to add support for creating a new button enum object, but also translate settings to InputAction so custom buttons can be internally sound.

## How to use
To use this system, create a new `GameInput.Button` object using our enum handler, then specify the settings as desired.

Below is an example that demonstrates the usage of the new system, and how to listen to a button press.
```csharp
public GameInput.Button PrintButton = EnumHandler.AddEntry<GameInput.Button>("MyPrint")
    .CreateInput(InputActionType.Button)
    .WithKeyboardBinding("<Keyboard>/l") // Set the default keyboard binding to the button "L"
    
private void Update()
{
   if (GameInput.GetButtonDown(PrintButton))
   {
      ErrorMessage.AddDebug("Button was pressed!");
   }
}
```

~~Since this system if primarily intended to be used inside Unity projects, unfortunately there's no public source that lists every binding path (technically they're a form of regular expression, so they can be many things). However, we can generate a list of constant bindings, similar to the [prefab paths](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Documentation/resources/SN1-PrefabPaths.json).~~

There's now a class that lists every possible button in every known device in [GameInputHandler.Paths](https://github.com/SubnauticaModding/Nautilus/blob/master/Nautilus/Handlers/GameInputHandler.cs#L11).

## How to migrate from keybind options to the new input system
Because `GameInput.Button` is simply a representation for a set of multiple keybinds and can be many things, we've decided to move them out of the `ModOptions` class and make them their own thing.

Below is a demonstration of how to convert a `KeyCode` keybind option to `GameInput.Button`
`KeyCode` keybind option:
```csharp
[Menu("My awesome configs")]
public class Config : ConfigFile
{
   [Keybind("Print Button")]
   public KeyCode PrintButton = KeyCode.L;
}

public Config MyConfig = OptionsPanelHandler.RegisterModOptions<Config>();

public void Update()
{
   if (Input.GetKeyDown(MyConfig.PrintButton))
   {
     ErrorMessage.AddMessage("Print button was pressed!");
   }
}
```
`GameInput.Button` equivalent:
```csharp
public GameInput.Button PrintButton = EnumHandler.AddEntry<GameInput.Button>("PrintButton")
   .CreateInput("Print Button")
   .WithKeyboardBinding("<Keyboard>/l");
   
public void Update()
{
   if (GameInput.GetButtonDown(PrintButton))
   {
     ErrorMessage.AddMessage("Print button was pressed!");
   }
}
```

Note that the input display names can be localized by the `Option{ENUM_NAME}` language string. For more information, check out [the example mod's localization file](https://github.com/SubnauticaModding/Nautilus/blob/master/Example%20mod/Localization/English.json).

## Categories
With the addition of the new "Mod Input" tab, we've also facilitated a way for modders to create new and name their categories for each button they make. Below demonstrates the usage of this system.

```csharp
public GameInput.Button PrintButton = EnumHandler.AddEntry<GameInput.Button>("PrintButton")
   .CreateInput("My Print")
   .WithKeyboardBinding("<Keyboard>/l")
   .WithCategory("NautilusExamplePrintCategory");
```
The above code produces the following category in the Mod Input tab:
image

Important to note that categories are unique, meaning if another mod already added a category with the same name, your buttons and the other mod's buttons will be merged into one category.

If a category was not set for a button, the mod name defined in your `BepinPlugin` will be used instead.

Category display names can also be localized by the same string passed to `WithCategory` method.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,7 +15,7 @@ jobs:
             - name: .NET Setup
               uses: actions/setup-dotnet@v3
               with:
-                  dotnet-version: 7.x
+                  dotnet-version: 8.x
               
             - name: Build Project
               run: dotnet build Nautilus.sln -c SN.STABLE         
@@ -28,7 +28,7 @@ jobs:
             -   name: .NET Setup
                 uses: actions/setup-dotnet@v3
                 with:
-                    dotnet-version: 7.x
+                    dotnet-version: 8.x
 
             -   name: Build Project
                 run: dotnet build Nautilus.sln -c BZ.STABLE
--- a/Example mod/ConfigExamples.cs
+++ b/Example mod/ConfigExamples.cs
@@ -185,6 +185,7 @@ public class Config : ConfigFile
     [Choice("My customised enum-based choice", "1", "2", "3"), OnChange(nameof(MyChoiceValueChangedEvent))]
     public CustomChoice ChoiceCustomEnum;
 
+#if BELOWZERO
     /// <summary>
     /// <para>A <see cref="KeybindAttribute"/> is represented in the mod options menu as a customistable keybind, where the
     /// user clicks the box to set the binding, stored as a <see cref="KeyCode"/>.</para>
@@ -193,6 +194,7 @@ public class Config : ConfigFile
     /// </summary>
     [Keybind("My keybind"), OnChange(nameof(MyKeybindValueChangedEvent))]
     public KeyCode KeybindKey;
+#endif
 
     /// <summary>
     /// <para>A <see cref="SliderAttribute"/> is used to represent a numeric value as a slider in the options menu, with a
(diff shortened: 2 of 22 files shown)


32Kallies (approved):
Looks good to me, works after extensive testing

PR #631, opened 2025-08-20
Implement MaterialLibrary

TheRealLyonic:

  - Added a new utility class; MaterialLibrary, for simple fetching and application of base-game materials.
  - Added two embedded resource files, which are required for MaterialLibrary's functionality.
  - Added a new method to MaterialUtils, called `RemoveInstanceFromMatName()`, making it easier to compare material names programmatically.

### Breaking changes

  - None


Nautilus/Nautilus.csproj
```diff
@@ -52,5 +52,12 @@
   <ItemGroup>
     <None Include="DocFX\logo.png" Pack="true" PackagePath="" />
   </ItemGroup>
+
+  <ItemGroup>
+    <None Remove="Resources\MatFilePathMapBZ.resources" />
+    <EmbeddedResource Include="Resources\MatFilePathMapBZ.resources" />
+    <None Remove="Resources\MatFilePathMapSN.resources" />
+    <EmbeddedResource Include="Resources\MatFilePathMapSN.resources" />
+  </ItemGroup>
```
32Kallies, line 61: This includes both embedded resources in both DLLs, right? The size may be insignificant, but since this is loaded into memory, and because the other game's materials would be completely irrelevant, they should be included/excluded based on the build target.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,352 @@
...
+            }
+
+            //Sadly, this is actually the simplest way to get the total number of entries in a .resources file.
+            int materialEntries = 0;
+            foreach (var _ in resourceSet)
+                materialEntries++;
+            
+            return materialEntries;
+        }
+    }
```
32Kallies: I would suggest converting `Size` from a property to a method, e.g. `GetMaterialEntriesCount` or `GetEntryCount`.

I know those names are a bit long, but it's often best (and even a C# convention) to be more clear. `Size` is a bit unclear to me. But also, I'm not certain if this would be necessary to expose in the first place.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,352 @@
...
+    
+    internal static void Patch()
+    {
+        #if SUBNAUTICA
+                string resourcePath = "Nautilus.Resources.MatFilePathMapSN";
+        #elif BELOWZERO
+                string resourcePath = "Nautilus.Resources.MatFilePathMapBZ";
+        #endif
+        
+        _resourceManager = new ResourceManager(resourcePath, Assembly.GetExecutingAssembly());
```
32Kallies: Couldn't Patch be removed entirely with this being done with the field's definition?

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,352 @@
...
+                    continue;
+                }
+
+                newMatList[i] = foundMaterial;
+                loadedVanillaMats.Add(foundMaterial);
+            }
+            
+            renderer.materials = newMatList;
+        }
+    }
```
32Kallies: I know I'm nitpicking here, I hope it's not too annoying. That being said, I *very much prefer* the name `ReplaceVanillaMaterials`, it's much clearer and C# naming conventions almost always include avoiding abbreviations. Also, I feel that `prefab` makes more sense than `customPrefab`, but that point hardly matters at all.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,352 @@
...
+                        return;
+                    }
+                }
+            }
+            
+            matCheckFailed = true;
+        });
+        
+        yield return new WaitUntil(() => materialSet || matCheckFailed);
+    }
```
32Kallies: This is the only part that I would request an entire revision of. Unless I am entirely misunderstanding or missing something, the current code means that *for every individual material that needs it*, an entire Unity scene is loaded and iterated across in its entirety, with all of its renderers and materials loaded into memory and compared individually. The implications of that for even just one prefab with 10 materials from a scene is at least a bit detrimental, in my opinion. Also, I have a feeling that the WaitUntil check could run indefinitely if initialized at the wrong point in time.

Nautilus/Utility/MaterialUtils.cs
```diff
@@ -333,6 +333,29 @@ public static void SetMaterialCutout(Material material, bool cutout)
...
+        // Or in the middle of their material names. Not likely to happen, but better safe than sorry.
+        if (matName.EndsWith(instanceString))
+        {
+            returnValue = matName.Substring(0, matName.Length - instanceString.Length);
+            return RemoveInstanceFromMatName(returnValue);
+        }
+        
+        return returnValue;
+    }
+
```
32Kallies: This could probably be moved to `GeneralExtensions.cs`, just like `GeneralExtensions.TrimClone`.

32Kallies (changes requested):
Very nice system, but with some concerning parts.

Nautilus/Nautilus.csproj
TheRealLyonic: Good point! Somehow missed that I could throw in conditions to the ItemGroups when I first made this commit. Fixed in the latest update to this PR.

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: I initially wanted to expose the library's entry-count to facilitate implementing index-based material retrieval later down the line, because I wanted to make it possible to easily iterate through every material in the library, but I've shelved that feature for the time being, so I've just removed this altogether for now. Will probably revisit this in a future PR.

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: Sensible change, added in the new commit! To be entirely honest, I had just seen that other utility classes were implementing Patch methods, and figured it'd be best practice for me to do the same. But yeah, no real reason for this class to have one, in hindsight.

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: Proper convention is always important, regardless of nitpick level. Changes made! (ReplaceVanillaMats does sound kinda weird, now that I say it outloud, compared to the new version).

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: So this is where things get a bit tricky. The performance oversight you pointed out with one prefab potentially having multiple base-game materials from the same large scene, leading to that scene being loaded and iterated over _multiple times for one prefab_ was very helpful, and I've since refactored the material-fetching system to support taking in multiple material names as parameters for one path, both in regards to scenes, and just regular old prefabs.

So now, any time a prefab a user passes in has multiple materials with the same path (I.e. a scene prefab path), those material names all get passed in to the fetch method at the same time, so they can be looked for synchronously, without any need for re-loading prefabs, or doubling back to re-check for the other materials later.

Having said that, the process for finding materials with paths that point towards scene-prefabs remains largely unchanged, and we still load entire scenes just to find one or two materials--Though we do break from the renderers loop immediately after finding the material(s) we need.

This is unfortunately the only real way to get these materials, though, and the good news is that there are really only a couple of materials in the entire game that have scene-prefabs as paths, because all entries in the prefab database and the Addressables-system were checked prior to the scenes. Meaning the only materials we look for on scene-prefabs are those that do not exist on any other prefab in the entire game, and that don't have a .mat file path to be loaded with the AddressablesUtility. Hopefully the new optimizations are enough to counteract any effects from needing to retrieve some of these more obscure materials in this way.

I did also add a timeout condition to the WaitUntil check you mentioned having some concern over, just in-case some odd-timing results in a broken system. I've set the timeout to occur after 5-minutes for the time being--Which is probably overkill, but I wanted to give ample time for the ordinary wait-condition to be met before we assume the process failed and manually cancel it, just to be on the safe side.

Nautilus/Utility/MaterialUtils.cs
TheRealLyonic: I somehow had no idea this class existed--Change made! Also renamed the method to `TrimInstance` so it fits better next to `TrimClone`.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,361 @@
...
+                    InternalLogger.Error($"Max retries limit reached when trying to load material at path: {matPath}.");
+                    InternalLogger.Error("Please ensure the material's path is valid, or up the maximum # of retries.");
+                    break;
+                }
+
+                fetchAttempts++;
+                InternalLogger.Debug($"Attempting to load material at path: {matPath}...");
+                
+                var handle = AddressablesUtility.LoadAsync<Material>(matPath);
+                yield return handle.Task;
```
Metious: This is an invalid `yield` instruction; Unity doesn't support yielding `Task`s, those are purely used in proper asynchronous methods with `async`/`await` keywords. This is also why the handle is returning back before it's done for you.

For coroutines, simply yielding the handle is enough to pause the method execution until it's done.
This also means you no longer need to have more than one fetch attempt and I recommend removing the fetching attempts logic.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,361 @@
...
+        var loadStartTime = Time.time;
+
+        bool materialsFound = false;
+        bool matCheckFailed = false;
+        LightmappedPrefabs.main.RequestScenePrefab(sceneName, scenePrefab =>
+        {
+            var foundMats = new List<string>();
+            foreach (var renderer in scenePrefab.GetAllComponentsInChildren<Renderer>())
+            {
+                foreach (var material in renderer.materials)
```
Metious: shouldn't this be `sharedMaterials`? If not, `FetchMaterials` needs to have a note suggesting that the returned materials are unmanaged and must be destroyed once the object using them is no longer in use.

Nautilus/Utility/MaterialLibrary.cs
```diff
@@ -0,0 +1,361 @@
...
+        }
+
+        var foundMats = new List<string>();
+        bool allMatsFound = false;
+        foreach (var renderer in prefab.GetAllComponentsInChildren<Renderer>())
+        {
+            if (allMatsFound)
+                break;
+            
+            foreach (var material in renderer.materials)
```
Metious: shouldn't this be `sharedMaterials`? If not, `FetchMaterials` needs to have a note suggesting that the returned materials are unmanaged and must be destroyed once the object using them is no longer in use.

Metious (changes requested):
Added some concerns I have. Once these are resolved, the PR is ready for merge.

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: Good catch! This issue has been bothering me massively, happy it was such an easy fix. Corrected in the latest commit.

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: Initially, I was avoiding accessing .sharedMaterials here because I had noticed that I got a larger mat-count when just searching in .materials, but upon further inspection, the only mat we miss when using .sharedMaterials is labeled 'Standard', and isn't actually used in a significant capacity anywhere throughout either game. For this reason, the change can safely be made, and this has been updated in the latest commit!

Nautilus/Utility/MaterialLibrary.cs
TheRealLyonic: Same initial cause as detailed above, resolved in latest commit.

Metious (approved):
LGTM

32Kallies (approved):
Tested and works as intended

PR #632, opened 2025-08-27
FEAT: Adds optional AdvancedExposeAlphaChannel to ModColorOption

Knight-Sirius:
Tiny extension of ModColorOption to include "AdvancedExposeAlphaChannel" an optional parameter in ModColorOption.Create() that allows you to hide the alpha slider controls for use in instances where it isn't necessary to show it.

Backwards compatible to the best of my knowledge, and defaults to being visible unless specified.

This is my first PR <3 Apologies in advance!

  - Adds "AdvancedExposeAlphaChannel" as an optional parameter to ModColorOption.Create

### Breaking changes

  - None. Should be fully backwards compatible

### DIFF
--- a/Nautilus/Options/ModColorOption.cs
+++ b/Nautilus/Options/ModColorOption.cs
@@ -33,6 +33,11 @@ public class ModColorOption : ModOption<Color, ColorChangedEventArgs>
     public string Tooltip { get; }
 
     /// <summary>
+    /// When using Advanced UI, whether to show the Alpha controls
+    /// </summary>
+    public bool AdvancedExposeAlphaChannel { get; set; }
+
+    /// <summary>
     /// The base method for adding an object to the options panel
     /// </summary>
     /// <param name="panel">The panel to add the option to.</param>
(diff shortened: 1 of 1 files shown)


Knight-Sirius:

Nautilus/Options/ModColorOption.cs
```diff
@@ -81,24 +86,28 @@ public override void AddToPanel(uGUI_TabbedControlsPanel panel, int tabIndex)
...
                 SliderLabelMode.Percent, "{0:F0}", "The opaqueness of the color. The lower the value the more transparent.");
+            }
         }
 
         OptionGameObject = colorPicker.transform.parent.gameObject;
         base.AddToPanel(panel, tabIndex);
     }
 
-    private ModColorOption(string id, string label, Color value, bool advanced = false, string tooltip = null) : base(label, id, value)
+    private ModColorOption(string id, string label, Color value, bool advanced = false, string tooltip = null, bool advancedExposeAlphaChannel = true) : base(label, id, value)
```
32Kallies, line 106: I like the idea overall, but please keep in mind that changing existing constructors is a breaking change for all mods that used that constructor, even if you are only adding optional parameters.

32Kallies (changes requested):
Good concept, but has an unnecessary breaking change

Nautilus/Options/ModColorOption.cs
Knight-Sirius: Ah, apologies I'm a noob. 

I did test this change with my own mods that make extensive use of ModColorOption and it didn't break specifically and only because I added it as the final parameter in the constructor so prior implementations would just go with the default value automatically.

ModColorOption.Create() specifically is what I'm referring to in my case, but I suppose there are other usecases you're referring to I didn't take into consideration...?

Thanks for reviewing!

32Kallies: I'M BLIND! This is a private constructor. It's fine, sorry!!

32Kallies (approved):
That was a mistake on my part, sorry

Knight-Sirius:
> That was a mistake on my part, sorry

Ayyy self-worth restored! Hahaha
Thanks for reviewing it, appreciate you!

32Kallies:
In the future it would be nice to have this reflected in the attributes too, but I will merge this for now.

PR #639, opened 2025-10-19
Mouse paths hotfix

Metious:

  - Fixed mouse paths not being recognizable by the game's input formatter.
  - Added Left and Right variants for: `shift`, `alt`, and `ctrl`
  - Turned all path fields to `static readonly` so the paths can be updated by Nautilus with it also affecting all mods that use them.


32Kallies (approved):
All the mouse paths are fixed, though Control still needs to be Ctrl

PR #649, opened 2025-11-09
Docs: Documented support for language IDs in `ChoiceAttribute`'s `Options` list

xale:

  - Modified doc comments for `ChoiceAttribute` constructors to mention localization-lookup support in the `options` list parameters.

### Breaking changes

  - n/a

### DIFF
--- a/Nautilus/Options/Attributes/ChoiceAttribute.cs
+++ b/Nautilus/Options/Attributes/ChoiceAttribute.cs
@@ -50,7 +50,7 @@ public sealed class ChoiceAttribute : ModOptionAttribute
     /// </remarks>
     /// <param name="label">The label for the choice. If none is set, the name of the member will be used.</param>
     /// <param name="options">The list of options for the user to choose from. Values may be either ordinary strings for
-    /// direct display, or IDs for lookup with <see cref="Language.Get(string)"/> for translation. Use the
+    /// direct display, or IDs for lookup with <c>Language.Get(string)</c> for translation. Use the
     /// <see cref="Handlers.LanguageHandler"/> API to register translation strings.</param>
     public ChoiceAttribute(string label = null, params string[] options) : base(label)
     {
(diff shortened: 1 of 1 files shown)


xale:

32Kallies (approved):
Maybe consider editing it to refer to the Subnautica `Language` system in general, since this applies to more strings than just those handled by Nautilus, but looks good to me.

Nautilus/Options/Attributes/ChoiceAttribute.cs
```diff
@@ -49,7 +49,9 @@ public sealed class ChoiceAttribute : ModOptionAttribute
     /// <see cref="Enum"/> choices can also be parsed from their values by merely omitting the <paramref name="options"/>.
     /// </remarks>
     /// <param name="label">The label for the choice. If none is set, the name of the member will be used.</param>
-    /// <param name="options">The list of options for the user to choose from.</param>
+    /// <param name="options">The list of options for the user to choose from. Values may be either ordinary strings for
+    /// direct display, or IDs for lookup with <see cref="Language.Get(string)"/> for translation. Use the
```
32Kallies: I have noticed that we should limit or end our use of `cref` for game-specific classes, because for instance `<see cref="Language.Get(string)"/>` creates a link on the documentation website that takes users to [this URL](https://learn.microsoft.com/en-us/dotnet/api/system.string?view=net-10.0). We probably need to retroactively fix other instances of this same issue. Otherwise, looks good to me, thanks for the quick update!

image

32Kallies (changes requested):
Minor issue with page rendering

Nautilus/Options/Attributes/ChoiceAttribute.cs
xale: Ah, I see what you mean. Replaced it with basic fixed-width formatting.

32Kallies (approved):

PR #652, opened 2025-11-17
Remove extra plugins folder from zip folder

Law-Abiding-Developer:
I disliked the fact that there was an extra plugins folder in the zip file created upon build, which is a discrepancy between many mods, but I'd like it to start here, where people are going to assume this is how modding works. I don't mind if this gets rejected, but if it does, I would still appreciate feedback.

  - Updated `PostBuild.targets` to remove an extra plugins folder from the zip folder automatically made upon build

### Breaking changes
n/a

### DIFF
--- a/PostBuild.targets
+++ b/PostBuild.targets
@@ -54,7 +54,7 @@
         <MakeDir Directories="$(TemporaryBuildDir)" />
 
         <!-- Main zip file -->
-        <Copy SourceFiles="@(PluginsFiles)" DestinationFolder="$(TemporaryBuildDir)\plugins\%(RecursiveDir)" SkipUnchangedFiles="true" />
+        <Copy SourceFiles="@(PluginsFiles)" DestinationFolder="$(TemporaryBuildDir)\%(RecursiveDir)" SkipUnchangedFiles="true" />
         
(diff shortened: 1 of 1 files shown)


MrPurple6411:
there was a reason for that folder btw.

https://github.com/toebeann/subnautica-support/tree/8f62f7e0f7f42d5e17deccbfde040b7792790706?tab=readme-ov-file#how-to-make-my-mod-compatible-with-this-extension

MrPurple6411:
though I guess it looks like he accounted for it and just made it work for with and or without now.
but yea it used to be like that to make it work better with vortex
though if we ever go back to adding patchers then it will need that back.

mostly I make ALL my mods so they can just be extracted in the bepinex folder.

Law-Abiding-Developer:
ah interesting

EldritchCarMaker:
Yea, as purple mentioned, the extra folder is actually required to distinguish between patchers and plugins. It's actually better to include the extra folder, rather than assuming everything will be a plugin.

Law-Abiding-Developer:
Okay, then we should tell everyone to do that

MrPurple6411:
I did many years ago lol... noone listens lol

PR #657, opened 2025-12-16
Recipe data removal

### LINKED ISSUE #654: Unable to delete recipes
**Describe the bug**
With the changes made to accomodate the SN 2025 update, it is no longer possible to delete a recipe by setting its value to `null` via `CraftDataHandler.SetRecipeData()`.

**To Reproduce**
1. Pass `null` as the recipeData argument to `SetRecipeData`
2. Get a nullref

**Expected behavior**
The recipe should be removed from the game.

Metious:

  - Added a new `RemoveRecipeData` method
  - Setting `SetRecipeData(TechType, null)` now removes the recipe data for TechType, if it exists.

Fixes #654

### DIFF
--- a/Nautilus/Handlers/CraftDataHandler.cs
+++ b/Nautilus/Handlers/CraftDataHandler.cs
@@ -16,10 +16,16 @@ public static class CraftDataHandler
     /// <para>Can be used for existing TechTypes too.</para>
     /// </summary>
     /// <param name="techType">The TechType whose RecipeData you want to edit.</param>
-    /// <param name="recipeData">The RecipeData for that TechType.</param>
+    /// <param name="recipeData">The RecipeData for that TechType. If null and the tech type already has a recipe, it will get removed instead.</param>
     /// <seealso cref="RecipeData"/>
     public static void SetRecipeData(TechType techType, RecipeData recipeData)
     {
+        if (recipeData is null)
+        {
+            RemoveRecipeData(techType);
+            return;
+        }
+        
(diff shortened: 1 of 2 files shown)


tinyhoot (changes requested):
[L29-L30](https://github.com/SubnauticaModding/Nautilus/pull/657/files#diff-8bbb6335a67493cc66c76ab062df4a314993eed24e37d6981004cf3780d76dcfR29)
Throws a nullref when a recipe is set to null using `SetTechData(TechType, null)`

tinyhoot (commented):
The new approach fixes the NRE within the context of Nautilus, but still leaves the library vulnerable to that same exception if some mod deletes a recipe without it.

Metious:
> The new approach fixes the NRE within the context of Nautilus, but still leaves the library vulnerable to that same exception if some mod deletes a recipe without it.

Can you elaborate more with an example?

tinyhoot:
Assume two mods. Mod A does not use Nautilus and removes the battery recipe. Mod B does use Nautilus and also edits the battery recipe.

I was under the impression that it was possible for mod A to set the jsondata to `null` on its own, which would then have caused the above-mentioned NRE in Nautilus once mod B tried to touch it. But I double checked the game's code just to be sure and even if Nautilus were to throw an NRE the game's `TechData` class would do so too, so there's not too much point to guarding against it.

I have no further concerns, looks good to me.

tinyhoot (approved):

Metious:
Yeah, the NRE comes from the `JsonValue`'s Equality operator which throws if any value for its keys is null. Not an exclusive Nautilus NRE.

32Kallies (approved):

PR #665, opened 2026-02-12
SaveDataCache carries ref values across saves

### LINKED ISSUE #658: `SaveDataCache` not reset when starting a new save file
## Describe the bug
Data stored in a `SaveDataCache` subclass associated with one save file may "leak" if a new save file is created during the same game session.

## To Reproduce
1. Create/install a Nautilus mod that includes per-save data persisted in a `SaveDataCache` subclass
    * ([Pinup v1.3.2](https://codeberg.org/call_me_xale/Subnautica-Pinup/releases/tag/v1.3.2) can be used to reproduce this)
1. Launch *Subnautica*
1. Load or create a save file
1. Perform actions that populate the `SaveDataCache`
    * (E.g., for Pinup, pin a few recipes to the HUD)
1. Save and return to the main menu
1. Create a new save file

### Expected behavior
No information associated with another save file's `SaveDataCache` should be carried over into a new save.

### Observed behavior
Data stored in the `SaveDataCache` for the first save file opened or created will also appear in the second.
  * E.g., in Pinup, the player will see the same recipes pinned to the HUD after exiting the lifepod for the first time.

## Additional notes
Reproducible in both *Subnautica* and *Below Zero*.

I *suspect* this bug would also occur if the player were to (instead of creating a new save file) load an existing save file with no `SaveDataCache` associated with it, such as when launching an existing save file with a new mod installed.

Metious:

  - Fixed a bug where if you get into a save that has a save data cache, then you quit to main menu, then create a new world, the cache from the previous save would get carried to the new one.

#600 only fixed the issue for value types; this one should make it work for any member.
Fixes #658

### Builds (up-to-date with 92613e5)
- SN1: [Nautilus_SN.STABLE_1.0.0.1.49.zip](https://github.com/user-attachments/files/25263016/Nautilus_SN.STABLE_1.0.0.1.49.zip)
- BZ: [Nautilus_BZ.STABLE_1.0.0.1.49.zip](https://github.com/user-attachments/files/25263062/Nautilus_BZ.STABLE_1.0.0.1.49.zip)

### DIFF
--- a/Nautilus/Utility/JsonUtils.cs
+++ b/Nautilus/Utility/JsonUtils.cs
@@ -18,21 +18,6 @@ namespace Nautilus.Utility;
 /// </summary>
 public static class JsonUtils
 {
-    private static class Defaults
-    {
-        private static Dictionary<Type, object> _defaults = new();
-
-        public static object GetValue(Type type)
-        {
-            if (_defaults.TryGetValue(type, out var result))
-            {
-                return result;
-            }
-            
-            return _defaults[type] = Activator.CreateInstance(type);
-        }
-    }
-    
     private static string GetDefaultPath<T>(Assembly assembly) where T : class
     {
         return Path.Combine(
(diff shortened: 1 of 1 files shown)


32Kallies (approved):
LGTM

PR #667, opened 2026-03-11
Add a priority system for mod loading

Law-Abiding-Developer:

  - Changed `WaitScreenTask` to have an internal read only int for priority.
  - Force sorting of the list of load tasks from highest to lowest priority every time a task is registered
  - Added an optional int parameter when registering tasks for said priority

### Notes
There is no limit to how high (or how low) you can set the integer for priority. I'm taking suggestions on that
I could not test if mods were unable to load due to method signature change or something else; I just got a `MissingMethodException`
A test build with this priority system is included. Along with a new version of All In One Hand Held Fabricator showcasing this change
When unzipping the test builds, you will see a `_MACOSX` folder. Ignore it. I'm not sure what it is, but it's included on any and all zip files that macOS zipped.
I understand this is niche, but I don't think I'm the only one who thinks is is needed regardless
[AIOHHFTestBuild.zip](https://github.com/user-attachments/files/25920360/AIOHHFTestBuild.zip)
[NautilusTestBuild.zip](https://github.com/user-attachments/files/25920361/NautilusTestBuild.zip)


Nautilus/Handlers/WaitScreenHandler.cs
```diff
@@ -26,16 +26,19 @@ public static class WaitScreenHandler
     /// <param name="description">An optional description to give users detailed information about what your task is
     /// doing. Can be updated even while your task is executing by setting
     /// <see cref="WaitScreenTask"/>.<see cref="WaitScreenTask.Status"/>. Accepts language keys for localisation.</param>
-    public static void RegisterEarlyLoadTask(string modName, Action<WaitScreenTask> loadingFunction, string description = null)
+    /// <param name="priority">The priority at which your mod must load. The higher priority number the earlier the mod will load in the load order</param>
+    public static void RegisterEarlyLoadTask(string modName, Action<WaitScreenTask> loadingFunction, string description = null, int priority = 10)
```
EldritchCarMaker: This system has enough use that I don't think we can afford a breaking change here. Make it backwards compatible by overloading rather than using optional arguments

Nautilus/Handlers/WaitScreenHandler.cs
```diff
@@ -26,16 +26,19 @@ public static class WaitScreenHandler
...
     {
-        WaitScreenPatcher.EarlyInitTasks.Add(new WaitScreenTask(modName, loadingFunction, description));
+        WaitScreenPatcher.EarlyInitTasks.Add(new WaitScreenTask(modName, loadingFunction, description + $" - Priority {priority}!", priority));
```
EldritchCarMaker: Why are we changing the description? The user doesn't need to know an internal task's priority, there's no reason for us to change the description of the task. Add debug logging if necessary, but not front facing info.

EldritchCarMaker (changes requested):
I'd argue this PR feels largely unnecessary, as we already have three separate loading stages between early, normal, and late, but there's bare minimum changes to be made before we even consider merging imo.

Nautilus/Handlers/WaitScreenHandler.cs
Law-Abiding-Developer: Was for testing, can be removed before merging

Nautilus/Handlers/WaitScreenHandler.cs
Law-Abiding-Developer: Will do, changing now

Law-Abiding-Developer:
Request for updated test build(s) will be accepted

Indigocoder1:
Why is there an entirely different method just to add the priority tasks? I feel like an overload would be a lot cleaner for both the implementation and the user-end of things. I also don't see the point of the FreezeRegistry variable, which, by the way, is never set back to false. As for the Priority enum, I don't see why that needs to be an enum. Just make it an int.

Law-Abiding-Developer:
Hm...I just had a though that yeah those methods are uneccissary

EldritchCarMaker:
> As for the Priority enum, I don't see why that needs to be an enum. Just make it an int.

It was added as a half hearted attempt at providing standardization in priorities and timings. I think we do need standardization but that's as much on documentation as it is on implementation.

Law-Abiding-Developer:
Okay, that should do it

Indigocoder1:
> > As for the Priority enum, I don't see why that needs to be an enum. Just make it an int.
> 
> It was added as a half hearted attempt at providing standardization in priorities and timings. I think we do need standardization but that's as much on documentation as it is on implementation.

I can see the point about standardization, and I think it would be better handles with recommended values in the docs rather than an enum. That would provide users with some standard values while also making it more clear that there aren't only a few set values to choose from. If we do decide to keep the enum, then I think the capitalization should at least be fixed.

32Kallies:
Closed for further consideration at least until we work out PR #677 (67!!)

PR #668, opened 2026-03-15
Remove any and all references to QMods and QModManager in Nautilus

Law-Abiding-Developer:

  - Remove any and all references to QMods and QModManager in Nautilus

### Breaking changes

  - (not sure if this is breaking) Renamed variable in class SaveDataCache from `QModID` to `ModID`

### DIFF
--- a/Nautilus/Json/ConfigFile.cs
+++ b/Nautilus/Json/ConfigFile.cs
@@ -80,7 +80,7 @@ public abstract class ConfigFile : IJsonFile
     /// {
     ///     public KeyCode ActivationKey { get; set; } = KeyCode.Escape;
     ///     public MyConfig() : base("options", "Config Files") { }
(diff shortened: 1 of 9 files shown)


Nautilus/Json/ConfigFile.cs
```diff
@@ -80,7 +80,7 @@ public ConfigFile()
     /// {
     ///     public KeyCode ActivationKey { get; set; } = KeyCode.Escape;
     ///     public MyConfig() : base("options", "Config Files") { }
-    ///     // The config file will be stored at the path "QMods\YourModName\Config Files\options.json"
+    ///     // The config file will be stored at the path "BepInEx\plugins\YourModName\Config Files\options.json"
```
32Kallies: This is not true, it will not be stored there

32Kallies (changes requested):
Good catch, but needs a fix

32Kallies (approved):
LGTM

PR #669, opened 2026-03-17
Utility: Attribute Based Registration System

jbeast291:
  - New Utility `RegistryEventUtils` that can parse an Assembly for **static** methods attributed with `[RegisterEvent()]` and execute them. 
    - Each attribute requires a *unique ID* (including across mods). Each Attribute can depend on another based on ID (including from a separate mod) allowing control of the load order.
    - Cyclic dependencies throw and exception and the cyclic chain is logged for debugging. 
  - When methods are attributed and executed, their method parameters can be matched to a specific dependency injector. Currently, there are 3:
     - `PrefabInfoInjector` passes a `PrefabInfo` using the RegistryID of the attribute to assign a `TechType`
     - `TechTypeInjector` passes a modded `TechType` for convenient use in recipes based on the parameter name 
     - `AssetBundleAssetInjector` this injector loads assets from an asset bundle (*optionally* provided when calling `ExecuteAssemblyAttributeRegistries`) and passes them in. The parameter must be attributed with [AssetLoad] for this injector to trigger. Very useful for item icons.
  - All injectors can be overridden if desired as well. Others can create their own and pass them in to the loader to be used, but only for their assembly as to not break others. 

### Breaking changes
  - None, this is just a simple Utility to make large projects easier to manage/load items in a specific order with other mods. Only Nautilus code touched is in `WaitScreenPatcher` (A warning is logged after all load tasks in the case of unloaded registries due to missing dependencies).

### Example Usage:
  - The most basic of attributes for registering a prefab is defined like this (though this system is not limited to prefabs)
```C#
[RegisterEvent("MyCoolItem")]
private static void Register(PrefabInfo info) {}
//The prefab info passed in will be assigned the TechType "MyCoolItem"
```
  - To execute this registry, a call to `RegistryEventUtils` is required
```C#
RegistryEventUtils.ExecuteAssemblyAttributeRegistries(Assembly.GetExecutingAssembly());
/*With optional paramters to filter the namespace that will be searched and
an asset bundle for the AssetBundleAssetInjector*/
RegistryEventUtils.ExecuteAssemblyAttributeRegistries(Assembly.GetExecutingAssembly(),
      "MyModsNameSpace.Prefabs.OnlyLoadThese", MyAssetBundle);
//Ideally these are within a wait screen handler task but this is not required
```
  - Optionally injectors can be user defined in case more functionality is needed.
```C#
RegistryEventUtils.ExecuteAssemblyAttributeRegistriesCustomInjectors(Assembly.GetExecutingAssembly(), 
      [new MyEpicPrefabInjector(), new AssetBundleAssetInjector(bundle)]);
```
  - Other ways to define registries are below, showcasing the differing injectors/dependencies
      - Dependencies/execution order are supported and defined as follows
```C#
//Although there is only 1 dependency here, you can define any amount within the Attribute
[RegisterEvent("MyCoolItem", "OtherCoolItem")]
private static void Register(PrefabInfo info) {}
//The method will be executed only after "OtherCoolItem" is registered
```
 - Dependencies from other mods
```C#
[RegisterEvent("MyCoolItem", "OtherEPICItemFromSomeDifferentMod")]
private static void Register(PrefabInfo info) {}
/*The method will be executed only after "OtherEPICItemFromSomeDifferentMod" is registered. 
If it never registers in the case the other mod isn't present, MyCoolItem will never execute 
and Nautilus will log a warning after the WaitScreenHandler tasks execute*/
```
  - Cyclic/Circular dependencies are detected and throw an error
```C#
[RegisterEvent("OtherEPICItemFromSomeDifferentMod", "MyCoolItem")]
private static void Register(PrefabInfo info) {}
//it doesn't matter how large the chain is or if it goes across mods, it will be detected.
//The chain of dependencies will be logged within the exception for debugging.
```
  - Modded TechTypes can be automatically passed into the method
```C#
[RegisterEvent("MyCoolTool", "MyCoolIngredient")]
private static void Register(PrefabInfo info, TechType myCoolIngredient) {}
//The name of the parameter (case does not matter) is used to obtain the TechType.
/*Do note that while "MyCoolIngredient" is set to load before this registry, it is not required for the
TechType to successfully parse (just highly recomended/required in most cases for the TechType to be valid)*/ 
```
  - Asset Bundle assets can be loaded, assuming an AssetBundle was provided when executing the Utility and [AssetLoad] is attributed to the method parameter
```C#
//There are 2 ways to define this, either using the parameter name as the asset to load
[RegisterEvent("MyCoolResource")]
public static void Register(PrefabInfo info, [AssetLoad] Sprite myCoolIcon)
//OR using an optional parameter of the attribute, the parameter name is then irrelevant
[RegisterEvent("MyCoolResource")]
public static void Register(PrefabInfo info, [AssetLoad("MyCoolIcon")] Sprite icon)
```


jbeast291:

Nautilus/Utility/AttributeRegistration/Injectors/PrefabInfoInjector.cs
```diff
@@ -0,0 +1,38 @@
...
+using System.Reflection;
+using Nautilus.Assets;
+
+namespace Nautilus.Utility.AttributeRegistrationUtils.Injectors;
+
+
+/// <summary>
+/// Represents an injector that checks for <see cref="PrefabInfo"/> as a parameter type
+/// </summary>
+public sealed class PrefabInfoInjector : IDependencyArgumentInjector
```
EldritchCarMaker: Would a CustomPrefab injection be more useful? Many of our prefabs only use PrefabInfo as an intermediary to be given to a CustomPrefab instance anyway

Nautilus/Utility/AttributeRegistrationUtils/Injectors/TechTypeInjector.cs
```diff
@@ -0,0 +1,50 @@
...
+    {
+        if(argument.ParameterType == typeof(TechType))
+        {
+            if(EnumHandler.TryGetValue(argument.Name, out TechType techType))
+            {
+                result = techType;
+                return true;
+            }
+            
+            bool techTypeIDPresentInDependencies = attribute.loadAfterIDs.Any(loadAfterID => loadAfterID.ToLower() == attribute.registryID.ToLower());
```
EldritchCarMaker: This doesn't look right to me. This is checking if any of the loadafter ids are the same as the current registry id? Did you mean to check against the techtype argument name?

Example mod/RegisterServiceExamples.cs
```diff
@@ -0,0 +1,103 @@
...
+         IMPORTANT: if your Registry's iD is the exact same from another mod,
+         your registry may fail to execute. Ensure you give it a unique name that won't cause,
+         any conflicts with other mods using the system. In most cases, if your TechType is
+         unique, you won't have to worry about this.
+         */
+    }
+    
+    //Register Events can depend on 1 (or multiple) registries to be loaded before it.
+    [RegisterEvent("MyEpicItem", "MyCoolItem")]
+    private static void RegisterMyEpicItem(PrefabInfo info, TechType myCoolItem)
```
EldritchCarMaker, line 35: Mention the TechType argument injection, clarify where it's coming from. These are examples for new modders who wouldn't know what's supplying that techtype

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// Nautilus' injectors can be instantiated, or you can create your own that implement <see cref="IDependencyArgumentInjector"/>
+    /// </summary>
+    /// <param name="assemblyToSearch">Assembly to search through. For <i>most</i> mods, the result of <see cref="Assembly.GetExecutingAssembly"/>> should be passed as the argument.
+    ///  Every Type within the Assembly will be checked for methods containing the Attribute <see cref="RegisterEventAttribute"/>.</param>
+    /// <param name="injectors">List of injectors to use during parsing. Feel free to implement <see cref="IDependencyArgumentInjector"/> in your own class to extend
+    /// the functionality of this registration system.</param>
+    /// <param name="namespaceFilter">Namespace to search and execute. If null/blank the entire Assembly will be searched.
+    ///  Should be formatted like <i>"Nautilus.SomeSubNamespace.SubSubNamespace"</i> but with your mod's namespace</param>
+    public static void ExecuteAssemblyAttributeRegistriesCustomInjectors(Assembly assemblyToSearch, IDependencyArgumentInjector[] injectors, string namespaceFilter = null)
```
EldritchCarMaker: It feels like there should be an easier way to add custom injectors while retaining the default injectors. But I can't immediately think of what to do for that. Maybe an optional bool argument for including default injectors? Alternatively turn this into an overload for ExecuteassemblyAttributeRegistries and have the CustomInjectors method also include default ones.

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttribute.cs
```diff
@@ -0,0 +1,33 @@
...
+    /// <summary>
+    /// List of IDs that should be loaded before this one. Can be from a different registry execute call or mod.
+    /// </summary>
+    public readonly string[] loadAfterIDs = loadAfterIDs;
+    
+    /// <summary>
+    /// The associated method with this attribute.
+    /// </summary>
+    /// <remarks>NOTE: the methodInfo is null until a <see cref="RegisterEventAttributeLoader"/> parses this attribute.</remarks>
+    public MethodInfo methodInfo { internal set; get; }
```
EldritchCarMaker: Internal

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    /// <param name="assemblyToSearch">Assembly to search through. For <i>most</i> mods, the result of <see cref="Assembly.GetExecutingAssembly"/>> should be passed as the argument.
+    ///  Every Type within the Assembly will be checked for methods containing the Attribute <see cref="RegisterEventAttribute"/>.</param>
+    /// <param name="namespaceFilter">Namespace to search and execute. If null/blank the entire Assembly will be searched.
+    ///  Should be formatted like <i>"Nautilus.SomeSubNamespace.SubSubNamespace"</i> but with your mod's namespace</param>
+    /// <param name="assetBundle">When using the <see cref="AssetBundleAssetInjector"/>>, the injector will pull from this asset bundle.
+    ///  This can safely be left null if you don't load assets through argument parameters</param>
+    public static void ExecuteAssemblyAttributeRegistries(Assembly assemblyToSearch, string namespaceFilter = null, AssetBundle assetBundle = null)
+    {
+        RegisterEventAttributeLoader.ExecuteAssemblyAttributeRegistries(assemblyToSearch, namespaceFilter, assetBundle);
```
EldritchCarMaker: Should probably just call CustomInjectors and supply the default ones (and if given an asset bundle instantiate AssetBundleAssetInjector to add)

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
```diff
@@ -0,0 +1,259 @@
...
+    {
+        foreach (IDependencyArgumentInjector injector in injectors)
+        {
+            if (injector.InjectorTargetType() != null)
+                _typedDependencyArgumentInjectors.Add(injector.InjectorTargetType(), injector);
+            _dependencyArgumentInjectors.Add(injector);
+        }
+    }
+    
+    internal static void ExecuteAssemblyAttributeRegistries(Assembly assemblyToSearch, string namespaceFilter = null, AssetBundle assetBundle = null)
```
EldritchCarMaker: Read comment in Utils, but I think this method shouldn't exist. It's basically just an internal wrapper for the Custom version, and wrappers aren't necessary if everything's internal like this.

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
```diff
@@ -0,0 +1,259 @@
...
+        RegisterEventAttributeLoader loader = new(assemblyToSearch, namespaceFilter);
+        loader.RegisterInjectors(injectors);
+        loader.ProcessAttributeRegistriesForAssembly();
+    }
+
+    private void ProcessAttributeRegistriesForAssembly()
+    {
+        foreach (Type type in _assemblyToSearch.GetTypes())
+        {
+            if (_namespaceToSearch != null && type.Namespace != null && !type.Namespace.StartsWith(_namespaceToSearch))
```
EldritchCarMaker: So what was the necessity of this? Because in 99% of cases the assembly would be the same namespace, and this is the only kind of namespace checking we have in nautilus that I've seen?

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
```diff
@@ -0,0 +1,259 @@
...
+            
+            firstUnloadedDependency = dependentID;
+            return false;
+        }
+        firstUnloadedDependency = null;
+        return true;
+    }
+    
+    //NOTE: when calling ensure the method is ready to execute, dependency checks don't happen here
+    private void ExecuteMethod(RegisterEventAttribute attribute)
```
EldritchCarMaker: Maybe name it UnsafeExecuteRegister or something similar? ExecuteMethod is nondescript, and if you have to add a comment saying it's unsafe, just put it in the method name.

Nautilus/Utility/AttributeRegistration/Injectors/TechTypeInjector.cs
```diff
@@ -0,0 +1,50 @@
...
+public sealed class TechTypeInjector : IDependencyArgumentInjector
+{
+    /// <summary>
+    /// Checks whether the argument is of type <see cref="TechType"/>. 
+    /// </summary>
+    /// <param name="attribute">Attribute attached to the method with.</param>
+    /// <param name="argument">Parameter argument to check and inject for.</param>
+    /// <param name="result">The result of the injection (can be safely cast to a <see cref="TechType"/>). Is null when this method returns false.</param>
+    /// <returns>True if there is a valid injection for the current argument, otherwise false.</returns>
+    /// <exception cref="Exception"> is thrown when there is no TechType for the associated argument</exception>
```
EldritchCarMaker: Exception should probably be an invalid argument exception, rather than just basic Exception

EldritchCarMaker (changes requested):
Many suggestions, all are minor. Foundation of the system looks good to me, but would like to hold off on merging for some extra feedback. Want this system to be as good as possible

Nautilus/Utility/AttributeRegistration/Injectors/PrefabInfoInjector.cs
jbeast291: I'm open to it but if we go this route we need an easy way to set the PrefabFileName as well. Kallie's structure helper organizes based on it and it can only be changed in Info I believe.

Maybe some kind of optional attribute for the parameter like [Prefab("MyMod/MyFolder/MyItem")] for the CustomPrefab? We could also throw more utilities into that attribute like loading a sprite from an asset bundle optionally as well.

Nautilus/Utility/AttributeRegistrationUtils/Injectors/TechTypeInjector.cs
jbeast291: oh yes, good catch

Example mod/RegisterServiceExamples.cs
jbeast291: it does specify below where i describe what the method is doing and how Nautilus will supply the techtype but I suppose the comment heading could also specify a bit of that.

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
jbeast291: Fair point, but i want to offer the flexibility or replacing default injectors if desired.
We could do 3 methods: default Nautilus Injectors, Default Injectors with overrides (if one targeting the same type is defined, it will replace the Nautilus default), and then full custom where every injector desired is defined.

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttribute.cs
jbeast291: no, when someone needs to make a custom injector they may desire to get additional information about the method (additional attributes that mark things maybe). If this is internal there would be no way for a custom injector to get this.

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
jbeast291: will do

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
jbeast291: agreed

EldritchCarMaker:
Wrong button, my bad

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
jbeast291: fair point, will do

Nautilus/Utility/AttributeRegistration/Injectors/TechTypeInjector.cs
jbeast291: will do

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
EldritchCarMaker: How about we combine the first two? So one method for default, with optional overrides, and a second for no defaults. Although I wonder if it might be worth it to allow multiple injectors to handle the same type? Not sure though. Could get messy without a strict priority system.

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttributeLoader.cs
jbeast291: I originally thought It could be used to register things for the main menu and the game at separate times but now that I am thinking that over, it makes less sense now. Considering it would have to be run everytime the main menu is opened, the ID's would already be registered so they would fail within the nautilus system. I'll remove it

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttribute.cs
EldritchCarMaker: Fair point, internal protected then? Public is simply too open for this imo.

Nautilus/Utility/AttributeRegistration/Injectors/PrefabInfoInjector.cs
EldritchCarMaker: Sprite feels a touch unnecessary when we already have an asset bundle injection, but maybe I'd just have to see it to know for sure. On second thought, yea, maybe PrefabInfo is best. Because it's a struct it can make it hard to work with if you try to access it from further downstream (like through CustomPrefab instance).

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttribute.cs
jbeast291: Turns out I was wrong. internal protected wouldn't fix the issue but we don't need to expose this outside Nautilus Anyway. A reference to `MethodInfo` can be obtained through `ParameterInfo.Member` casted to `MethodInfo` apparently. Since we pass a `ParameterInfo` to injectors already, there is no need for injectors to use the attributes. Will make it internal

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
jbeast291: Resolved in latest fixes. `ExecuteAssemblyAttributeRegistries` takes optional overrides that can either override the defaults/add onto the injector list if they don't override a type.

Nautilus/Utility/AttributeRegistration/Injectors/PrefabInfoInjector.cs
jbeast291: seems to be a bad idea/too restrictive

jbeast291:
~Quick~ question, do we want to handle optional registries that add features between mods. Lets say there is both Mod A and Mod B. Mod B adds a new item and mod A wants to add another that is an upgrade or something of that, depending on Mod B's item. However, if Mod B is never present/never registers that registry, Nautilus will send a console warning and the registration will just sit there. Not the end of the world but it seems jank to me. 
There is no way currently to tell nautilus to only register if Mob B is present.

One way we could fix this is making the RegisterEvent require another mod to execute the event. Would have to remove `params` but it would look something like this:
```C#
[RegisterEvent("MyModdedSubUpgrade", ["AnotherModdsPrefabSubUpgrade"], ["com.group.anothermod"])]
```
The GUID of that other mod means that Nautilus will only register the event IF the other mod is present in the chainloader, otherwise it is ignored. This could also make compatibility checks easier for optional dependencies, where people could just have their compat initialization inside a registry event.

Syntax wise i think is fine but my problem is it seems logically confusing. We have a registry ID then requires other registry IDs that are required (that if missing, nautilus sends a warning), then Mod GUID's that are required but if not present nautilus completely ignores the registry. Would be convenient but strange for new modders imo to wrap their head around.

Another method could be to add a second attribute

```C#
[RegisterIfGuidPresent("com.group.anothermod")]
[RegisterEvent("MyModdedSubUpgrade", "AnotherModdsPrefabSubUpgrade")]
```
This to me seems cleaner syntax wise as well but also makes more sense being separated out. Downside is its more code and idk how to feel about adding attributes on attributes in the future.

How should we proceed?

jbeast291:
In the latest push I have added a new interface that attributes can implement: 
```C#
public interface IRegistryRequirement {
    public bool RequirementsMet();
}
```
This interface, when implemented by an attribute, can be attached to a `[RegistryEvent()]` to either ignore or execute the registry. This is extendable by other mods if they want their own functionality but the built in reason for its existence is to optionally register items based on other mods.

For example, if lets say Mod A has a soft dependency for mod B and adds some items when mod B is installed but still functions without it, it can define the following registry.
```C#
[RequireGuid("com.groupName.modB")]
[RegisterEvent("modAItem", "ModBItem")]
public static void Register(PrefabInfo info, TechType modBItem)
{
    //Then this item only exists when ModB is present and can use its techTypes and what not
}
```
Nautilus will only register when every `IRegistryRequirement` is met. If one returns false, then it is ignored and no warning is thrown (before you could ignore the warnings but these explicit requirements better state the intentions of the registry)

Since other mods can create additional attributes that implement `IRegistryRequirement`, they can make whatever niche combination of requirements a mod might need in the future. But I'm open to more built in Nautilus attributes if we can think of any generic enough.

Nautilus/Utility/AttributeRegistrationUtils/Injectors/AssetBundleAssetInjector.cs
```diff
@@ -0,0 +1,52 @@
...
+        //fallback to name of argument
+        value = bundle.LoadAsset(arg.Name, arg.ParameterType);
+        return true;
+    }
+    
+    /// <returns>Returns the result of typeof(<see cref="AssetLoadAttribute"/>)</returns>
+    public Type InjectorTargetType()
+    {
+        return typeof(AssetBundle);
+    }
```
32Kallies: Either rename this method to use a verb in the name, or convert it to a property. I would suggest making it a property.

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+using Nautilus.Assets.Gadgets;
+using Nautilus.Crafting;
+using Nautilus.Utility.AttributeRegistrationUtils;
+
+namespace Nautilus.Examples;
+[BepInPlugin("com.snmodding.nautilus.registerevent", "Nautilus RegisterEvent Example Mod", Nautilus.PluginInfo.PLUGIN_VERSION)]
+[BepInDependency("com.snmodding.nautilus")]
+public class RegisterEventExamples  : BaseUnityPlugin
+{
+    //Register Events allow you to define Prefabs with less code
```
32Kallies:
```suggestion
    // Register Events allow you to define Prefabs with less code
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+        /*
+         IMPORTANT: if your Registry's iD is the exact same from another mod,
+         your registry may fail to execute. Ensure you give it a unique name that won't cause,
+         any conflicts with other mods using the system. In most cases, if your TechType is
+         unique, you won't have to worry about this.
+         */
+    }
+    
+    //Register Events can depend on 1 (or multiple) registries to be loaded before it.
+    //Registers can also supply you modded TechType's with ease
```
32Kallies:
```suggestion
    // Register Events can depend on 1 (or multiple) registries to be loaded before it.
    // Registers can also supply you modded TechType's with ease
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+                    /*
+                     We can set the crafting here to use a modded TechType without having to
+                     touch EnumHandler
+                     */
+                    new Ingredient(myCoolItem, 1)
+                }
+            })
+            .WithFabricatorType(CraftTree.Type.Fabricator)
+            .WithStepsToFabricatorTab("Personal", "Equipment");
+        //Although this can be crafted you won't be able to pick it up due to missing components
```
32Kallies:
```suggestion
        // Although this can be crafted you won't be able to pick it up due to missing components
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+         */
+        /*
+         NOTE: You can also depend on registries from other mods to load first. Simply put
+         whatever their registry uses as an ID into your dependency list and Nautilus will
+         handle the rest. This is also why your IDs have to be unique, as conflicts with other
+         mods can occur.
+         */
+    }
+    
+    //Now, to have Nautilus execute your registries, you must use the RegistryEventUtils
```
32Kallies:
```suggestion
    // Now, to have Nautilus execute your registries, you must use the RegistryEventUtils
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+    private void Awake()
+    {
+        //Here we ask Nautilus to search our assembly and execute our registries
```
32Kallies:
```suggestion
        // Here we ask Nautilus to search our assembly and execute our registries
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+        RegistryEventUtils.ExecuteAssemblyAttributeRegistries(Assembly.GetExecutingAssembly());
+        
+        /*
+         If you need more debug info, enable Nautilus's Debug toggle in settings.
+         RegistryEventUtils will log every time a registry is executed.
+        */
+    }
+    
+    //Bonus Info Below
```
32Kallies:
```suggestion
    // Bonus Info Below
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+
+    /*Register Events are not limited to prefabs however
+     they can execute anything and are useful in managing load order of your mod*/
+    [RegisterEvent("MyCoolSounds", "SomeRegistryIWantToLoadFirst")]
+    private static void RegisterMySounds()
+    {
+        // code for loading sounds that load after something else or whatever
+    }
+    
+    //Register Events can also load assets from an asset bundle (provided to Nautilus in a later step)
```
32Kallies:
```suggestion
    // Register Events can also load assets from an asset bundle (provided to Nautilus in a later step)
```

Example mod/RegisterEventExamples.cs
```diff
@@ -0,0 +1,108 @@
...
+    /* Though this is outside the scope of a basic tutorial, the code below should get you started */
+    
+    /*[RegisterEvent("MyEpicItem")]
+    private static void RegisterItemWithAsset(PrefabInfo info, [AssetLoad("myBundleAssetIcon")] Sprite icon)
+    {
+        info.WithIcon(icon);
+        CustomPrefab myPrefab = new CustomPrefab(info);
+        myPrefab.Register();
+        //Commented out as no asset bundle is present, will throw an exception if the asset fails to load
```
32Kallies:
```suggestion
        // Commented out as no asset bundle is present, will throw an exception if the asset fails to load
```

Nautilus/Utility/AttributeRegistration/RegisterEventAttribute.cs
```diff
@@ -0,0 +1,34 @@
+using System;
+using System.Reflection;
+
+namespace Nautilus.Utility.AttributeRegistrationUtils;
```
32Kallies: I would rename the `AttributeRegistrationUtils` folder.

The world Utils is redundant. AttributeRegistration is an alright name. Or RegistrationAttributes? I think I prefer **RegistrationAttributes**.

Nautilus/Utility/AttributeRegistrationUtils/RegisterEventAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+namespace Nautilus.Utility.AttributeRegistrationUtils;
+
+
+/// <summary>
+/// Attributes a method for use with the <see cref="RegisterEventAttributeLoader"/>.
+/// </summary>
+/// <param name="iD"><b><i>Unique</i></b> ID for this registration load event. Failure to have a unique ID (including between other mods) will result in errors</param>
+/// <param name="loadAfterIDs">List of IDs that should be loaded before this one. Can be from a different registry execute call or mod.</param>
+[AttributeUsage(AttributeTargets.Method)]
+public sealed class RegisterEventAttribute(string iD, params string[] loadAfterIDs) : Attribute
```
32Kallies: I would probably change `iD` to `id` for consistency.

Nautilus/Utility/AttributeRegistration/RegisterAttribute.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace Nautilus.Utility.AttributeRegistrationUtils;
+
+
+/// <summary>
+/// Attributes a method for use with the <see cref="RegisterEventAttributeLoader"/>.
+/// </summary>
+/// <param name="iD"><b><i>Unique</i></b> ID for this registration load event. Failure to have a unique ID (including between other mods) will result in errors</param>
+/// <param name="loadAfterIDs">List of IDs that should be loaded before this one. Can be from a different registry execute call or mod.</param>
+[AttributeUsage(AttributeTargets.Method)]
```
32Kallies, line 12: If it is fine to use Rider features, then it would be amazing to add the `[MeansImplicitUse]` attribute so that the IDE marks any methods using the attribute as in-use.

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
```diff
@@ -0,0 +1,69 @@
...
+using Nautilus.Utility.AttributeRegistrationUtils.Injectors;
+using UnityEngine;
+
+namespace Nautilus.Utility.AttributeRegistrationUtils;
+
+
+/// <summary>
+/// Utilities for searching for and calling methods attributed with <see cref="RegisterEventAttribute"/>
+/// </summary>
+public static class RegistryEventUtils
```
32Kallies: This class name might be a bit vague. It honestly just seems soulless to me, like something that blends into the void of built-in C# features. Something more memorable maybe? AutomaticRegistrationUtils? AttributeRegistrationUtils? That's just the same as the current namespace though.

32Kallies (changes requested):
Various naming and styling suggestions and nitpicks for before I do a full review

Nautilus/Utility/AttributeRegistrationUtils/RegistryEventUtils.cs
jbeast291: i like AttributeRegistrationUtils

Nautilus/Utility/AttributeRegistration/AttributeRegistrationUtils.cs
```diff
@@ -0,0 +1,69 @@
...
+    /// </list>
+    /// If your looking to define your own injectors, please use: <see cref="ExecuteAssemblyAttributeRegistriesCustomInjectors"/>
+    /// </summary>
+    /// <param name="assemblyToSearch">Assembly to search through. For <i>most</i> mods, the result of <see cref="Assembly.GetExecutingAssembly"/>> should be passed as the argument.
+    ///  Every Type within the Assembly will be checked for methods containing the Attribute <see cref="RegisterEventAttribute"/>.</param>
+    /// <param name="assetBundle">When using the <see cref="AssetBundleAssetInjector"/>>, the injector will pull from this asset bundle.
+    ///  This can safely be left null if you don't load assets through argument parameters</param>
+    /// <param name="overrideInjectors">Optional injectors to override the defaults and or add new injectors to the default list. If you desire more control,
+    ///  look into <see cref="ExecuteAssemblyAttributeRegistriesCustomInjectors"/></param>
+    public static void ExecuteAssemblyAttributeRegistries(Assembly assemblyToSearch, AssetBundle assetBundle = null, IDependencyArgumentInjector[] overrideInjectors = null)
```
EldritchCarMaker: What are thoughts on support for multiple bundles within a single mod? Larger mods often split their assets up into separate bundles (for example rota has a separate bundle exclusively for the garg, and cotv has separate bundles for creatures from the rest).

32Kallies: Necessary for the PR. TRP also uses multiple bundles.

jbeast291: I had not thought about this but I agree, def an issue.

I think having the user direct what asset bundle to load from would get too verbose. This is an example of what I DONT want:
```C#
[RegisterEvent("MyPrefab")]
private static void Register(PrefabInfo info, [AssetLoad("SomeBundleDifferentiator", "MyCoolIcon")] Sprite icon) { }
//This to me looks terrible and if this is required, it would be easier to just have a static ref to the bundle somewhere and load assets yourself.
``` 
So we are left with Nautilus having to decide what bundle to load from. When the user calls `ExecuteAssemblyAttributeRegistries` they will have the option to pass an array of asset bundles. From there, when the AssetBundleAssetInjector is created, it will read all of the asset names from each bundle and store them in a dictionary like this: `Dictionary<string, AssetBundle> assets`

Then when an asset is requested, Nautilus will be able to efficiently search the bundles and determine which bundle to load from. Downside being that every asset name MUST be unique as to not cause collisions but I'm not sure why people would have duplicate assets between bundles to begin with.

Will work on this when I get a chance but lmk if you think there are any problems with this implementation

EldritchCarMaker: I think caching ALL of the assets names is overengineering it. We can likely just go one by one, try load the asset with that name, and if it returns nothing (iirc asset bundle loading doesn't throw, only returns null) then just try the next bundle.

jbeast291: Fair point, Unity likely is using some kind of efficient hash based lookup internally anyway. Ill go with your idea

32Kallies (approved):
Read over EVERYTHING and it looks great. Already pushed any minor adjustments for edge cases and documentation.

32Kallies:
> [!WARNING]
> PR MERGE INCOMING
> Merge will commence [2 days after this comment](https://www.tickcounter.com/countdown/10460905/attribute-registration-merge).

jbeast291:
Notes of Metious' Requested changes:
- Singleton system.
   - If no injector is valid for a type, but the type has a singleton within the service, then give that singleton. Saves from making an injector that just injects a singleton for end users.
   - Injectors can also pull singletons. Maybe the IRegistryRequirements should get some context as well for future proofing
   - If multiple singletons are added for a type, the last one is given in the injector. But a list/array can be retrieved by some method call in order of registration.
   - Keyed singletons should also be an option. If a keyed singleton is added, it can be retrieved from the service through its key.
- Make the service constructor take `BaseUnityPlugin` (over a string) and extract the mod's GUID from there. In the future we can pull more from this
- Asset Bundle Injector:
   - Injector should pull the bundles from the singleton system, will need to get this from the `InjectionContext`
   - If only 1 `AssetBundle` is registered as a singleton, it will be used and no differentiation required. If multiple bundles are registered, they can only be retrieved through keyed singletons, meaning the `[AssetLoad]` must define the key for the asset. If the desired bundle key couldn't be found, then exception.
   - Optionally, a callback can be added to the `AssetBundle` Injector. It will supply the `InjectionContext` and require the asset bundle key back. End users can extract the methodInfo and whatnot from their to do their own filtering.
       - If a callback is added, it will take precedence over whatever is in the `[AssetLoad]` for the bundle name. Even if the asset bundle name is defined in the `AssetLoad`, it will be ignored.

Nautilus/Utility/AttributeRegistration/Injectors/AssetBundleAssetInjector.cs
```diff
@@ -0,0 +1,70 @@
...
+    private object LoadFromBundle(string bundleName, string assetName, Type assetType, InjectionContext context)
+    {
+        if (_bundles.TryGetValue(bundleName, out AssetBundle bundle))
+        {
+            return bundle.LoadAsset(assetName, assetType);
+        }
+        throw new InjectorException(context, $"Unknown AssetBundle: {bundleName}! The bundle name should match the internal name set within UnityEditor.");
+    }
+
+    private object LoadBruteForce(string assetName, Type assetType)
```
32Kallies: The easiest compromise here in my opinion would be checking all bundles for duplicates and throwing an exception if it finds more than one match.

I simply don't feel comfortable shipping the brute force method as-is. Again, performance isn't really the concern. It's the potential for the 'wrong' thing to be found and code to continue executing with a lack of feedback that concerns me.

32Kallies (changes requested):

Metious (approved):
LGTM

RamuneNeptune:
> LGTM

Traitor

Nautilus/Utility/AttributeRegistration/RegisterAttribute.cs
```diff
@@ -0,0 +1,24 @@
...
+{
+    /// <summary>
+    /// <b><i>Unique</i></b> ID for this registration load event. Failure to have unique IDs will result in errors.
+    /// </summary>
+    public string RegistryID { get; } = id;
+    
+    /// <summary>
+    /// List of IDs that should be loaded before this one. Can be from a different registry execute call.
+    /// </summary>
+    public string[] LoadAfterIDs { get; } = loadAfterIDs;
```
EldritchCarMaker, line 23: Why is this part of the RegisterAttribute rather than a separate IRegistryRequirement?

Nautilus/Utility/AttributeRegistration/RegisterAttributeServiceAPI.cs
```diff
@@ -0,0 +1,127 @@
...
+        AddSingleton(inst);
+        _keyedSingletons.Add(typeKey, inst);
+    }
+
+    /// <summary>
+    /// Retrieves the last added singleton instance for a given type within the service. EX: if multiple instances were added to the service for the same type, the last one will be returned here.
+    /// </summary>
+    /// <typeparam name="T">Singleton Type to retrieve</typeparam>
+    /// <returns>The last added singleton for a given type. If no singleton was added for the type, null (or default for valuetypes) is returned.</returns>
+    public T GetSingleton<T>()
```
EldritchCarMaker: Maybe GetLatestSingleton would be more clear how it interacts with multiple singletons of a type?

Nautilus/Utility/AttributeRegistration/RegisterAttributeServiceExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// If multiple asset bundles are used, the register MUST determine which bundle to load from. There are 2 ways to accomplish this,
+        /// both using the key from the <see cref="RegisterAttributeService.AddKeyedSingleton">AddKeyedSingleton</see>:
+        /// <list type="bullet">
+        ///   <item>The <see cref="AssetLoadAttribute">[AssetLoad("AssetName", "BundleKey")]</see> attribute has an optional argument for the bundle key.</item>
+        ///   <item>An <see cref="IAssetBundleKeyResolver"/> can be added as a <see cref="RegisterAttributeService.AddSingleton">Singleton</see> to the service.
+        ///   Ensure the type <see cref="IAssetBundleKeyResolver"/> is used for the singleton and NOT the implementing class. If a resolver is added to the service,
+        ///   it will take precedence over anything defined in a <see cref="AssetLoadAttribute">[AssetLoad]</see> attribute</item>
+        /// </list>
+        /// </summary>
+        public void AddAssetBundleAssetInjector() =>
```
EldritchCarMaker, line 44: Utility overload that includes a bundle argument and forwards to AddSingleton? Optionally multiple bundles but maybe one is best for convenience argument.

EldritchCarMaker (approved):
Very few complaints, very good looking overall. Only minor feedback

Nautilus/Utility/AttributeRegistration/RegisterAttribute.cs
jbeast291: I didn't think about this but I do agree that this prob should be resolved. The only issue is that `IRegistryRequirements` currently work as follows:

- If a single one of a registry's `IRegistryRequirements` returns false (assuming they exist), then the registry is *essentially* ignored completely. It's not added to some dependency check later

While I agree it would be better to somehow use the `IRegistryRequirements` system so that others could expand similar functionality, I'm not sure how to do that given the current setup.

Nautilus/Utility/AttributeRegistration/RegisterAttributeServiceExtensions.cs
jbeast291: I feel like this might be unnecessary. I'd rather people interact with the singleton system so they can get a feel for what is going on at least.

it might be very confusing for a person to go from 1 bundle to multiple if they were originally using this overload. It also is literally only saving 1 line...

EldritchCarMaker: That's fair enough, I can agree with that

Nautilus/Utility/AttributeRegistration/RegisterAttribute.cs
32Kallies: I think I understand the suggestion, but that's already basically a core feature of the library with determining the order of registration, and that existing interface wouldn't really work at all. These strings here are solely used to determine the registration order.

32Kallies (approved):
Looks great. The only comments I have would be regarding the potential for future features, and after checking over, it seems sufficiently modular to the point that I don't see anything that would prevent us from adding said things down the line.

PR #685, opened 2026-05-14
feat: Add helpers to check primary/secondary binding states independently

RamuneNeptune:

- Allows you to check the users primary and/or secondary binding states *independently* for a single GameInput.Button
- Previously you could only check if the button itself was down/held/etc. and that would be true if the primary or secondary was down with no distinction

### Breaking changes

  - None, these are new optional helper methods

### DIFF
--- a/Nautilus/Handlers/GameInputHandler.cs
+++ b/Nautilus/Handlers/GameInputHandler.cs
@@ -1,4 +1,6 @@
-﻿namespace Nautilus.Handlers;
+using System.Collections.Generic;
+
+namespace Nautilus.Handlers;
 
 /// <summary>
 /// A handler class for dealing with the game's UnityEngine.InputSystem implementation.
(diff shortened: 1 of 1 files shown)


RamuneNeptune:
Pending some docs (not my forte)

Example usage:
```cs
bool modulePrimaryHeld = InputUtils.GetButtonHeld(SeaglideUpgradesModules.OpenModuleStorage, GameInput.BindingSet.Primary);
bool moduleSecondaryHeld = InputUtils.GetButtonHeld(SeaglideUpgradesModules.OpenModuleStorage, GameInput.BindingSet.Secondary);
bool modulePrimaryDown = InputUtils.GetButtonDown(SeaglideUpgradesModules.OpenModuleStorage, GameInput.BindingSet.Primary);
bool moduleSecondaryDown = InputUtils.GetButtonDown(SeaglideUpgradesModules.OpenModuleStorage, GameInput.BindingSet.Secondary);

// If both primary and secondary binding pressed at same time, do
if(modulePrimaryHeld && moduleSecondaryHeld && (modulePrimaryDown || moduleSecondaryDown) && moduleStorage != null && !moduleStorage.open)
{
    moduleStorage.Open();
}
```

Nautilus/Handlers/GameInputHandler.cs
```diff
@@ -715,4 +717,168 @@ public static class Gamepad
...
+    {
+        if(bindingSet != GameInput.BindingSet.Secondary)
+            return bindingSet;
+
+        var secondaryBindingPath = GameInput.GetBinding(device, button, bindingSet);
+
+        return string.IsNullOrWhiteSpace(secondaryBindingPath) ? GameInput.BindingSet.Primary : GameInput.BindingSet.Secondary;
+    }
+
+    private static bool IsBindingPressed(string bindingPath)
```
32Kallies, line 863: This method is basically reinventing some sort of wheel, but if it's absolutely necessary for what you are accomplishing here, I suppose it's fine

Nautilus/Handlers/GameInputHandler.cs
```diff
@@ -715,4 +717,168 @@ public static class Gamepad
             public static readonly string RightTrigger = "<Gamepad>/rightTrigger";
         }
     }
+
+    private struct BindingState
```
32Kallies, line 721: Any reason this isn't a class? Seems like they are already cached anyway and are treated more like a class than a singular data type.

Nautilus/Handlers/GameInputHandler.cs
```diff
@@ -715,4 +717,168 @@ public static class Gamepad
...
+        public int lastUpdatedFrame;
+        public string bindingPath;
+        public bool previousPressed;
+        public bool currentPressed;
+    }
+
+    private static readonly Dictionary<(GameInput.Device device, GameInput.Button button, GameInput.BindingSet bindingSet), BindingState> _bindingStates = [];
+
+    /// <summary>
+    ///
```
32Kallies, line 732: Looking for at least some documentation, it's a little unclear at a glance even to me

32Kallies (changes requested):
I can definitely see the value in something like this, though it is pretty cursed and doesn't seem to be quite done yet (in terms of docs at least)

PR #687, opened 2026-05-25
Add FMOD asset duration retrieval methods

Indigocoder1:

  - Add `GetFmodAssetDuration` and `GetMultiFmodAssetDuration` methods in `AudioUtils` for retrieving the length of an FMOD asset's audio clip.
  - Add `internal Sound[] GetSoundClips()` method to the `IFModSound` interface

### Breaking changes

  - None

### DIFF
--- a/Nautilus/FMod/Interfaces/IFModSound.cs
+++ b/Nautilus/FMod/Interfaces/IFModSound.cs
@@ -13,4 +13,6 @@ public interface IFModSound
     /// <param name="channel">The channel on which the sound was created.</param>
     /// <returns>If the sound was reported as played.</returns>
     bool TryPlaySound(out Channel channel);
+
+    internal Sound[] GetSoundClips();
 }
--- a/Nautilus/Utility/AudioUtils.cs
+++ b/Nautilus/Utility/AudioUtils.cs
@@ -5,6 +5,8 @@ using System.Runtime.InteropServices;
 using FMOD;
 using FMOD.Studio;
 using FMODUnity;
+using Nautilus.FMod;
+using Nautilus.Patchers;
 using UnityEngine;
 
 namespace Nautilus.Utility;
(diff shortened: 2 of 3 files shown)


Nautilus/Utility/AudioUtils.cs
```diff
@@ -142,6 +144,66 @@ public static FMODAsset GetFmodAsset(string path, string id = null)
...
+        if (CustomSoundPatcher.CustomSounds.TryGetValue(asset.id, out var sound) || 
+            CustomSoundPatcher.CustomSounds.TryGetValue(asset.path, out sound))
+        {
+            result = sound.getLength(out length, timeUnit);
+        }
+        else
+        {
+            // Not a custom sound
+            FMODUWE.GetEvent(asset).getChannelGroup(out var group);
+            group.getChannel(channelIndex, out var channel);
```
Indigocoder1, line 176: One thing to note here is that I'm not entirely sure what this `channelIndex` actually means. I'm assuming it's if a vanilla sound is played multiple times on different channels, but I don't really know how to test that theory

32Kallies (approved):

Nautilus/FMod/Interfaces/IFModSound.cs
```diff
@@ -13,4 +13,6 @@ public interface IFModSound
     /// <param name="channel">The channel on which the sound was created.</param>
     /// <returns>If the sound was reported as played.</returns>
     bool TryPlaySound(out Channel channel);
+
+    internal Sound[] GetSoundClips();
```
32Kallies, line 17: Are we certain that `internal` works in the way that is assumed here for interface members? This is a public interface that I am certain some mods have implemented. If it causes them to cease to function, this could be a pretty severe breaking change. I believe @Metious has more knowledge on this part of the code as well, so I'll wait on a review from him.

32Kallies (changes requested):

